PR #60, opened 2021-03-02
fix(cli): ui:create:vue command not invoking @coveo/typescript correctly

y-lakhdar:
The difference between add and invoke:

- `add [options] <plugin> [pluginOptions]`       install a plugin and invoke its generator in an already created project
- `invoke [options] <plugin> [pluginOptions]`  invoke the generator of a plugin in an already created project

Vue CLI was invoking `@coveo/typescript` before it was added to the project

### DIFF
--- a/packages/cli/src/commands/ui/create/vue.ts
+++ b/packages/cli/src/commands/ui/create/vue.ts
@@ -72,8 +72,8 @@ export default class Vue extends Command {
     const {providerUsername} = await this.getUserInfo();
 
     const cliArgs = [
-      'invoke',
-      '@coveo/vue-cli-plugin-typescript',
+      'add',
+      '@coveo/typescript',
       '--orgId',
       cfg.organization,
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/ui/create/vue.ts
```diff
@@ -72,8 +72,8 @@ export default class Vue extends Command {
     const {providerUsername} = await this.getUserInfo();
 
     const cliArgs = [
-      'invoke',
-      '@coveo/vue-cli-plugin-typescript',
+      'add',
+      '@coveo/typescript',
```
y-lakhdar, line 76: Since the package now respects the Vue-cli plugin standard, we can use the short version, which is pretty cool.

olamothe (approved):

louis-bompart (approved):
Was scared for a bit I thought we were using a custom version of typescript.

PR #62, opened 2021-03-02
fix(cli): vue Cli does not accept "latest" dependencies in plugin

y-lakhdar:
The Vue CLI is doing a `semver.validRange` validation on the dependency version that we want to inject.
Since `semver.validRange('latest')` returns `null`, the associated dependencies are never added to the project.

### DIFF
--- a/packages/vue-cli-plugin-typescript/generator/index.js
+++ b/packages/vue-cli-plugin-typescript/generator/index.js
@@ -6,8 +6,8 @@ module.exports = (api, options, rootOptions) => {
       'start-server': 'node ./scripts/start-server.js',
     },
     dependencies: {
-      '@coveo/headless': '^0',
-      '@coveo/search-token-server': '^0',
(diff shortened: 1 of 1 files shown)


packages/vue-cli-plugin-typescript/generator/index.js
```diff
@@ -6,8 +6,8 @@ module.exports = (api, options, rootOptions) => {
       'start-server': 'node ./scripts/start-server.js',
     },
     dependencies: {
-      '@coveo/headless': 'latest',
-      '@coveo/search-token-server': 'latest',
+      '@coveo/headless': '^0',
```
olamothe, line 9: Could this be `*` ?

y-lakhdar: Yes. It is accepted.

olamothe (approved):

PR #64, opened 2021-03-03
ci(release): change github token

olamothe:
Apparently, secrets.COMMIT_TOKEN is only available for private repositories, not public one 🤷 
 
https://coveord.atlassian.net/browse/CDX-134


y-lakhdar (approved):

louis-bompart (approved):
An attempt was made 😅

PR #68, opened 2021-03-05
test(cli): setup login util for E2E

y-lakhdar:


packages/cli-e2e/entrypoints/entry.js
```diff
@@ -57,7 +57,11 @@ try {
   execSync(
     `docker run --name=${DOCKER_CONTAINER_NAME} -v "${repoHostPath}:${repoDockerPath}" -p "9229:9229" -${
       process.argv[2] === '--bash' ? 'it' : 'i'
-    } --cap-add=SYS_ADMIN ${DOCKER_IMAGE_NAME} ${dockerEntryPoint}`,
+    } --cap-add=SYS_ADMIN --env PLATFORM_USER_NAME=${
```
y-lakhdar, line 60: TODO: pass env variable in .env file

packages/cli-e2e/entrypoints/dockerHeadless.sh
```diff
@@ -10,5 +10,12 @@ cd /home/cli-copy
```
y-lakhdar, line 10: TODO: update with CDX-142 before merge

packages/cli-e2e/entrypoints/dockerX11Entry.sh
```diff
@@ -9,5 +9,10 @@ cd /home/cli-copy
 
 sudo -u notGroot npm run setup
```
y-lakhdar, line 10: TODO: update with CDX-142 before merge

packages/cli-e2e/entrypoints/dockerX11Entry.sh
```diff
@@ -9,5 +9,10 @@ cd /home/cli-copy
...
 cd packages/cli-e2e
+
+mkdir dotenv
+echo "process.env.PLATFORM_USER_NAME='$PLATFORM_USER_NAME'" > dotenv/config.ts
```
olamothe, line 14: Similar to my question below, why do we need to set that in a dotenv file/config file ? 
Is that for jest ?

If they are set as env variable when we spawn docker, could we just pick them up inside jest ?

olamothe (approved):

packages/cli-e2e/entrypoints/dockerX11Entry.sh
y-lakhdar: Yes it's for Jest. 
And no, I was not able to use env variables in the tests without installing `dotenv`.

Instead, I used the `setupFiles` parameter to read the `dotenv/config.ts` file.

packages/cli-e2e/entrypoints/entry.js
y-lakhdar: it's just to prevent having a really long command. But this is not important. I can put this in the backlog.

packages/cli-e2e/utils/login.ts
```diff
@@ -0,0 +1,109 @@
...
+    });
+  });
+}
+
+async function staySignedIn(page: Page) {
+  await page.waitForSelector('input[type="submit"]');
+  await page.click('input[type="submit"]');
+}
+
+async function possiblyAcceptCustomerAgreement(page: Page) {
```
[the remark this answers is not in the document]
y-lakhdar: The first time I ran the. I kept that function there because I don't know if we will hit that page again.

packages/cli-e2e/utils/login.ts
```diff
@@ -0,0 +1,109 @@
+import {Page} from 'puppeteer-core';
+import type {Browser} from 'puppeteer-core';
+import {ChildProcessWithoutNullStreams, spawn} from 'child_process';
+import {answerPrompt, CLI_EXEC_PATH, isYesNoPrompt} from './cli';
+
+function isLoginPage(page: Page) {
+  return page.url().match(/https:\/\/platform.*cloud\.coveo\.com\/login/);
+}
+
+function isLoggedIn(page: Page) {
```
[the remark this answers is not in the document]
y-lakhdar: Will the user alway be redirected to the "redirect uri" page when logged in?
There probably a better way to let the test know it is logged in.

packages/cli-e2e/entrypoints/entry.js
louis-bompart: I think you could include https://www.npmjs.com/package/dotenv and that'd be way easier and cleaner than now.

How's it's done is really too hackish imho.

louis-bompart (changes requested):
dotenv is imo needed in this PR, the piping to the ts file is really hackish :(

packages/cli-e2e/utils/login.ts
```diff
@@ -0,0 +1,109 @@
...
+  return page.url().match(/http:\/\/127\.0\.0\.1:32111/);
+}
+
+function waitForLoginPage(browser: Browser) {
+  return new Promise<Page>((resolve) => {
+    browser.on('targetchanged', (page: Page) => {
```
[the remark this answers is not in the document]
y-lakhdar: Sorry, I'm not sure to understand how I can use `promisify` here :S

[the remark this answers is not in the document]
olamothe: > Takes a function following the common error-first callback style, i.e. taking a (err, value) => ... callback as the last argument, and returns a version that returns promises.

So, actually never mind, it is not an `an error first callback pattern`, so it would not work in this case.

If it had been however, you could use (I think) `utils.promisify` on `browser.on('targetChanged')`

So you'd get something like: 

```
const page = await utils.promisify(browser.on('targetchanged'))
```

Anyway, does not apply in this case ! NVM me ;)

packages/cli-e2e/utils/login.ts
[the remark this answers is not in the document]
olamothe: One thing we could do is login once at the beginning of the tests suite, and keep the token in the keychain.

So the `loggedIn` state would actually be "is the token available in the keychain".

But I understand what you mean with the final redirection.

I think it's mostly a misunderstanding with the function name.

`isLoginFlowFinished` might be better ?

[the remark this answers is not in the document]
y-lakhdar: I think logging in at the beginning of the test suite is a good idea and also much cleaner. However, it causes the test in `auth.spec.ts` to fail since the redirection to the login page is not happening when already logged in (automatically redirected to the redirect uri).

I was thinking about these 2 options:
1. for every test file, login in the setup phase and logout in the teardown phase
2. discard the `auth.spec.ts` since the `auth:login` command is already  tested other files

Anyway, let me know what you think...

packages/cli-e2e/utils/login.ts
```diff
@@ -8,14 +10,20 @@ function isLoginPage(page: Page) {
...
 
-function isLoggedIn(page: Page) {
-  return page.url().match(/http:\/\/127\.0\.0\.1:32111/);
+export async function isLoggedin() {
+  const currentAccount = userInfo().username;
+  const accessToken = await getPassword(
+    'com.coveo.cli.access.token',
+    currentAccount
+  );
+  return accessToken !== null;
```
y-lakhdar, line 19: The `isLoggedin` function is now based on the keychain rather than a redirection to the redirect uri

packages/cli-e2e/utils/login.ts
```diff
@@ -8,14 +10,20 @@ function isLoginPage(page: Page) {
...
 }
 
 function waitForLoginPage(browser: Browser) {
   return new Promise<Page>((resolve) => {
-    browser.on('targetchanged', (page: Page) => {
-      if (page.url && isLoginPage(page) !== null) {
+    browser.on('targetchanged', async (target: Target) => {
+      const page = await target.page();
+      if (page && isLoginPage(page) !== null) {
```
y-lakhdar, line 26: I had the wrong type. [targetchanged](https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-event-targetchanged)  returns a [Target](https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-class-target) object... not a page object.

packages/cli-e2e/utils/login.ts
```diff
@@ -93,17 +98,24 @@ async function startLoginFlow(browser: Browser) {
...
   const loginProcess = runLoginCommand();
   cliProcesses.push(loginProcess);
 
   await startLoginFlow(browser);
 }
+
+export async function logout() {
+  const currentAccount = userInfo().username;
+  await deletePassword('com.coveo.cli.access.token', currentAccount);
+}
```
y-lakhdar, line 121: Added a `logout` function to use in the teardown phases

packages/cli-e2e/utils/login.ts
```diff
@@ -0,0 +1,121 @@
...
+  });
+}
+
+async function staySignedIn(page: Page) {
+  await page.waitForSelector('input[type="submit"]');
+  await page.click('input[type="submit"]');
+}
+
+async function possiblyAcceptCustomerAgreement(page: Page) {
+  if (page.url().match(/https:\/\/platform.*cloud\.coveo\.com\/eula.?/)) {
```
louis-bompart, line 39: Mhhh. I would create a URL object from `page.url()` and manipulate it instead. (I personnally, tho it's taste/flavoring don't like using regexp for URL validations).

That's my opinion tho!

packages/cli-e2e/utils/login.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  await waitForLoginPage(browser);
+
+  const pages = await browser.pages();
+  const page = pages.find((page) => isLoginPage(page));
+  if (!page) {
+    throw new Error('Unable to find login page');
+  }
+
+  await page.waitForSelector('#loginWithOffice365');
```
louis-bompart, line 80: If I were you, I would create a class that contains static to describe each selector:

```ts
class CoveoLoginPage {
public static readonly loginWithOfficeButton = '#loginWithOffice365';
}
```

I would then import it here

packages/cli-e2e/utils/login.ts
```diff
@@ -0,0 +1,121 @@
...
+  await page.click('#loginWithOffice365');
+
+  await page.waitForNavigation();
+
+  await page.waitForSelector('input[type="email"]');
+  await page.type('input[type="email"]', username);
+  await page.waitForSelector('input[type="submit"]');
+  await page.click('input[type="submit"]');
+
+  await page.waitForTimeout(500);
```
louis-bompart, line 90: Avoid timeout when possible.
Here you should be able to uses https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-pagewaitforselectorselector-options

packages/cli-e2e/utils/login.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  await page.waitForSelector('input[type="password"]');
+  await page.type('input[type="password"]', password);
+  await page.waitForSelector('input[type="submit"]');
+  await page.click('input[type="submit"]');
+
+  await page.waitForTimeout(500);
```
louis-bompart, line 97: Same comment as before, see if there's not something we could expect to have instead of just some timeout :)

louis-bompart (changes requested):
Good job overall!
I would however avoid the uses of timeout ;)

packages/cli-e2e/utils/login.ts
y-lakhdar: true! I tried again and it works with `waitForSelector`. I just needed to add the visible parameter
`await page.waitForSelector('input[type="submit"]', { visible: true });`

I'll change it

packages/cli-e2e/utils/login.ts
y-lakhdar: There are some cases where waitForSelector will not work. One example would be when the button is visible but is not clickable. I was able to replace only one `waitForTimeout`. The other one is needed.

packages/cli-e2e/utils/login.ts
y-lakhdar: I'll just hardcode the dev environment for now. We can update the validation in CDX-98

packages/cli-e2e/utils/login.ts
louis-bompart: Visible but not clickable? How so/what gives?

packages/cli-e2e/utils/login.ts
louis-bompart: add a todo comment then ;)

y-lakhdar: Yes, I added it in my [last commit](https://github.com/coveo/cli/pull/68/commits/21e110246c15224f256806d69ddafe55ef907605#diff-f882e2d46f5ef618509857e8a5a62df6468201ee6ed6fa39b35c83a2b2d4d765R10).

packages/cli-e2e/utils/login.ts
y-lakhdar: The element is available in the DOM, but not ready to be interacted with. I added a breakpoint right after the `await page.waitForSelector('input[type="password"]', { visible: true });` instruction and saw that:
image

y-lakhdar:
I added a `isElementClickable` function to prevent relying on the timeout.
It's a bit overkill but it works every time.

packages/cli-e2e/utils/login.ts
louis-bompart: I would widen the check to falsy value
```suggestion
  return Boolean(accessToken);
```

I know you can use `!!` to do the same thing, but I think it's clearer because explicit.

packages/cli-e2e/utils/login.ts
```diff
@@ -0,0 +1,126 @@
...
+  await page.waitForSelector(LoginSelectors.SubmitInput);
+  await page.click(LoginSelectors.SubmitInput);
+}
+
+async function possiblyAcceptCustomerAgreement(page: Page) {
+  // TODO: CDX-98: URL should vary in fonction of the targeted environment.
+  if (page.url().startsWith('https://platformdev.cloud.coveo.com/eula')) {
+    await page.waitForSelector(LoginSelectors.coveoCheckboxButton);
+    await page.click(LoginSelectors.coveoCheckboxButton);
+    await page.waitForTimeout(200); // wait for the button to be enabled
```
louis-bompart, line 48: Can't you use isClickable here?

louis-bompart (approved):
LGTM, check to remove the last timeout tho

packages/cli-e2e/utils/login.ts
y-lakhdar: I can't validate anymore since we already accepted the agreement with that account

packages/cli-e2e/tsconfig.json
```diff
@@ -7,7 +7,6 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "forceConsistentCasingInFileNames": true,
-    "types": ["jest"],
-    "lib": ["dom"]
```
y-lakhdar: I added it to this PR, but it is finally unnecessary.
It also causes the build to fail.

PR #69, opened 2021-03-05
test(cli): add E2E tests for ui:create:vue

y-lakhdar:


packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -0,0 +1,113 @@
...
+        'ui:create:vue',
+        projectName,
+      ]);
+
+      expect(buildProcess.status).toEqual(0);
+
+      const waitForProjectToStart = new Promise<void>((resolve) => {
+        const startServerProcess = spawn('npm', ['run', 'start'], {
+          cwd: projectName,
+          detached: true,
```
y-lakhdar, line 36: So I finally manage to prevent process leaking.
Every time I spawn a process that will later create its own family of processes, I am using the `detach` option.

> On non-Windows platforms, if `options.detached` is set to true, the child process will be made the **leader** of a new process group and session
[reference](https://nodejs.org/api/child_process.html#child_process_options_detached)

That way, the entire process family can be killed using the leader process id.
It is equivalent to run the following command `kill -TERM -- -<leader-process>`. 
Notice the negative sign before the process id

packages/cli-e2e/utils/cli.ts
```diff
@@ -9,6 +9,16 @@ export function killCliProcess(cliProcess: ChildProcessWithoutNullStreams) {
   return waitForKill;
 }
 
+export function killCliProcessFamily(
+  cliProcessLeader: ChildProcessWithoutNullStreams
+) {
+  const waitForKill = new Promise<void>((resolve) => {
+    cliProcessLeader.on('close', () => resolve());
+  });
+  process.kill(-cliProcessLeader.pid);
```
y-lakhdar, line 18: As explained above, this command kill and process within the same group.

CHANGELOG.md
```diff
@@ -5,67 +5,54 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
```
louis-bompart, line 5: #71 created to avoid this

packages/cli-e2e/utils/cli.ts
louis-bompart: 🥇 nice finding!

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -0,0 +1,113 @@
...
+      browser = await getBrowser();
+
+      await loginWithOffice(browser, cliProcesses);
+
+      const buildProcess = spawnSync(CLI_EXEC_PATH, [
+        'ui:create:vue',
+        projectName,
+      ]);
+
+      expect(buildProcess.status).toEqual(0);
```
louis-bompart, line 31: Why is there an expect in a beforeAll?

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -0,0 +1,113 @@
...
+      const page = await browser.newPage();
+      const searchboxSelector = '#search-page .autocomplete input';
+      let isInterfaceLoadSearch = true;
+
+      page.on('request', (request) => {
+        if (
+          request.method() === 'POST' &&
+          request.url().indexOf(
+            // TODO: CDX-98: URL should vary in fonction of the targeted environment.
+            'https://platformdev.cloud.coveo.com/rest/search/v2?organizationId'
```
louis-bompart, line 90: This if is lengthy and hard to read, you should refactor this in a method :) 
e.g. `isSearchQuery`

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    it('should trigger search queries', async () => {
+      const page = await browser.newPage();
+      const searchboxSelector = '#search-page .autocomplete input';
+      let isInterfaceLoadSearch = true;
+
+      page.on('request', (request) => {
+        if (
+          request.method() === 'POST' &&
+          request.url().indexOf(
```
louis-bompart, line 88: You can uses `startsWith`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -0,0 +1,113 @@
...
+        token: expect.stringMatching(/.+/),
+      });
+    });
+
+    it('should trigger search queries', async () => {
+      const page = await browser.newPage();
+      const searchboxSelector = '#search-page .autocomplete input';
+      let isInterfaceLoadSearch = true;
+
+      page.on('request', (request) => {
```
louis-bompart, line 85: I would do something a bit more crude maybe for the request interception but that would help with the reading of the test:
- Push all the requests in an array on the hook.

Then, using an async-retry do something like
`requests.some(isSearchRequest)` with `iSearchRequest being a function taking a request in param and returning true or false.

I would do it _once_ before interacting with the page, to check for the initial query.
Then, I would clear the array and do the search.
And do the requests.some again.

Why I'd do it like this?

Well, I think that setting up the hook on request is a 'setup' phase and should be done early and cleanly without any filtering. This makes clear that 'OK we listen on request for the test'.

Then you do an action:
`await page.goto(searchPageEndpoint, {waitUntil: 'networkidle2'});`
and right after you assert that there should be a request made to Platform.

Like this, the action and what it causes are close together and clear to the reader/tester/developer.

packages/cli-e2e/utils/cli.ts
```diff
@@ -9,6 +9,16 @@ export function killCliProcess(cliProcess: ChildProcessWithoutNullStreams) {
   return waitForKill;
 }
 
+export function killCliProcessFamily(
+  cliProcessLeader: ChildProcessWithoutNullStreams
+) {
+  const waitForKill = new Promise<void>((resolve) => {
```
louis-bompart, line 15: You prolly could use promisify?

louis-bompart (commented):
Good overall!
Some tweaks to help with the readability of the test that should not be too costly and will set a good trend and :shipit:

packages/cli-e2e/__tests__/vue.specs.ts
y-lakhdar: oh I did not think about it. That's way cleaner 👍 .
I'm just not sure why we would need the async retry here.
Since we have specified the `waitUntil:"networkidle2"` option, we should be able to read all the intercepted requests right after the `page.goto` line.

As it goes for the second reading,  we can simply add an
`await page.waitForNetwork({ waitUntil: "networkidle0" });` to make sure we wait after all the search requests.

packages/cli-e2e/utils/cli.ts
y-lakhdar: I just copied your `killCliProcess` function :P 
But yeah, we can use use `promisify`

louis-bompart: Ahahah that ricochet xD

packages/cli-e2e/__tests__/vue.specs.ts
y-lakhdar: Oh nvm. we need the async retry for the second reading :P

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -0,0 +1,111 @@
...
+    // TODO: CDX-90: Assign a dynamic port for the search token server on all ui projects
+    const clientPort = '8080';
+    const projectName = 'vue-project';
+    const searchPageEndpoint = `http://localhost:${clientPort}`;
+    const tokenProxyEndpoint = `http://localhost:${clientPort}/token`;
+    let interceptedRequests: HTTPRequest[] = [];
+    let page: Page;
+    let waitForProjectToStartTimeout: NodeJS.Timeout;
+
+    beforeAll(async () => {
```
olamothe, line 25: I think there's some thing we could extract in a shared file, similar to what you have for `loginWithOffice`.

Let's say a `createUIProject()` function, that receives the CLI command to execute ( in this case: `ui:create:vue`) , and a command to run when the project is generated ( in this case npm run start ).

It takes care of doing all the boilerplate here, as well as the not so obvious `detached process` configuration.

I can definitely see this being reused for all UI commands, at least.

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    beforeEach(async () => {
+      page = await browser.newPage();
+
+      page.on('request', (request: HTTPRequest) => {
+        interceptedRequests.push(request);
+      });
+    });
+
+    afterAll(async () => {
```
olamothe, line 61: Similar comment as above about this `afterAll` being potentially generic for all UI commands.

packages/cli-e2e/utils/platform.ts
```diff
@@ -0,0 +1,8 @@
+import {HTTPRequest} from 'puppeteer-core';
+
+export function isSearchRequest(request: HTTPRequest) {
+  return request.url().startsWith(
```
olamothe, line 4: I would just do a regex match valid for any environment:

```
https:\/\/platform[a-z]?\.cloud.coveo.com\/rest\/search
```

packages/cli-e2e/utils/platform.ts
```diff
@@ -0,0 +1,8 @@
+import {HTTPRequest} from 'puppeteer-core';
+
+export function isSearchRequest(request: HTTPRequest) {
```
olamothe, line 3: This file is a bit weird as it's the only function in there.

How about we have just one "test utils" file, and we split it when we see that it start becoming bloated with a ton of stuff ?

Having a bunch of different utils file like that causes other types of problems, where it's hard to know what already exists or not.

packages/cli-e2e/utils/platform.ts
y-lakhdar: I'll do a mass update of url platform validation in CDX-98

packages/cli-e2e/utils/platform.ts
y-lakhdar: Maybe we can move it in the browser utils, so we can have 3 utils (for now):
* cli
* browser
* login (or auth)

PR #70, opened 2021-03-08
test(cli): modify test env to be keyring-compatible

louis-bompart:
This PR modify the docker substantially by making the user being `notGroot` by default (so accessing root will require `sudo` now and not the other way around)

The DockerFile is also augmented with the responsibility of initializing some files for `notGroot` mainly its keychain. By using an unencrypted keychain (i.e. no passphrase) we avoid any GUI prompt from gnome-keychain when trying to write or read from the keychain.

---

https://coveord.atlassian.net/browse/CDX-142


y-lakhdar:
Do we need to run the docker container with the `--privileged` flag?

louis-bompart:
> Do we need to run the docker container with the `--privileged` flag?

Hm I forgot to check the entry.js I was a bit too hasty.

louis-bompart:
@y-lakhdar so, no --privileged was a bit an 'umbrella' but only IPC_LOCK is needed.

y-lakhdar (approved):

PR #72, opened 2021-03-10
feat(cli): support for auto update packaging 

olamothe:
Uses/add https://github.com/oclif/plugin-update to support an `update` command.

The plugin is closely coupled to the way `oclif-dev pack` works, and the folder structure it generates.

Previously, we would only generate the install kit for the different platform for the CLI, with `oclif-dev pack:macos`, `oclif-dev pack:win`, `oclif-dev pack:deb` respectively.

Those install kit are put under `./dist/coveo-v0.4.0.pkg`,  `./dist/coveo-v0.4.0-x64.exe`,  `./dist/coveo-v0.4.0-x86.exe`, etc.. 

Those install kit will be used by our end users to do a fresh install of the CLI.

oclif dev also exposes a `oclif-dev pack` command, which does not generate an install kit, but purely the tarball (.tar.gz) files that contains the CLI per say, and not any "installation" code or instruction.

That command generates a tarball for all platform (AKA: download and package the corresponding node binary for each platform).

Once those are generated, they are put under `./dist/coveo-v0.4.0/coveo-v0.4.0-darwin-x64.tar.gz`, `./dist/coveo-v0.4.0/coveo-v0.4.0-win32-x64.tar.gz` `./dist/coveo-v0.4.0/coveo-v0.4.0-linux-x64.tar.gz`, etc.

The `oclif-dev pack` command also generates a bunch of "JSON" file for each platform. Those JSON files are put under `./dist/darwin-x64`, `./dist/win32-x86`, `./dist/win32-x64` etc..

Those file contains information like this:

```
{
  "version": "0.4.0",
  "channel": "stable",
  "baseDir": "coveo",
  "gz": "https://static.cloud.coveo.com/cli/coveo-v0.4.0/coveo-v0.4.0-darwin-x64.tar.gz",
  "sha256gz": "9bb8ab561dffebae4f29bd6b28fa7571b9d3335e6c1662f62a95a9b9128a1ce9",
  "node": {
    "compatible": ">=14.15.1",
    "recommended": "14.15.1"
  }
}
```

And then finally, in `package.json`, we specify the "bucket" (in this case, not a bucket but a CDN link) from where to fetch the new versions. 

When you run the `update` command, it will look at the top level of the specific bucket (https://static.cloud.coveo.com/cli/) for the corresponding platform JSON file. 

It will compare the version specified with the current one.
If an update can be made, it then uses the `gz` key from the JSON to get the download URL.

It then replace the CLI tarball installed locally on the machine with the new one.

Hopefully that was clear enough ! ;)


louis-bompart (approved):
LGTM,
How does the hierarchy work between channel and OSes?

Do we have one channel per OS, or do we have one channel for all OS, and oclif is smart enough to uses the good OS?

PR #74, opened 2021-03-10
fix(cli): add a `--not-git` flag on the command `ui:create:vue`

y-lakhdar:
This flag is needed to prevent the tests from crashing.

The idea is to have a hidden `--test` flag for each `ui:create:` command. That way, we can adjust the instruction if it is running in test mode.

In the case of Vue.js, the test mode will add the `--no-git` flag, which will prevent the initialization of a git repository during the project creation.

Otherwise, the Vue CLI will return the following error.
```
missing username and email in git config, or failed to sign a commit
```

If you run the E2E tests in your local machines while having uncommitted changes, the Vue Cli will prompt a warning and wait for the user input.
Adding the `--no-git` flag will prevent this scenario

### DIFF
--- a/packages/cli/src/commands/ui/create/vue.ts
+++ b/packages/cli/src/commands/ui/create/vue.ts
@@ -11,7 +11,7 @@ import {Storage} from '../../../lib/oauth/storage';
 import {platformUrl} from '../../../lib/platform/environment';
 import {spawnProcess} from '../../../lib/utils/process';
 
-interface IprojectOptions {
+interface IProjectOptions {
   preset: {};
   test: boolean;
 }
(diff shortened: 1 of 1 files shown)


louis-bompart:
@y-lakhdar for funsies, can you try to init the git config of not groot in the dockerfile or the bash files?
I ain't fond of doing a 'test path' in our logic, especially for E2E

olamothe:
How about we add a `-- passthrough` kind of command ? 

That we document/make official. And you can use it to just "pass through options to the vue cli" ?
And then in our tests we would do `--passthrough not-git` or something similar to this ?

There could be some use by end users to be able to pass options to the vue CLI underneath.

y-lakhdar:
I like the passthrough approach!

y-lakhdar:
I'll close this one for now and will update the Dockerfile as proposed by @louis-bompart .
If we really need a `passthrough` feature, I'll create a separate Jira.

PR #78, opened 2021-03-12
test(cli): add global setup and teardown

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-150


packages/cli-e2e/setup.ts
```diff
@@ -0,0 +1,23 @@
+import type {Browser} from 'puppeteer';
+import {connectToChromeBrowser} from './utils/browser';
+import {clearKeychain, loginWithOffice} from './utils/login';
+
+async function clearChromeBrowsingData(browser: Browser) {
+  const pages = await browser.pages();
+
+  const pageClearDataPromise = [];
+  for (const page of pages) {
```
louis-bompart, line 9: Do we need to do it on all pages 🤔 ?

louis-bompart (approved):
Looks good!

packages/cli-e2e/setup.ts
y-lakhdar: Tbh, I am not sure. I stupidly applied it to all pages just in case. :S

olamothe (approved):

packages/cli-e2e/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "skipLibCheck": true,
     "sourceMap": true,
     "forceConsistentCasingInFileNames": true,
-    "types": ["jest"]
+    "types": ["jest"],
+    "lib": ["dom", "es2017"]
```
y-lakhdar, line 11: @louis-bompart, this additional line is required for the `isElementClickable` function

PR #82, opened 2021-03-15
fix(vue): remove environment variables type warning

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-156


y-lakhdar:
It is safe to assume the environment variables will always be defined because of the `isEnvValid` validation done upstream.

olamothe (approved):

PR #83, opened 2021-03-15
fix(cli): make oclif-dev pack:win works on Windows

louis-bompart:
### Changes
 - Replaced `rm` by `rimraf` for Windows compatibility
 - Patched oclif-dev to make it works on Windows:
    - Removed throw on windows.
    - Removed grep on logs, it was useless.
    - Added Windows alternative for checksum using windows cmd native commands (plus some node for string manipulations)
    - Added Windows alternative for `/dev/null`, which is NUL.
- Changed the CLI structure so that the 'packaging' of each OS target occurs on relevant OS (darwin is packed on mac, win-xxx is packed on windows, Linux is packaged on Linux).
----------
https://coveord.atlassian.net/browse/CDX-155


.github/workflows/build-binaries.yml
```diff
@@ -1,8 +1,11 @@
 name: Build binaries
 
 on:
-  release:
-    types: [published]
+  push:
```
olamothe, line 4: We should remove this/put it back to just `release: types: [published]`

.github/workflows/build-binaries.yml
```diff
@@ -16,18 +19,17 @@ jobs:
             pack-command: :macos
             additional-install: echo "no additional install"
             for: 'macos install kit'
-          - os: ubuntu-latest
+            package-targets: 'darwin-x64'
+          - os: windows-latest
             pack-command: :win
-            additional-install: sudo apt-get install nsis p7zip-full -y
+            additional-install: echo "no additional install"
```
olamothe, line 25: I think all the `additional-install` in the matrix could now be removed: it was needed when trying to pack:win on linux, makensis + 7zip were installed.

Now it does not seem like it's of any use.

.github/workflows/build-binaries.yml
```diff
@@ -38,13 +40,17 @@ jobs:
         run: npm run setup
       - name: Additional install
         run: ${{ matrix.additional-install }}
-      - name: Create package
+      - name: Create executables
```
olamothe, line 43: Suggestion: `Create install kit`

olamothe (commented):
Have you been able to test a full loop by deleting all the assets in `v0.5.0`, and running the workflow ?

olamothe:
Also, I would actually flag this commit as a `fix`: it's fixing a bug that made the CLI unusable on some platform

patches/@oclif+dev-cli+1.26.0.patch
```diff
@@ -0,0 +1,82 @@
+diff --git a/node_modules/@oclif/dev-cli/lib/commands/pack/index.js b/node_modules/@oclif/dev-cli/lib/commands/pack/index.js
```
olamothe, line 1: This should be generated at `packages/cli/patches` , aka one level down, and not at root of the repo.

This means that when we `npm i` in each subpackage, the patch will only be applied there.

louis-bompart:
So the build worked, but the upload is having trouble because there are conflicting files with the uploading.
To fix it:
- [ ] Use artifacts in build steps
- [ ] Create a separate step for uploading the binaries by consuming the artifacts

patches/@oclif+dev-cli+1.26.0.patch
louis-bompart: Fixed, resolving convo

louis-bompart:
@olamothe, check the artifact on https://github.com/coveo/cli/pull/83/commits/7c2d22c0885ca254a398ef77593910f276399915
I'm not sure how the auto-update works really, but if you check the `coveo-v0.5.0.tar.gz` in each artifact, you'll see that it's the zippin' of what's end up packaged and it's referred by the `version` file.

What are those files for (`version` and `coveo-v0.5.0.tar.gz`)?

olamothe (approved):

PR #85, opened 2021-03-15
fix(react): use and detect local npx to create-react-app

olamothe:
Try to detect valid `node` version 10.16.0 ( https://github.com/facebook/create-react-app#creating-an-app ) + npx.

Otherwise, warn and exit, with some (hopefully) helpful message about how to install node with nvm.

https://coveord.atlassian.net/browse/CDX-102

### DIFF
--- a/packages/cli/src/commands/ui/create/react.ts
+++ b/packages/cli/src/commands/ui/create/react.ts
@@ -5,11 +5,11 @@ import {
 } from '../../../hooks/analytics/analytics';
 import {Config} from '../../../lib/config/config';
 import {platformUrl} from '../../../lib/platform/environment';
-import {spawnProcess} from '../../../lib/utils/process';
+import {spawnProcess, spawnProcessStdio} from '../../../lib/utils/process';
 import {Storage} from '../../../lib/oauth/storage';
 import AuthenticationRequired from '../../../lib/decorators/authenticationRequired';
 import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
-
+import {lt} from 'semver';
 export default class React extends Command {
   static description =
     'Create a Coveo Headless-powered search page with the React web framework. See https://docs.coveo.com/en/headless and https://reactjs.org/.';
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/ui/create/react.ts
```diff
@@ -97,4 +104,65 @@ export default class React extends Command {
...
+    const stdio = await spawnProcessStdio('npx', ['--version']);
+
+    if (stdio.stderr && stdio.stderr.match(/ENOENT/i)) {
+      this.warn(`${this.id} requires npx to run.`);
+      this.warn('Newer version Node.js comes bundled with npx.');
+      this.warnHowToInstallNode();
+      return false;
+    }
+
+    if (stdio.stderr) {
```
y-lakhdar, line 149: npx depends on [update-notifier](https://github.com/npm/npx/blob/latest/package.json#L42).
What would happen if the user has an older version of npx and gets the update warning message? Will it be redirected to the stderr?

packages/cli/src/lib/utils/process.ts
```diff
@@ -34,3 +35,31 @@ export function spawnProcess(
...
   });
 }
+
+export async function spawnProcessStdio(
+  command: string,
+  args: string[],
+  options: SpawnOptions = {}
+): Promise<{stdout: string; stderr: string}> {
+  return new Promise((resolve) => {
+    const stdio = {stdout: '', stderr: ''};
```
louis-bompart, line 45: Add exitCode to the stdio (prolly should be renamed to something else like `processOutputs`)
```suggestion
    const stdio = {stdout: '', stderr: '', exitCode: null};
```

packages/cli/src/lib/utils/process.ts
```diff
@@ -34,3 +35,31 @@ export function spawnProcess(
...
+    });
+
+    child.on('error', (e) => {
+      stdio.stderr += e.message;
+      resolve(stdio);
+    });
+
+    child.on('close', () => {
+      resolve(stdio);
+    });
```
louis-bompart, line 63:
```suggestion
    child.on('close', (code) => {
      stdio.exitCode = code;
      resolve(stdio);
    });
```

louis-bompart (changes requested):
Few suggestions.
More importantly, I would write a markdown page (maybe on the wiki of the repo?) to explains in more length what's what to the user and put a link to it in the message

packages/cli/src/commands/ui/create/react.ts
```diff
@@ -97,4 +112,77 @@ export default class React extends Command {
 
     return await platformClient.user.get();
   }
+
+  private async checkIfUserHasNodeGreaterThan10() {
+    const output = await spawnProcessOutput('node', ['--version']);
+    if (this.isMissingExecutable(output)) {
+      this.warn(`
```
olamothe, line 119: You will see I've split up all the `warn` message with an empty line at the start and end of the message.

It help readability in the CLI at the end, with a nice "paragraph" separation kinda thing.

eg:

```
›   Warning:
›         ui:create:react requires a valid Node.js installation to run.
›         An unknown error happened while trying to determine your node version with node --version
›         spawn foo ENOENT
```

louis-bompart (approved):

PR #86, opened 2021-03-16
fix(angular): add definite assignment assertion to props

y-lakhdar:
Another option would be to turn off strict type checking on the project creation...

https://coveord.atlassian.net/browse/CDX-101


louis-bompart:
> Another option would be to turn off strict type checking on the project creation...
> 
> https://coveord.atlassian.net/browse/CDX-101

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#strict-class-initialization

I don't think we should, it's a good thing IMO to check this. It ensures proper typings in the end.

louis-bompart (approved):

PR #87, opened 2021-03-16
build(cli): testing tests from pull request

louis-bompart:
🥖 hon hon hon


louis-bompart:
Yeah 's just to test the E2E no need for review.

louis-bompart:
Test prolly fixed in #88

PR #89, opened 2021-03-17
fix(angular): copy schema only once in postbuild script

louis-bompart:
https://coveord.atlassian.net/browse/CDX-163


y-lakhdar (approved):
👍

PR #91, opened 2021-03-17
ci(cli): modify dep config to include snyk certifier

louis-bompart:
https://coveord.atlassian.net/browse/CDX-148


louis-bompart:
@olamothe , do you know how to check that it works?
Running dep pipeline?

olamothe (approved):

PR #92, opened 2021-03-17
Inspect docker container memory

y-lakhdar:


y-lakhdar:
No need to update docker configuration as github actions already provides 7G of RAM memory (link)

image

However, one can experience timeout issues with E2E tests if the Docker memory limit is too low.

Just increase the docker memory limit to something bigger than 2g on your machine

PR #93, opened 2021-03-17
fix(angular): `npm start` command on Angular project returns warning

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-164

### DIFF
--- a/packages/angular/src/common-rules/dependencies.ts
+++ b/packages/angular/src/common-rules/dependencies.ts
@@ -1,9 +1,45 @@
 import {Rule, SchematicContext, Tree} from '@angular-devkit/schematics';
 import {NodePackageInstallTask} from '@angular-devkit/schematics/tasks';
+import {CoveoSchema} from '../schema';
+import {normalize} from '@angular-devkit/core';
 
 export function runPackageInstallTask(workingDirectory?: string): Rule {
   return (tree: Tree, context: SchematicContext) => {
     const cfg = workingDirectory ? {workingDirectory} : {};
     context.addTask(new NodePackageInstallTask(cfg), []);
+  };
+}
+
+export function allowCommonJsDependencies(options: CoveoSchema): Rule {
+  return (tree: Tree, _context: SchematicContext) => {
+    const workspaceBuffer = tree.read(normalize('./angular.json'));
+    if (workspaceBuffer === null || !options.project) {
+      return;
+    }
+    try {
+      const workspaceConfig = JSON.parse(workspaceBuffer.toString());
+
+      const allowedCommonJsDependencies =
+        workspaceConfig.projects[options.project].architect.build.options[
+          'allowedCommonJsDependencies'
+        ] || [];
+
+      allowedCommonJsDependencies.push('@coveo/headless');
+
+      workspaceConfig.projects[options.project].architect.build.options[
+        'allowedCommonJsDependencies'
+      ] = allowedCommonJsDependencies;
+
+      tree.overwrite(
+        normalize('./angular.json'),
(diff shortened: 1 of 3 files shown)


packages/angular/src/ng-add/index.ts
```diff
@@ -15,11 +15,12 @@ export default function (options: CoveoSchema): Rule {
...
       );
-      context.addTask(new RunSchematicTask('ng-add-setup-project', options), [
-        dependencyInstallationTaskId,
-      ]);
+      const setupProjectTaskId = context.addTask(
+        new RunSchematicTask('ng-add-setup-project', options),
+        [dependencyInstallationTaskId]
+      );
       context.addTask(new RunSchematicTask('search-token-server', options), [
-        dependencyInstallationTaskId,
```
y-lakhdar, line 22: Making sure all 3 tasks are running one after the other to prevent multiple rules modifying the same file at the same time.

y-lakhdar:
More info here https://angular.io/guide/build#configuring-commonjs-dependencies

.github/workflows/build-binaries.yml
```diff
@@ -40,11 +40,11 @@ jobs:
         run: |
           ./node_modules/@oclif/dev-cli/bin/run pack -t="${{ matrix.package-targets }}"
           mv ./dist/version ./dist/${{ matrix.os }}-version
-          mv ./dist/coveo-v${{ github.ref }}/coveo-v${{ github.ref }}.tar.gz ./dist/coveo-v${{ github.ref }}/${{ matrix.os }}-coveo-v${{ github.ref }}.tar.gz
+          mv ./dist/coveo-${{GITHUB_REF/refs\/tags\//}}/coveo-${{GITHUB_REF/refs\/tags\//}}.tar.gz ./dist/coveo-${{GITHUB_REF/refs\/tags\//}}/${{ matrix.os }}-coveo-${{GITHUB_REF/refs\/tags\//}}.tar.gz
```
olamothe, line 43: Another cross branch pollution it seems :D

packages/angular/src/common-rules/dependencies.ts
```diff
@@ -1,9 +1,45 @@
...
 
 export function runPackageInstallTask(workingDirectory?: string): Rule {
   return (tree: Tree, context: SchematicContext) => {
     const cfg = workingDirectory ? {workingDirectory} : {};
     context.addTask(new NodePackageInstallTask(cfg), []);
   };
 }
+
+export function allowCommonJsDependencies(options: CoveoSchema): Rule {
+  return (tree: Tree, _context: SchematicContext) => {
```
olamothe, line 14: Nice the github warning ;) 

If it's not used, just discard the argument.

olamothe (commented):
Headless has different target, we can include the es module one. So the warning would go away for "the good reasons" ;)

y-lakhdar:
> Headless has different target, we can include the es module one. If you look into node_modules/@coveo/headless/dist, you should see .esm files.
> 
> So the warning would go away for "the good reasons" ;)

It seems like webpack is targeting the correct module file, but the Angular builder doesn't like the `const nodeAbort = require('node-abort-controller');` in the headless.esm.js file.

It returns a warning when the builder goes through all dependencies and finds that headless.esm.js contains a CommonJs depencency.
image

I also tried 
```
"allowedCommonJsDependencies": [
    "node-abort-controller"
],
```
but it does not work because the `hasCommonJsDependencies` will always return a warning for the parent if one of its dependencies is a commonJs or AMD dep.

olamothe (approved):
Alright then ! 🤔 

I'll log a story in Headless to see if we could do anything about it.

PR #94, opened 2021-03-17
test(cli): add E2E tests for ui:create:angular command

y-lakhdar:
* Deleted all spec files from the angular project. There's no point until we are not able to mock the headless engine.
* Complete E2E tests for angular project

https://coveord.atlassian.net/browse/CDX-153


packages/angular/.gitignore
```diff
@@ -1,7 +1,4 @@
 # Outputs
-src/**/*.js
-src/**/*.js.map
-src/**/*.d.ts
 dist
```
y-lakhdar, line 2: When I created the initial version of the schematic, it was outputing compiled JS files directly into the src folder. I updated the output folder but forgot to update the .gitignore.

packages/angular/src/search-token-server/files/scripts/start-server.js
```diff
@@ -0,0 +1,27 @@
+const {spawnSync} = require('child_process');
```
y-lakhdar, line 1: File that not intentionally excluded by the .gitignore

packages/cli-e2e/utils/cli.ts
```diff
@@ -48,13 +50,30 @@ export function answerPrompt(
...
 export async function setupUIProject(
-  commandArgs: string,
+  uiCommand: string,
   projectName: string,
-  cliProcesses: ChildProcessWithoutNullStreams[]
+  cliProcesses: ChildProcessWithoutNullStreams[],
+  options: ISetupUIProjectOptionsArgs = {}
 ) {
+  const uiProjectFolderName = 'ui-projects';
+  ensureDirSync(uiProjectFolderName);
```
louis-bompart, line 65: Where does that ends up?
It should be added to the ignore list of prettier, eslint, git, npm and jest

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      await page.waitForSelector(searchboxSelector);
+      await page.focus(searchboxSelector);
+      await page.keyboard.type('my query');
+      await page.keyboard.press('Enter');
+
+      await retry(async () => {
+        expect(interceptedRequests.some(isSearchRequest)).toBeTruthy();
+      });
+    });
```
louis-bompart, line 75: A thing you could do is to split the test in two tests:
`should send a search query when the page is loaded`
`should send a search query when the searchbox is used` (or something else)

packages/cli-e2e/jest.config.ts
```diff
@@ -7,4 +7,10 @@ export default {
   testEnvironment: 'node',
   globalSetup: './setup.ts',
   globalTeardown: './teardown.ts',
+  /**
+   * TODO: CDX-160: Include spec files from ui projects once we have a mock of the headless engine
+   * In th meantmime, do not execute tests outside the __tests__ folder.
+   * More info -> https://jestjs.io/docs/configuration#testmatch-arraystring
+   */
```
louis-bompart, line 14: I'm not sure that should be part of the End-to-end tests. (and that those tests are end-to-end tests)

louis-bompart (commented):
Some questions :)

packages/cli-e2e/jest.config.ts
y-lakhdar: That's fine, but I still need to update the `testMatch` parameter or these non-E2E tests will be executed

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    afterAll(async () => {
+      await browser.close();
+      await teardownUIProject(cliProcesses);
+    }, 5e3);
+
+    it('should contain a search page section', async () => {
+      await page.goto(searchPageEndpoint, {
+        waitUntil: 'networkidle2',
```
olamothe, line 45: What is `networkidle2` ?

packages/cli-e2e/__tests__/angular.specs.ts
olamothe: Indeed, try to have one expect per test maximum (as a rule of thumb).

packages/cli/src/commands/ui/create/angular.ts
```diff
@@ -28,6 +31,10 @@ export default class Angular extends Command {
     const {args, flags} = this.parse(Angular);
     await this.createProject(args.name, flags.defaults);
     await this.addCoveoToProject(args.name, flags.defaults);
+    await this.config.runHook(
```
olamothe, line 34: 👍

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -0,0 +1,79 @@
...
+      expect(JSON.parse(await tokenResponse.text())).toMatchObject({
+        token: expect.stringMatching(/^eyJhb.+/),
+      });
+    });
+
+    it('should trigger search queries', async () => {
+      const searchboxSelector = 'app-search-page app-search-box input';
+      await page.goto(searchPageEndpoint, {waitUntil: 'networkidle2'});
+
+      // Request from interface load
```
olamothe, line 63: Similar to the comment Louis added, if this was split in two, the comment could be removed, since the expect would match exactly the test title.

olamothe (approved):

packages/cli-e2e/__tests__/angular.specs.ts
y-lakhdar: `networkidle2` - consider navigation to be finished when there are no more than 2 network connections for at least 500 ms.

packages/cli-e2e/__tests__/angular.specs.ts
y-lakhdar: 👍

y-lakhdar:
Need to fix first CDX-190 first for the tests to pass

y-lakhdar:
blocked by #111

PR #103, opened 2021-03-22
test(e2e): add screenshots

louis-bompart:
With that, we should have screenshot artifacts on the tests.

Jest has been quite disappointing here.  you cannot get nice/good information on what the current test is (name, status, etc).

So either I was taking the screenshot every time, or I was adding complexity for the screenshots.
Well, we're not charged for the artifact loading, so I chose the first option.

### DIFF
--- a/packages/cli-e2e/utils/browser.ts
+++ b/packages/cli-e2e/utils/browser.ts
@@ -65,6 +65,7 @@ export async function captureScreenshots(browser: Browser): Promise<void> {
       });
     } catch (error) {
       console.warn('Failed to record screenshot.');
+      console.warn(error);
     }
   }
 }
--- a/packages/cli-e2e/utils/browser.ts
+++ b/packages/cli-e2e/utils/browser.ts
@@ -61,11 +61,10 @@ export async function captureScreenshots(browser: Browser): Promise<void> {
       await page.screenshot({
         fullPage: true,
         type: 'png',
-        path:
-          resolve(
-            SCREENSHOTS_PATH,
-            expect.getState().currentTestName.trim().replace(/ /g, '_')
-          ) + '.png',
+        path: resolve(
+          SCREENSHOTS_PATH,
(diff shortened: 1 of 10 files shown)


louis-bompart:
(For now waiting to see how it's going in the CI 🍿)

packages/cli-e2e/utils/browser.ts
```diff
@@ -50,6 +54,21 @@ export async function getNewBrowser(): Promise<Browser> {
...
 }
 
+export async function captureScreenshots(browser: Browser): Promise<void> {
+  mkdirSync(SCREENSHOTS_PATH, {recursive: true});
+  for (const page of await browser.pages()) {
+    try {
+      await page.screenshot({
+        fullPage: true,
+        type: 'png',
+        path: resolve(SCREENSHOTS_PATH, expect.getState.name),
```
y-lakhdar, line 64: Should we clean the string before (remove dots for instance)? Maybe it's not important.

y-lakhdar:
Can we also get screenshots after the login flow?

y-lakhdar (approved):

packages/cli-e2e/utils/browser.ts
```diff
@@ -50,6 +54,25 @@ export async function getNewBrowser(): Promise<Browser> {
   });
 }
 
+export async function captureScreenshots(browser: Browser): Promise<void> {
+  mkdirSync(SCREENSHOTS_PATH, {recursive: true});
```
olamothe, line 58: Could add a preventive check on "folder does not already exists"

olamothe (approved):

olamothe:
So, where do we go exactly to see the screenshots ?

louis-bompart:
> So, where do we go exactly to see the screenshots ?
> 
> image
> 
> I'm seeing "no artifacts"
> 🤔

still debugging the CI 😅

PR #107, opened 2021-03-23
feat(login): improve feedback message on sucessful login 

olamothe:
Improve the "login message" on success, which displays current org, region, and environment.

Also add a check on "does the user has access to the org he specified" during login. 

Otherwise they will most probably get cryptic errors for other subsequent commands, when trying to generate an API key for an org that does not exists, or for which their user does not have access to, when they try to run the ui:* commands.

+ Some UTs

https://coveord.atlassian.net/browse/CDX-182

### DIFF
--- a/packages/cli/src/commands/auth/login.ts
+++ b/packages/cli/src/commands/auth/login.ts
@@ -49,6 +49,7 @@ export default class Login extends Command {
   async run() {
     await this.loginAndPersistToken();
     await this.persistRegionAndEnvironment();
+    await this.verifyOrganization();
     await this.persistOrganization();
     await this.feedbackOnSuccessfulLogin();
     this.config.runHook('analytics', buildAnalyticsSuccessHook(this, flags));
@@ -110,7 +111,7 @@ export default class Login extends Command {
       return;
     }
 
-    this.log('You have no access to any Coveo organization.');
+    this.error('You have no access to any Coveo organization!');
   }
 
   private async pickFirstAvailableOrganization() {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/auth/login.ts
```diff
@@ -116,4 +128,29 @@ export default class Login extends Command {
...
+      const found = orgs.find((o) => o.id === flags.organization);
+      if (!found) {
+        this.error(
+          `You either do not have access to organization ${flags.organization}, or it does not exists.`
+        );
+      }
+    }
+
+    if (orgs.length === 0) {
+      this.error('You do not have access to any Coveo organization.');
```
y-lakhdar, line 153: Maybe the user chose the wrong region or environment. So maybe we can say something like "no orgs available in this region/env"?

packages/cli/src/commands/auth/login.ts
```diff
@@ -116,4 +128,29 @@ export default class Login extends Command {
   private get configuration() {
     return new Config(this.config.configDir, this.error);
   }
+
+  private async getAllOrgsUserHasAccessTo() {
+    const orgs = await (
+      await new AuthenticatedClient().getClient()
+    ).organization.list();
+    return (orgs as unknown) as OrganizationModel[];
```
louis-bompart, line 136: You're sure about this, it doesn't look like `list()` returns this kind of array, at least directly.
Without checking out your branch, and by checking https://github.com/coveo/platform-client/blob/aeda85c635bb13c5de97bfc08ac8f2e7f08e4c30/src/resources/BaseInterfaces.ts,
I think that the `OrganizationModel[]` is in `.list().items` imo

I'll check out the branch and check later.

louis-bompart: image
Yep.

olamothe: Yeah, it's a bug in the platform client, if you ask for "all" orgs, non paginated, it returns a different type than if you ask for paginated call.

So in this case the typing is invalid.

It's something I had identified as a bugfix for platform client, but never got the time to submit a fix...

Yeah, my bad.

louis-bompart (approved):

packages/cli/src/commands/auth/login.ts
olamothe: @louis-bompart :

PR for platform-client here: https://github.com/coveo/platform-client/pull/274

PR #108, opened 2021-03-24
feat(oauth): remove keytar dependencies

olamothe:
Removed the `Storage` (OAuthStorage) module using keytar, and simply use the configuration for the access token obtained in the login flow.

https://coveord.atlassian.net/browse/CDX-193


packages/cli/src/commands/config/get.ts
```diff
@@ -10,6 +10,7 @@ export default class Get extends Command {
 
   async run() {
     const cfg = await new Config(this.config.configDir, this.error).get();
+    delete cfg.accessToken;
```
olamothe, line 13: Delete the access token from the config before printing it to console with `coveo config:get`

y-lakhdar: Does that mean we have to run `auth:login` after doing a `config:get` command?

olamothe: No, it just means we delete it from the object we use to display in the console.

It does not actually remove it from the config, on your disk.

y-lakhdar (approved):
Should the keytar dependencies be removed from the Dockerfile in this PR too?

louis-bompart (changes requested):
> Should the keytar dependencies be removed from the Dockerfile in this PR too?

This and :shipit: I'd say :D

packages/cli-e2e/package.json
```diff
@@ -20,16 +20,16 @@
     "test:headless": "node ./entrypoints/entry.js",
     "test:debug": "node ./entrypoints/entry.js --debug",
     "test:bash": "node ./entrypoints/entry.js --bash",
-    "jest": "jest",
+    "jest": "jest --verbose",
```
louis-bompart, line 23: Why verbose?

louis-bompart, line 23: Why verbose

louis-bompart (approved):
lgtm, not sure about verbose tho

packages/cli-e2e/package.json
olamothe: It's nicer to look at :D

PR #111, opened 2021-03-25
ci(githubactions): publish test versions for E2E tests

y-lakhdar:


.github/workflows/build.yml
```diff
@@ -23,9 +30,23 @@ jobs:
...
+      - name: Release test versions
+        run: |
+          echo "E2E tests should target template version ${{ env.UI_TEMPLATE_VERSION }}"
+          git config --global user.email action@github.com
+          git config --global user.name GitHub Action
+        # node_modules/.bin/lerna version ${{ env.UI_TEMPLATE_VERSION }} --no-changelog --no-push -y
+        # echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }} > ~/.npmrc"
+        # Uncomment when ready
+        # npm run npm:publish:template
+        # node scripts/wait-for-published-packages.js -v ${{ env.UI_TEMPLATE_VERSION }}
```
y-lakhdar, line 45: I need to make sure the action works as expected before uncommenting these lines

scripts/get-ui-templates.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * All the UI templates used by 3rd-party CLIs.
+ *
+ */
+const getUiTemplates = () => [
+  '@coveo/cra-template',
+  '@coveo/vue-cli-plugin-typescript',
+  '@coveo/angular',
+];
+
```
y-lakhdar, line 10: It's the only place where we have to hardcode the ui templates that we need to publish for  E2E tests.

scripts/wait-for-published-packages.js
```diff
@@ -0,0 +1,35 @@
...
+
+async function waitForPackage(packageName, version) {
+  const opts = {
+    resources: [`https://www.npmjs.com/package/${packageName}/v/${version}`],
+    delay: 1e3, // initial delay in ms
+    interval: 5e3, // poll interval in ms
+    timeout: 5 * 60 * 1e3, // timeout in ms
+    window: 1e3, // stabilization time in ms
+    log: true,
+  };
```
y-lakhdar, line 15: Waiting for the packages to be ready. I don't know if a timeout of 5 minutes is enough.

scripts/get-ui-templates-for-lerna.js
```diff
@@ -0,0 +1,8 @@
+const {getUiTemplates} = require('./get-ui-templates');
+
+const getUiTemplatesForLerna = () =>
+  getUiTemplates()
+    .map((t) => `--scope ${t}`)
+    .join(' ');
+
+console.log(getUiTemplatesForLerna());
```
y-lakhdar, line 8: That way, we only publish ui templates.

Another option would be to add a `npm:publish:template` script to template packages only. So we don't rely on the `--scope` option.

scripts/get-next-test-version.js
```diff
@@ -0,0 +1,60 @@
...
+
+function getGreatestVersion(versions = []) {
+  const validVersions = versions.filter((v) => valid(v));
+  const reducer = (previousVersion, currentVersion) =>
+    gte(currentVersion, previousVersion) ? currentVersion : previousVersion;
+  return validVersions.reduce(reducer);
+}
+
+async function main() {
+  const templates = getUiTemplates();
```
y-lakhdar, line 51: Generates versions with this format 0.0.x.

.github/workflows/build.yml
louis-bompart: Is this PR ready for review? 🤔

y-lakhdar: yes. I need your input before uncommenting these lines.

scripts/ui-template-utils.js
```diff
@@ -0,0 +1,55 @@
...
+    'show',
+    `${packageName}@0.0.*`,
+    'version',
+    '--json',
+  ]);
+
+  spawnProcess.stdout.on('data', (data) => {
+    const allVersions = data.toString().split('\n');
+    lastVersion = getLastValidVersion(allVersions);
+  });
```
louis-bompart, line 45: I think spawnSync and parsing afterward would be simpler.

louis-bompart (commented):
looks good to test 👍

scripts/ui-template-utils.js
y-lakhdar: I did it that way because of the `Promise.all()`.
so we can get the version of all packages in "parallel"

.github/workflows/build.yml
```diff
@@ -23,9 +30,23 @@ jobs:
...
         run: tsc -b
+      - name: Set the template version
+        run: |
+          echo "UI_TEMPLATE_VERSION=$(node scripts/get-next-test-version.js)" >> $GITHUB_ENV
+      - name: Release test versions
+        run: |
+          echo "E2E tests should target template version ${{ env.UI_TEMPLATE_VERSION }}"
+          git config --global user.email action@github.com
+          git config --global user.name GitHub Action
+        # node_modules/.bin/lerna version ${{ env.UI_TEMPLATE_VERSION }} --no-changelog --no-push -y
```
y-lakhdar, line 41: @louis-bompart 
I don't think this line will work because it will only bump packages that have changed since the last release.
The goal is to bump all the templates to the same 0.0.x version, so we can pass that version to the E2E tests.

Should we instead just use `npm bump` instead of lerna?

y-lakhdar: or maybe `lerna version --force-publish=<package-1>,<package-2>,...`?

olamothe (approved):

PR #115, opened 2021-03-26
fix(angular): fix search-token-server path

louis-bompart:
See the issue here: https://github.com/angular/angular-cli/issues/17921
Tho, the `files://` did work for me.

I invite you to checkout the branch and test to ensures it still works in an UNIX environment (no reason why tho, the `files://` is supposed to works everywhere :tm:.

----
https://coveord.atlassian.net/browse/CDX-192


y-lakhdar (approved):
Works for me.

eng-dibo:
nice: thanks for investigating the issue <3

PR #121, opened 2021-03-30
chore(cli): refactor checks into decorators

louis-bompart:
- Split the Node and NPX checks from the React command and into preconditions
- Refactor the Authentication decorator into a precondition
- Added a Preconditions checker decorator.

Unit Tested 💯 
image

----

CDX-203


packages/cli/package.json
```diff
@@ -19,6 +19,7 @@
     "cli-ux": "^5.5.1",
     "coveo.analytics": "^2.18.4",
     "create-react-app": "^4.0.3",
+    "dedent": "^0.7.0",
```
louis-bompart, line 22: Neat library to avoid oopsies in indentation with template string
https://www.npmjs.com/package/dedent

packages/cli/package.json
```diff
@@ -32,15 +33,17 @@
     "@coveo/vue-cli-plugin-typescript": "^0.7.0",
     "@oclif/dev-cli": "^1.26.0",
     "@oclif/test": "^1",
+    "@types/dedent": "^0.7.0",
     "@types/fs-extra": "^9.0.6",
+    "@types/jest": "^26.0.22",
```
louis-bompart, line 38: To have describe in here.

packages/cli/package.json
```diff
@@ -32,15 +33,17 @@
...
     "@types/node": "^10",
     "chai": "^4.2.0",
     "jest": "^26.6.3",
     "patch-package": "^6.2.2",
     "rimraf": "^3.0.2",
     "ts-jest": "^26.5.1",
     "ts-node": "^8",
-    "typescript": "^3.3"
+    "typescript": "4.2.3"
```
louis-bompart, line 46: Could be reverted:
I wanted to migrate the tsconfig to a composite project to split the test tsconfig from the source tsconfig (to avoid having describe everywhere), but was unable because of https://github.com/oclif/oclif/issues/299.
Oclif went hay-wire when I tried to split it.

packages/cli/src/commands/auth/login.spec.ts
```diff
@@ -75,12 +75,12 @@ describe('auth:login', () => {
   test
     .command(['auth:login', '-e', 'foo'])
     .catch(/Expected --environment=foo/)
-    .it('reject invalid environment');
+    .it('reject invalid environment', async () => {});
```
louis-bompart, line 78: I noted timing issue and promise timing out here.
I'm honestly not too sure of what's going in in this tests, the syntax is really painful to read.

packages/cli/src/commands/ui/create/react.ts
```diff
@@ -5,25 +5,23 @@ import {
...
+  IsNodeVersionAbove,
+  IsNpxInstalled,
+} from '../../../lib/decorators/preconditions';
 
 export default class React extends Command {
   static templateName = '@coveo/cra-template';
+  static requiredNodeVersion = '10.16.0';
 
-  static description = `Create a Coveo Headless-powered search page with the React web framework. See ${linkToReadme}`;
+  static description =
```
louis-bompart, line 23: Because the page about the command was really shallow after reworking it, I removed the link.

packages/cli/src/hooks/init/set-global-config.ts
```diff
@@ -1,13 +1,4 @@
-/* eslint-disable @typescript-eslint/no-namespace */
-import {Hook, IConfig} from '@oclif/config';
-
-declare global {
-  namespace NodeJS {
-    interface Global {
-      config: IConfig;
-    }
-  }
-}
```
louis-bompart, line 10: I put the typing shenanigan in a declaration file to keep things tidy.

packages/cli/src/lib/decorators/preconditions/node.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+      };
+
+      await expect(
+        IsNodeVersionAbove('foo')((fakeCommand as unknown) as Command)
+      ).resolves.toBe(false);
+      expect(fakeCommand.warn).toHaveBeenCalledTimes(1);
+      expect(fakeCommand.warn).toHaveBeenCalledWith(dedent`
+        Required version invalid: "foo".
+        Please report this error to Coveo: https://github.com/coveo/cli/issues/new
+      `);
```
louis-bompart, line 30: See below for more explanation ;)

packages/cli/src/lib/decorators/preconditions/node.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+    });
+
+    it('should return false and warn', async () => {
+      const fakeCommand = {
+        id: 'foo',
+        warn: jest.fn(),
+      };
+
+      await expect(
+        IsNodeVersionAbove('0.0.1')((fakeCommand as unknown) as Command)
```
louis-bompart, line 50: I prolly could do a bit cleaner than the double cast with some Jest-pirouette, but I didn't think it was worth the hassle.

packages/cli/src/lib/decorators/preconditions/node.ts
```diff
@@ -0,0 +1,64 @@
...
+const linkToReadme =
+  'https://github.com/coveo/cli/wiki/Node.js,-NPM-and-NPX-requirements';
+
+export function IsNodeVersionAbove(requiredVersion: string) {
+  return async function (target: Command) {
+    if (!isVersionValid(requiredVersion)) {
+      target.warn(dedent`
+        Required version invalid: "${requiredVersion}".
+        Please report this error to Coveo: https://github.com/coveo/cli/issues/new
+      `);
```
louis-bompart, line 18: If this warning is triggered, there's a big chance that it's not the user machine that ain't properly set up, but that we messed up and didn't call properly the function.

Hopefully, this code will never run, but if it does, at least the user will know what to do (contact us so we can fix it).

packages/cli/src/lib/decorators/preconditions/node.ts
```diff
@@ -0,0 +1,64 @@
...
+    Please visit ${linkToReadme} for more detailed installation information.
+  `);
+}
+
+export function isMissingExecutable(output: SpawnProcessOutput) {
+  return (
+    output.exitCode !== null &&
+    Math.abs(output.exitCode) === constants.errno.ENOENT
+  );
+}
```
louis-bompart, line 64: Will be refactorized on a subsequent PR as utils, for now, that was not worth it. (when introducing decorator for Angular).

packages/cli/src/lib/decorators/preconditions/preconditions.ts
```diff
@@ -0,0 +1,23 @@
+import type Command from '@oclif/command';
+
+export type PreconditionFunction = (
+  target: Command
+) => Boolean | Promise<Boolean>;
+
+export function Preconditions(...preconditions: PreconditionFunction[]) {
+  return function (
```
louis-bompart, line 8: Fun anecdote: don't put async here, typings get messy.

packages/cli/src/lib/decorators/preconditions/node.ts
y-lakhdar: It wouldn't hurt to have this check. 👍

packages/cli/src/lib/decorators/preconditions/preconditions.ts
```diff
@@ -0,0 +1,23 @@
...
+    target: Command,
+    _propertyKey: string,
+    descriptor: TypedPropertyDescriptor<() => Promise<void>>
+  ) {
+    const originalRunCommand = descriptor.value!;
+    descriptor.value = async function () {
+      for (const precondition of preconditions) {
+        if (!(await precondition(target))) {
+          return;
```
y-lakhdar, line 17: Should we add some kind of message/logging here?

packages/cli/src/lib/decorators/preconditions/preconditions.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+
+      expect(orderChecker).toHaveBeenCalledTimes(preconditions.length + 1);
+      for (let index = 0; index < preconditions.length; index++) {
+        expect(orderChecker).toHaveBeenNthCalledWith(index + 1, index);
+      }
+      expect(fakeOriginalFunction).toHaveBeenCalled();
+      expect(orderChecker).toHaveBeenNthCalledWith(
+        preconditions.length + 1,
+        'final'
+      );
```
y-lakhdar, line 46: I've heard somewhere that we should aim for one expect per test maximum.
I'm kidding. I understand the difference here :P
👌

y-lakhdar (commented):
Nice Job

packages/cli/src/lib/decorators/preconditions/npx.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      const fakeCommand = {
+        id: 'foo',
+        warn: jest.fn(),
+      };
+
+      await expect(
+        IsNpxInstalled()((fakeCommand as unknown) as Command)
+      ).resolves.toBe(false);
+      expect(fakeCommand.warn).toHaveBeenCalledTimes(2);
+      expect(fakeCommand.warn).toHaveBeenCalledWith(dedent`
```
olamothe, line 65: I'm not sure I would have a test that verifies the "exact same message contained in the command, just copy pasted". 

Whenever we want to change a comma for the warning message, the tests will start to fail for no "good" reason.

I would focus on regex matching a couple important part: the readme link, a "high level check" on some important part of the message (eg: `/some random error oh no/` or `/a valid npx installation/`, stuff like this).

The exact wording itself does not seem very important to validate with a UT.

PR #122, opened 2021-03-30
chore(cli): refactor checks into decorators

louis-bompart:
- Split the Node and NPX checks from the React command and into preconditions
- Refactor the Authentication decorator into a precondition
- Added a Preconditions checker decorator.

Unit Tested 💯 
image

(Reopened from #121, TIL renaming a branch in GitHub delete it)

----

CDX-203


packages/cli/src/commands/ui/create/react.ts
```diff
@@ -42,15 +40,13 @@ export default class React extends Command {
     {name: 'name', description: 'The target application name.', required: true},
   ];
 
-  @AuthenticationRequired()
+  @Preconditions(
+    IsAuthenticated(),
+    IsNodeVersionAbove(React.requiredNodeVersion),
+    IsNpxInstalled()
+  )
```
y-lakhdar, line 47: Nice!

packages/cli/src/lib/decorators/preconditions/node.ts
```diff
@@ -0,0 +1,64 @@
...
+const linkToReadme =
+  'https://github.com/coveo/cli/wiki/Node.js,-NPM-and-NPX-requirements';
+
+export function IsNodeVersionAbove(requiredVersion: string) {
+  return async function (target: Command) {
+    if (!isVersionValid(requiredVersion)) {
+      target.warn(dedent`
+        Required version invalid: "${requiredVersion}".
+        Please report this error to Coveo: https://github.com/coveo/cli/issues/new
+      `);
```
y-lakhdar, line 18: Good. It wouldn't hurt to have this check.

packages/cli/src/lib/decorators/preconditions/preconditions.ts
```diff
@@ -0,0 +1,23 @@
...
+  return function (
+    target: Command,
+    _propertyKey: string,
+    descriptor: TypedPropertyDescriptor<() => Promise<void>>
+  ) {
+    const originalRunCommand = descriptor.value!;
+    descriptor.value = async function () {
+      for (const precondition of preconditions) {
+        if (!(await precondition(target))) {
+          return;
```
y-lakhdar, line 17: Should we add some kind of message/logging here?

y-lakhdar (approved):
Nice work

olamothe (approved):
I've left all my comments on the other PR :) 

https://github.com/coveo/cli/pull/121

packages/cli/src/lib/decorators/preconditions/preconditions.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+describe('preconditions', () => {
+  const preconditions = new Array<jest.Mock<Boolean>>(5);
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('if all preconditions succeed', () => {
+    it('should executes preconditions the original command', async () => {
+      let counter = 0;
+      const orderChecker = jest.fn();
```
louis-bompart, line 13: > I see you put quite a lot of emphasis in the tests on "the order is respected".

> Is that really an important/meaningful thing to test ?

> I guess it can be, if we have preconditions that depends on one another, but right now it does not seem to be the case that we have such a need.

>The most important thing to verify is that the original function is only called after all preconditions had a chance to execute.

> Maybe we could simplify the test to only focus on this type of verification?
@olamothe , https://github.com/coveo/cli/pull/121#discussion_r604172048

At the first I thought about not caring for the order (I mean, I could have do a `Promise.all(preconditions.map(precondition(target))` instead of the foreach that'd be prettier and neatier).

But then, as you said, I guess it is not crazy to think that we might want to have preconditions depending on one another. So I decided to care about the order and to enforce it. This goes by the test as well.

I think we should not simplify the test if it means they describe less accurately what the function is doing.

I know the unit test may seem daunting here or even overkill, however, I decided to do so because it was in my opinion the best balance between cost and risk:
 - Risk is very high IMO, because this function is bound to be ubiquitous in the code.
 - To cover such risk, we need to have a high confidence that the function is working properly. To that effect, the most efficient way was to unit test it the best I could.
 
 However, if you think the unit test itself is unclear, it may mean I need to reorganize the code and split it better so its more meaningful. If you do think so, could you give me some pointers?

packages/cli/src/lib/decorators/preconditions/preconditions.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      expect(fakeOriginalFunction).toHaveBeenCalled();
+      expect(orderChecker).toHaveBeenNthCalledWith(
+        preconditions.length + 1,
+        'final'
+      );
+    });
+  });
+
+  describe('if one precondition failed', () => {
+    it('should executes all the preconditions prior to the failing one and nothing else', async () => {
```
louis-bompart, line 51: > Similar comment as above, is that really something that we are interested in ?
>
>If all preconditions are independant from one another, the order is not actually important, just that the "command" does not get executed if at least one precondition fails..

@olamothe, https://github.com/coveo/cli/pull/121#discussion_r604173775

See https://github.com/coveo/cli/pull/122/files#r604263433

packages/cli/src/lib/decorators/preconditions/npx.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    });
+
+    it('should return false and warn', async () => {
+      const fakeCommand = {
+        id: 'foo',
+        warn: jest.fn(),
+      };
+
+      await expect(
+        IsNpxInstalled()((fakeCommand as unknown) as Command)
```
louis-bompart, line 33: > For readability's sake, I would add a function getFakeCommand(): Command {...} util function that does the cast + boilerplate.
>
> That you could reuse for (I think?) all the tests in this file.

@olamothe, https://github.com/coveo/cli/pull/121#discussion_r604177075

Good idea! 👍 

- [ ] Wrap `fakeCommand` as an util/factory

packages/cli/src/lib/decorators/preconditions/npx.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      ).resolves.toBe(false);
+      expect(fakeCommand.warn).toHaveBeenCalledTimes(2);
+      expect(fakeCommand.warn).toHaveBeenCalledWith(dedent`
```
louis-bompart, line 65: > I'm not sure I would have a test that verifies the "exact same message contained in the command, just copy pasted".
>
> Whenever we want to change a comma for the warning message, the tests will start to fail for no "good" reason.
>
> I would focus on regex matching a couple important part: the readme link, a "high level check" on some important part of the message (eg: /some random error oh no/ or /a valid npx installation/, stuff like this).
>
> The exact wording itself does not seem very important to validate with a UT.

@olamothe , https://github.com/coveo/cli/pull/121#discussion_r604182036

That would be true for end-to-end tests, but it is not for unit test imo.
Unit tests IMO must be as 'strict' as possible.

> Whenever we want to change a comma for the warning message, the tests will start to fail for no "good" reason.
That's technically a change of behavior, and it must be tested, as all other are. Do you prefer an automated test, or that a QA specialist spend time checking it? I for sure know that I'd prefer the former.

PR #123, opened 2021-03-30
feat(vue): add precheck for node version on `ui:create:vue`

louis-bompart:
Implementation of the refacto from #122.

Only decision made here was where to put the line on the Node version.
Vue does support as low as 8.x for Node.JS ([source](https://cli.vuejs.org/guide/installation.html))
However, their alpha version has already dropped it, and also dropped 10.x ([source](https://github.com/vuejs/vue-cli/blob/dev/CHANGELOG.md#500-alpha5-2021-02-23))

Therefore, Node 12 seems OK till next year. (all non LTS has been dropped by Vue in v5 too)

----

CDX-198

### DIFF
--- a/packages/cli/src/commands/ui/create/vue.ts
+++ b/packages/cli/src/commands/ui/create/vue.ts
@@ -8,6 +8,7 @@ import {Config} from '../../../lib/config/config';
 import {
   Preconditions,
   IsAuthenticated,
+  IsNodeVersionAbove,
 } from '../../../lib/decorators/preconditions';
 import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
 import {platformUrl} from '../../../lib/platform/environment';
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/ui/create/vue.ts
```diff
@@ -16,7 +17,7 @@ import {getPackageVersion} from '../../../lib/utils/misc';
 
 export default class Vue extends Command {
   static templateName = '@coveo/vue-cli-plugin-typescript';
-
+  static requiredNodeVersion = '12.21.0';
```
olamothe, line 20: I would add (here + for the react command I guess), some link to their documentation that explains this choice.

So that the next guy that needs to modify this knows "where to look for".

olamothe (approved):

packages/cli/src/commands/ui/create/vue.ts
louis-bompart: - [ ] Add doc link to Vue requirements

y-lakhdar (approved):

PR #128, opened 2021-03-30
feat(config): improve configuration validation on invalid org id

olamothe:
Started as a story to add UT for `config:set` command.

Ended up doing a bit more work internally: 

* Bumped platform-client to latest version, to get fixes related to types on `organization.list()`
* Reworked AuthenticatedClient to add two methods: 
  * `getAllOrgsUserHasAccessTo(): Promise<OrganizationModels[]>`
  * `getUserHasAccessToOrg(org): Promise<boolean>`
* Added new UT for `config:set`
* Added a check on `config:set -o invalid_org_id` to block writing an invalid org to config.
* Some internal rework on login command to leverage the changes to `AuthenticatedClient`
* Fix UT for login command

https://coveord.atlassian.net/browse/CDX-208


packages/cli/src/commands/config/set.ts
```diff
@@ -44,13 +46,15 @@ export default class Set extends Command {
     }),
   };
 
+  @AuthenticationRequired()
```
olamothe, line 49: To be able to list the organization the users has access to, you need to be logged in.

louis-bompart: Tho will need to be updated when #122 will be merged

packages/cli/src/lib/platform/authenticatedClient.ts
```diff
@@ -61,6 +61,17 @@ export class AuthenticatedClient {
...
+
+  async getAllOrgsUserHasAccessTo() {
+    const platformClient = await this.getClient();
+    return platformClient.organization.list();
+  }
+
+  async getUserHasAccessToOrg(org: string) {
+    const orgs = await this.getAllOrgsUserHasAccessTo();
+    const found = orgs.find((o) => o.id === org);
+    return found !== undefined;
```
louis-bompart, line 73: Either:
```suggestion
    const hasAccess = orgs.some((o) => o.id === org);
    return hasAccess;
```

louis-bompart, line 73: Or:
```suggestion
    return orgs.some((o) => o.id === org);
```

louis-bompart (approved):
LGTM, some changes to be more semantically correct in the `getUserHasAccessToOrg` func.

PR #134, opened 2021-03-31
test(e2e): Add local npm registry for testing

louis-bompart:
Based on @y-lakhdar work on #111, I restructured the test to works with https://verdaccio.org/.

Also include a shift from 'simple' docker to docker-compose: it allows to keep the docker configurations tidy and is a first step toward the better test caching.

Some modification was required in the vue specs:
Vue uses yarn, and yarn use per-package configuration, meaning the global npm registry configured was ignored.
To fix this, we have to create a .yarnrc before vue-cli actually installs stuff.
This requires new 'input hoops' in the CI (mostly, press down arrow key and enter)

Some improvement were made too to increase the overall speed of the test by removing arbitrary waits.

----
CDX-212


scripts/ui-template-utils.js
```diff
@@ -15,41 +12,16 @@ const getUiTemplates = () => [
...
-}
-
-function getPackageLastTestVersion(packageName) {
-  let lastVersion = '';
-  const spawnProcess = spawn('npm', [
+function getCiTestPackageVersion(packageName) {
+  const ciTestVersion = spawnSync('npm', [
     'show',
-    `${packageName}@0.0.*`,
+    `${packageName}@ci-test`,
```
y-lakhdar, line 18: Oh. I did not think about that. Nice!

.github/workflows/build.yml
```diff
@@ -24,26 +24,9 @@ jobs:
       - name: Build CLI
         working-directory: ./packages/cli
         run: tsc -b
-      - name: Set the template version
-        run: |
-          node scripts/store-next-test-version.js --variable=UI_TEMPLATE_VERSION --output=$GITHUB_ENV
-      - name: Bump template versions
-        run: |
-          echo "E2E tests should target template version ${{ env.UI_TEMPLATE_VERSION }}"
-          echo "//registry.npmjs.org/:_authToken=${{ env.NPM_TOKEN }}" > ~/.npmrc
```
y-lakhdar, line 33: You can also remove the `NPM_TOKEN` from the env section

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -1,13 +1,20 @@
...
+  isGenericYesNoPrompt,
+  setupUIProject,
+  teardownUIProject,
+} from '../utils/cli';
 import {getNewBrowser} from '../utils/browser';
 import {isSearchRequest} from '../utils/platform';
+import {EOL} from 'os';
 
-describe('ui', () => {
+describe.skip('ui', () => {
```
y-lakhdar, line 17: Did you forget to remove the skip?

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -29,8 +38,52 @@ describe('ui', () => {
...
+      cliProcesses.push(buildProcess);
+
+      buildProcess.stdout.on('data', async (data) => {
+        if (
+          /Pick an action: \(Use arrow keys\)/.test(stripAnsi(data.toString()))
+        ) {
+          await answerPrompt('\u001b[B', buildProcess);
+          return;
+        }
+        if (/❯ Merge/.test(stripAnsi(data.toString()))) {
```
y-lakhdar, line 57: I'm not sure about this for a couple of reasons:
* I think we should always overwrite the project. Merging could lead to unpredicted results
* Instead of expecting a bunch of prompts, we could simply test the default flow with the flags `--default` and `--force` using the passthrough strategy

Also, I did not include these prompts in the tests because I assumed the container running the E2E tests would always be initialized to a clean state.

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -29,8 +38,52 @@ describe('ui', () => {
...
+
+      const startServerProcess = spawn('npm', ['run', 'start'], {
+        cwd: projectPath,
+        detached: true,
+      });
+
+      cliProcesses.push(startServerProcess);
+      await new Promise<void>((resolve) => {
+        startServerProcess.stdout.on('data', async (data) => {
+          if (stripAnsi(data.toString()).indexOf('App running at:') !== -1) {
```
y-lakhdar, line 81: Good idea. That will speed up the tests instead of adding an arbitrary timeout

packages/cli-e2e/__tests__/vue.specs.ts
louis-bompart: What's going on with the Vue tests here:
 1. We create the directory where the project
 2. We write a `.yarnrc` that tells Yarn to uses Verdaccio
 
 Given those two, if we do not merge we will discard the `.yarnrc` and thus vue-cli won't use our registry.
 
 Therefore, without changing the behavior of any command, this is the only solution I found which is 'sound'
 (I could probably also spoof npm at the DNS level but... it's a bit more involved imo)

packages/cli-e2e/__tests__/angular.specs.ts
louis-bompart: I'm stabilizing the vue test and then will stabilize the angular ones.

packages/cli-e2e/docker/config/config.yaml
```diff
@@ -0,0 +1,84 @@
+#
+# This is the config file used for the docker images.
```
olamothe, line 2: I would cleanup the generic verdacio comments/commented out code. 

I assume you copied from their documentation ? Maybe a link would be better. 

And then, I would keep just the things that we actually use in this config, with the important/specific stuff that we need commented.

olamothe (approved):

olamothe:
:gg: That's going to save us a lot of troubles !

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -19,16 +27,59 @@ describe('ui', () => {
     let interceptedRequests: HTTPRequest[] = [];
     let page: Page;
 
+    // TODO: Refactor.
```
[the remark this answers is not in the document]
louis-bompart: Yeah the PR is still very rough around the edges, I'm still working on cleaning this up.

- [ ] refactor openPage utils that is shared across vue, react and angular specs

packages/cli-e2e/__tests__/react.specs.ts
```diff
@@ -21,9 +21,9 @@ describe('ui', () => {
 
     beforeAll(async () => {
       browser = await getNewBrowser();
-      await setupUIProject('ui:create:react', projectName, cliProcesses, {
-        timeout: 30e3,
-      });
+      // await setupUIProject('ui:create:react', projectName, cliProcesses, {
```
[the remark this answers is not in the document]
louis-bompart: I guess this is an 'old' comment, 'cause I uncommented

packages/angular/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Coveo schematic for angular",
   "scripts": {
     "test": "npm run build && jasmine dist/**/*_spec.js",
-    "build": "tsc -p tsconfig.json",
+    "build": "tsc -p tsconfig.schematic.json",
```
louis-bompart, line 7: - [ ] Remove after merging #135

packages/cli-e2e/__tests__/vue.specs.ts
louis-bompart: - [ ] Double check if we can do without the local `.yarnrc`

y-lakhdar (approved):

louis-bompart:
- [ ] Ensure tests fails when they fails (exit code is not correct)

PR #135, opened 2021-04-02
fix(angular): prevent d.ts files from being copied to project

y-lakhdar:
This only happens when testing the schematic locally without going through the one published in npm.

https://coveord.atlassian.net/browse/CDX-221


louis-bompart (approved):
Seems legit 👍

PR #136, opened 2021-04-02
fix(server): always return a JSON object

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-220


louis-bompart (approved):

olamothe (approved):

y-lakhdar:
> Mhhh... I would _maybe_ type check err.
> If it's an (error)[https://nodejs.org/api/errors.html#errors_class_error], maybe just add a statusCode and bubble it right up.
> 
> The Oopsy something brokey is always a bit frustrating tbh

Do you want to bubble up only the error message or the entire object (with the error stack)?
It's not like we are only returning an Oopsy message without no other message to the frontend. The error error stack trace will be printed in the terminal.

louis-bompart:
> > Mhhh... I would _maybe_ type check err.
> > If it's an (error)[https://nodejs.org/api/errors.html#errors_class_error], maybe just add a statusCode and bubble it right up.
> > The Oopsy something brokey is always a bit frustrating tbh
> 
> Do you want to bubble up only the error message or the entire object (with the error stack)?
> It's not like we are only returning an Oopsy message without no other message to the frontend. The error error stack trace will be printed in the terminal.

I think it's OK as it stands, I don't know what the error could be.
If we find out that there's too man oopsy message, we'll improve.

PR #137, opened 2021-04-02
feat(angular): redirect user to an error page if the `.env` file is invalid

y-lakhdar:


packages/angular/src/ng-add-setup-project/files/src/app/error/error.component.html
```diff
@@ -0,0 +1,19 @@
+<section class="warning-bg full-height center">
+  <div class="inner-container">
+    <p class="mat-display-1 mb-5">Invalid Environment variables</p>
```
louis-bompart, line 3: Either
```suggestion
    <p class="mat-display-1 mb-5">Invalid Environment Variables</p>
```
Or
```suggestion
    <p class="mat-display-1 mb-5">Invalid environment ariables</p>
```

louis-bompart (approved):
LGTM!

PR #139, opened 2021-04-08
Create codeql-analysis.yml

louis-bompart:
image
eh why not, let's see what gives.


louis-bompart:
No bandwith to finish that now-now, will revisit

PR #141, opened 2021-04-08
feat(angular): add node and npm preconditions

louis-bompart:
- Refactor precondition factories<sup>2</sup> to DRY
- Added NPM precondition factory.
- Added NPM and Node precondition factory on test.
- Uniformize messages.

-------

https://coveord.atlassian.net/browse/CDX-195


packages/cli/src/lib/decorators/preconditions/binPreconditionsFactory.ts
```diff
@@ -0,0 +1,127 @@
...
+      const output = await spawnProcessOutput(
+        binaryName,
+        appliedOptions.params
+      );
+
+      if (!isBinInstalled(target, binaryName, appliedOptions, output)) {
+        return false;
+      }
+
+      if (!satisfies(output.stdout, versionRange)) {
```
y-lakhdar, line 49: Do you need to strip the `v` prefix before doing the check first?
image

y-lakhdar (approved):

packages/cli/src/lib/decorators/preconditions/binPreconditionsFactory.ts
louis-bompart: nope: 
> A leading "=" or "v" character is stripped off and ignored.

https://www.npmjs.com/package/semver

PR #142, opened 2021-04-08
feat(cli): use API key in `.env` file

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-223


packages/cli-e2e/utils/cli.ts
```diff
@@ -67,4 +74,41 @@ export function setupUIProject(
...
 
+export function getConfig() {
+  const pathToConfig = resolve(
+    __dirname,
+    ...new Array(4).fill('..'),
+    '.config',
+    '@coveo',
+    'cli',
+    'config.json'
+  );
```
louis-bompart, line 85: What is the path you're trying to reach here?
(please put it in absolute :P) 

I have the feeling this is what causing issue in the E2E.

packages/cli/src/commands/ui/create/vue.ts
```diff
@@ -91,11 +91,11 @@ export default class Vue extends Command {
       '--orgId',
       cfg.organization,
       '--apiKey',
-      cfg.accessToken!,
+      apiKey.value!,
       '--platformUrl',
       platformUrl({environment: cfg.environment}),
       '--user',
-      providerUsername,
+      userInfo.providerUsername,
```
louis-bompart, line 98: Why `userInfo.providerUsername` and not `userInfo` (I don't see what changed to justify it) ?

packages/cli/src/commands/ui/create/vue.ts
```diff
@@ -80,9 +80,9 @@ export default class Vue extends Command {
 
   private async invokePlugin(applicationName: string) {
     const cfg = await this.configuration.get();
-    const {providerUsername} = await this.getUserInfo();
+    const {userInfo, apiKey} = await this.platformUserCredentials();
```
louis-bompart, line 83: `getUserInfo()` is now orphaned, should be deleted

louis-bompart (changes requested):
Tests are red and some tweak here and there's.
(Also axios is really nicer than the isomorphic thingies imho.)

packages/cli-e2e/utils/cli.ts
```diff
@@ -51,7 +57,8 @@ export function setupUIProject(
   options: ISetupUIProjectOptionsArgs = {}
 ) {
   const versionToTest = process.env.UI_TEMPLATE_VERSION;
-  let command = [commandArgs, projectName, ...(options.flags || [])];
+  const uniqueProjectName = `${process.env.GITHUB_ACTION}-${projectName}`;
+  let command = [commandArgs, uniqueProjectName, ...(options.flags || [])];
```
louis-bompart, line 61: Why would you need that?
This btw make your test fails.

packages/cli/src/commands/ui/create/vue.ts
y-lakhdar: Because of the method `platformUserCredentials` it returns a `{UserModel, ApiKeyModel}`

packages/cli/src/commands/ui/create/vue.ts
y-lakhdar: oups. forgot to delete it. thanks

packages/cli/src/commands/ui/create/vue.ts
louis-bompart: Oh I get it!
I missed the deconstruction that was happening before with `getUserInfo`
i.e. I read ` const {providerUsername} = await this.getUserInfo();` as ` const providerUsername = await this.getUserInfo();`

Which confused me when checking your changeset.

My bad!

packages/cli-e2e/utils/cli.ts
```diff
@@ -1,3 +1,7 @@
+require('isomorphic-fetch');
+require('abortcontroller-polyfill');
+require('isomorphic-form-data');
```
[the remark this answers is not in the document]
louis-bompart: - [ ] Axios

packages/cli/src/commands/ui/create/react.ts
```diff
@@ -123,11 +123,25 @@ export default class React extends Command {
...
     const authenticatedClient = new AuthenticatedClient();
     const platformClient = await authenticatedClient.getClient();
     await platformClient.initialize();
 
-    return await platformClient.user.get();
+    const userInfo = await platformClient.user.get();
+    const apiKey = await authenticatedClient.createImpersonateApiKey(args.name);
+
+    return {userInfo, apiKey};
+  }
```
louis-bompart, line 136: This method could be a function if it took an ApiKeyName param.

If so, you could refactor it and avoid repetition in several places.

louis-bompart (changes requested):
Two things mainly:
 - [x] Use Axios to have less dep and because it's more 'mainstream'
 - [ ] Refactor the platform clients func.

packages/cli-e2e/setup.ts
```diff
@@ -22,6 +23,8 @@ async function clearChromeBrowsingData(browser: Browser) {
 
 export default async function () {
   mkdirSync(SCREENSHOTS_PATH, {recursive: true});
+  process.env.GITHUB_ACTION =
+    process.env.GITHUB_ACTION || randomBytes(16).toString('hex');
```
y-lakhdar, line 27: @louis-bompart I think we will end up with issues if we don't convert to alphanumerical characters only.

louis-bompart: True, good catch!

packages/cli/src/lib/platform/authenticatedClient.spec.ts
```diff
@@ -102,4 +106,19 @@ describe('AuthenticatedClient', () => {
       AuthenticationStatus.LOGGED_IN
     );
   });
+
+  it('should create an API key with impersonate privileges', async () => {
```
louis-bompart, line 110: I think you should follow the naming convention used in the file, meaning:
```
it(`#funcName should...`)
```
So:
```suggestion
  it('#createImpersonateApiKey should create an API key with impersonate privileges', async () => {
```

louis-bompart (approved):
Apart from my suggestion, LGTM 🥇 !

PR #143, opened 2021-04-08
chore(cli): lint and enforce lint

louis-bompart:
https://coveord.atlassian.net/browse/CDX-226


packages/cli/src/commands/auth/login.ts
```diff
@@ -6,7 +6,6 @@ import {
   PlatformEnvironment,
   PlatformRegion,
 } from '../../lib/platform/environment';
-import {OrganizationModel} from '@coveord/platform-client';
```
louis-bompart, line 9: unused

packages/cli/src/commands/org/list.ts
```diff
@@ -1,6 +1,5 @@
 import {Command} from '@oclif/command';
 import {AuthenticatedClient} from '../../lib/platform/authenticatedClient';
-import {OrganizationModel} from '@coveord/platform-client';
```
louis-bompart, line 3: unused

y-lakhdar (approved):

PR #144, opened 2021-04-08
Fixing tests for CDX-223

louis-bompart:
Proposed changed for @y-lakhdar concerning #142.
No change in logic, just the tests


packages/cli-e2e/utils/cli.ts
```diff
@@ -76,8 +75,7 @@ export function setupUIProject(
 
 export function getConfig() {
   const pathToConfig = resolve(
-    __dirname,
-    ...new Array(4).fill('..'),
+    homedir(),
```
louis-bompart, line 78: IMO, if you go 'beyond' the root folder with `..` that's risky. so let's use homedir instead.

packages/cli-e2e/entrypoints/common.js
```diff
@@ -62,21 +63,22 @@ const ensureDockerImageIsPresent = () => {
...
     '.env',
-    environmentVariables
-      .map((variable) => `${variable}=${process.env[variable]}`)
-      .join('\n')
+    [
+      ...environmentVariables.map(
+        (variable) => `${variable}=${process.env[variable]}`
+      ),
+      `TEST_RUN_ID=${
+        process.env.GITHUB_ACTION || randomBytes(16).toString('hex')
```
louis-bompart, line 79: githubaction wont exist on your machine.  and you don't want to collide with other peeps running test on their machine too.
So if github_action ain't available, let's generate a pseudo-uuid and use it instead.

packages/cli-e2e/__tests__/react.specs.ts
```diff
@@ -40,8 +40,9 @@ describe('ui', () => {
       });
       await new Promise<void>((resolve) => {
         startServerProcess.stdout.on('data', async (data) => {
+          console.log(data.toString());
           if (
-            /You can now view react-project in the browser/.test(
+            /You can now view .*-react-project in the browser/.test(
```
louis-bompart, line 45: We need to match the projectId.

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -121,16 +125,18 @@ describe('ui', () => {
 
     it('should send a search query when the page is loaded', async () => {
       await page.goto(searchPageEndpoint, {waitUntil: 'networkidle2'});
+      await page.waitForSelector(searchboxSelector);
```
louis-bompart, line 128: So, I noticed that sometimes, the expectation are run too soon (i.e. the project javascript didn't have time to run).
To avoid that, I put a waitFor on the searchboxSelector.

This allows us to ensure that the scripts has been loaded etc.
Worst case, it didn't, and the expectation will still faceplant.

y-lakhdar (approved):

PR #147, opened 2021-04-12
chore(vue): lint and enforce on ci

louis-bompart:
note: do not fix the linting issue for the end-user, rules will need to be adapted as such when it'll be done.

----

https://coveord.atlassian.net/browse/CDX-231


packages/vue-cli-plugin-typescript/package.json
```diff
@@ -11,6 +11,7 @@
   "scripts": {
     "npm:publish": "npm publish --access public",
     "npm:publish:template": "npm publish --access public --tag ci-test",
-    "npm:bump:template": "npm --no-git-tag-version version"
+    "npm:bump:template": "npm --no-git-tag-version version",
+    "lint": "prettier --config ../../.prettierrc.js --check . && eslint ."
```
y-lakhdar, line 15: Can't we use the prettierrc.js in the same folder (`packages/vue-cli-plugin-typescript/.prettierrc.js `)?

y-lakhdar (approved):

PR #148, opened 2021-04-12
chore(angular): lint and enforce on ci

louis-bompart:
https://coveord.atlassian.net/browse/CDX-232


packages/angular/src/ng-add-setup-project/files/src/app/engine.service.ts
```diff
@@ -6,7 +6,7 @@ import {environment} from '../environments/environment';
   providedIn: 'root',
 })
 export class EngineService {
-  private engine: any;
+  private engine!: HeadlessEngine<typeof searchAppReducers>;
```
y-lakhdar, line 9: 👍

y-lakhdar (approved):

PR #149, opened 2021-04-13
feat(vue): remove eslint warnings 

y-lakhdar:
A couple changes:
1. Removed all the missing return type warnings
2. Refactored the `EngineService` to get the right typing definition (instead of `HeadlessEngine<any>`) ([commit](https://github.com/coveo/cli/commit/be27ab9264a68f45a9d6eb37a9beb24fff2b457c))
3. Updated Vue component with Typescript syntaxe


packages/vue-cli-plugin-typescript/generator/template/src/components/Facet.vue
```diff
@@ -13,33 +13,47 @@
...
   },
   methods: {
-    onToggle: function (facetValue) {
+    onToggle: function (facetValue: HeadlessFacetValue) {
       this.facet.toggleSelect(facetValue);
     },
   },
   created: function () {
     this.facet = buildFacet(this.engine, {
-      options: {field: this.field, title: this.title},
```
y-lakhdar, line 36: this one was a bug.
Thanks to the type def, I got an error saying that title was not part of the option interface

packages/vue-cli-plugin-typescript/generator/template/src/EngineService.ts
```diff
@@ -1,47 +1,25 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
```
louis-bompart, line 1: Why, precisely?
It _can_ be the solution for 'Removed all the missing return type warnings', I get that, but that's the last resort solution.
To approves that I need to know exactly what prevents to just comply with this rule.

packages/vue-cli-plugin-typescript/generator/template/src/components/Facet.vue
```diff
@@ -13,33 +13,47 @@
...
+  props: {
+    field: String,
+    title: String,
+  },
   components: {FacetValue},
   data: function () {
     return {
       state: {},
-    };
+    } as IFacet;
```
louis-bompart, line 41: This applies for all components:
I think you should not force-cast your object as such. That means that the interface is not doing its job properly.
The `facet` (or whatever for the other, you get the gist) should be optional and then either:
- If there's a chance that `facet` is null somewhere in your code then check it before using it.
- If you're absolutely positive that it's assigned, then you can use [non-null-assertion-operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator) to prevent the 'oh but this can be undefined too duh' error.

packages/vue-cli-plugin-typescript/generator/template/src/router/index.ts
```diff
@@ -25,11 +25,16 @@ const routes: Array<RouteConfig> = [
       if (!isEnvValid()) {
         next('/error');
       } else {
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         const res = await fetch(process.env.VUE_APP_TOKEN_ENDPOINT!);
```
louis-bompart, line 29: putting an eslint-disable is a last resort solution always.
We should comply with the rule imposed on us to the best of our ability.

See my comment on envUtils for guidance ;)

packages/vue-cli-plugin-typescript/generator/template/src/utils/envUtils.ts
```diff
@@ -3,7 +3,7 @@
  *
  * @returns true if the .env file is valid. false otherwise.
  */
-export function isEnvValid() {
+export function isEnvValid(): boolean {
```
louis-bompart, line 6: Check the https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

It should allow the typescript compiler to understand that "after isEnv is used, process.env has those keys"

louis-bompart (changes requested):
I'll get a better look 🔜, but I'm really not comfortable with the eslint-disable and eslint-ignore.

packages/vue-cli-plugin-typescript/generator/template/src/EngineService.ts
y-lakhdar: I also don't like this approach as well, but I made sure to minimize the amount of eslint-disable rules.
In this case, the `getEngine` function has the following return type: `HeadlessEngine<ReducersMapObject<SearchAppState, Action<any>>>`. 
However, `ReducersMapObject` is not exposed in Coveo Headless, hence the `ReturnType<typeof getEngine>`.

packages/vue-cli-plugin-typescript/generator/template/src/components/Facet.vue
y-lakhdar: mmh. I'm not sure how I can benefit from the typing definition without force-casting the return of the `data` function. Without the `IFacet` interface, I get typing errors even tho `this.facet` exists. `this.facet` is never null when I need to access it because it is defined in the `created` method.
image

louis-bompart:
Superseded by #150

PR #150, opened 2021-04-13
feat(cli): fix linting in vue project and rework headless initialization

louis-bompart:
OG PR is #149 differences are:
 - Rebased on #147
 - 038406a 🥖commit
 
# The 🥖commit:
- Add vetur.config.js to allow the use of vetur in a mono repo context (see https://vuejs.github.io/vetur/guide/setup.html#advanced)
- Disabled node/no-unpublished-import for Vue: We need to import devDeps for typechecking because the template is installing them on the fly. (already done on angular and react)
- Removed declaration: Cannot override typing of Vue $root (or at least I didn't find how)
- Morphed `isEnvValid` into a type-guard
- Smite all non-null assertion operators 🔨 
- And err the big one: Reworked all comps': they now instantiate their state in the data, because it makes so much more sense considering Vue life-cycle.
  To that end: the global mixin now instead does a `beforeCreate` that writes in the data of the root component. Inspired from https://vuejs.org/v2/guide/state-management.html#Simple-State-Management-from-Scratch
- Validate type check, both in our and in the end-user code (i.e. our linter is happy, and Vue's one happy too)

---- 
https://coveord.atlassian.net/browse/CDX-180


packages/vue-cli-plugin-typescript/generator/template/src/components/FacetValue.vue
```diff
@@ -7,16 +7,18 @@
...
   </li>
 </template>
-<script>
-export default {
+<script lang="ts">
+import Vue from 'vue';
+
+export default Vue.extend({
   name: 'FacetValue',
   props: ['facetValue'],
```
y-lakhdar, line 15: It's a shame we can't have prop types for custom object...

packages/vue-cli-plugin-typescript/generator/template/src/router/index.ts
```diff
@@ -22,14 +22,18 @@ const routes: Array<RouteConfig> = [
...
       } else {
-        const res = await fetch(process.env.VUE_APP_TOKEN_ENDPOINT!);
+        const res = await fetch(process.env.VUE_APP_TOKEN_ENDPOINT);
         const {token} = await res.json();
-        const engineService = new EngineService(token);
         // Adding Coveo Headless engine as a global mixin so it can be available to all components
-        Vue.mixin(engineService.mixin);
+        const engine = getEngine(token);
+        Vue.mixin({
+          beforeCreate: function () {
```
y-lakhdar, line 33: Thanks for finding that

y-lakhdar (approved):

PR #161, opened 2021-04-16
fix(cli): allow update to run without the analytics setting being set

louis-bompart:
This was blocking the auto-update on the first run 😢 

---

https://coveord.atlassian.net/browse/CDX-243

### DIFF
--- a/packages/cli/src/hooks/prerun/prerun.spec.ts
+++ b/packages/cli/src/hooks/prerun/prerun.spec.ts
@@ -33,14 +33,33 @@ describe('hooks:prerun', () => {
 
   test
     .do(() => {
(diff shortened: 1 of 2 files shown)


olamothe (approved):

olamothe:
But: if you ran `coveo update`, I'm guessing it would ask you the question to enable/disable analytics, no ? 

And then, if you accept/deny, calling coveo update will work afterwards ?

Was there another problem, if you say it "blocked", meaning it crashed or something ?

olamothe (commented):
But: if you ran coveo update, I'm guessing it would ask you the question to enable/disable analytics, no ?

And then, after you accept/refuse, calling coveo update will work afterwards.

So I'm not sure why that is a problem exactly ? Or was it crashing somehow ?

PR #162, opened 2021-04-16
doc(cli): readme-review

jpmarceau:
Je trouvais que le readme principal manquait un peu de viande, alors j'en ai ajouté.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Install via npm:
 $ npm install -g @coveo/cli
 ```
 
-In practice, you'll typically want to [`login`](#coveo-authlogin) to your Coveo Organization, [`configure`](#coveo-configset) the CLI to connect to this organization, and finally create a search page (see [`coveo ui:create:angular NAME`](#coveo-uicreateangular-name), [`coveo ui:create:react NAME`](#coveo-uicreatereact-name) and [`coveo ui:create:vue NAME`](#coveo-uicreatevue-name)).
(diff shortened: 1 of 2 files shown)


olamothe:
This is auto generated. 
This means it's going to get overwritten all the time.

jpmarceau:
> This is auto generated.
> This sadly means it's going to get overwritten all the time.

Ho noes, that's unfortunate.
I see where the docstrings for the `Commands` section are extracted, but I can't seem to find how the `Usage` section is generated. Do you know of any simple way I could tweak it? I think an introductory paragraph or two would really help, but it's not a huge deal either. I want to add basically the same info to the readme at the root of the repo anyways.

louis-bompart:
> > This is auto generated.
> > This sadly means it's going to get overwritten all the time.
> 
> Ho noes, that's unfortunate.
> I see where the docstrings for the `Commands` section are extracted, but I can't seem to find how the `Usage` section is generated. Do you know of any simple way I could tweak it? I think an introductory paragraph or two would really help, but it's not a huge deal either. I want to add basically the same info to the readme at the root of the repo anyways.

We can investigate how to customize this message, but that'd prolly require some leg work.
In the meantime, I invite you to not alter this ReadMe.md and works on the 'root' one ;)

I logged a spike to check what's up CDX-254. However I don't think it'll be our priority soon 😅

louis-bompart (changes requested):
As explained above, bring the changes to the root readme instead.
We will tackle this one later.

louis-bompart (approved):

PR #163, opened 2021-04-18
ci(release): ensure release commit includes proper package-lock.json

louis-bompart:
Lerna usually symlinks leaf repositories and thus does not 'install' them per se. So for example, this is why there's no `@coveo/search-token-server` in the `@coveo/angular` package, despite the former being a dependency of the latter.

This causes issues for Snyk because it expect a package-lock.json coherent with the package.json

This PR satisfies this need by ensuring that the release commit is made after an `npm i` is done in every leaf packages that needs it (so far `angular` et `cli`).

----

CDX-248

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,6 @@
     "release:changelog": "node ./scripts/release-changelog.js",
     "npm:publish:template": "lerna run --no-bail npm:publish:template",
     "npm:bump:template": "lerna run --no-bail npm:bump:template",
-    "npm:genPkgLock": "lerna run --no-bail npm:genPkgLock"
+    "npm:genPkgLock": "lerna exec npm i"
   }
 }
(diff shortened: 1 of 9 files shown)


y-lakhdar (approved):

package.json
```diff
@@ -74,8 +76,8 @@
     "commit": "git-cz",
     "pr:report": "node ./scripts/pr-bot.js",
     "release:changelog": "node ./scripts/release-changelog.js",
-    "npm:publish": "lerna run --no-bail npm:publish",
     "npm:publish:template": "lerna run --no-bail npm:publish:template",
-    "npm:bump:template": "lerna run --no-bail npm:bump:template"
+    "npm:bump:template": "lerna run --no-bail npm:bump:template",
+    "npm:genPkgLock": "lerna exec npm i"
```
olamothe, line 81: Small pet peeve: but lowercase `npm:pkglock` to fit with other scripts

olamothe (approved):

PR #166, opened 2021-04-19
feat(vue): speed up project creation

y-lakhdar:
**Faster project generation**
All the project configuration is now defined inside the preset.
There is no need to execute an additional invoke command to fetch the Coveo plugin from npm and executing it.

All the plugins will be fetched at the same time (reducing the number of `npm install`s)

**Missing Readme**
Added a README file on the Coveo Vue plugin repo.

https://coveord.atlassian.net/browse/CDX-189


louis-bompart (commented):
LGTM. Approval held till 1.0.2 released ;)

louis-bompart (approved):

PR #175, opened 2021-04-21
ci(release): copy binary to 'latest'

louis-bompart:
Copy binary with a predictable, version-free name.
Example: `coveo-v1.0.1-x64.` ➡️  `coveo-latest-x64.exe`
Example 2: `coveo-v123.456.789-42424242-blablablablaextension.exe` ➡️  `coveo-latest-blablablablaextension.exe`

https://coveord.atlassian.net/browse/CDX-275


louis-bompart:
Testing in CI hang in there, that's why the code is rotten in build-binaries.yml

louis-bompart:
Test run results, see artifact: https://github.com/coveo/cli/actions/runs/772083242
image
Screenshot of the zip

.github/workflows/build-binaries.yml
```diff
@@ -46,10 +46,12 @@ jobs:
         run: ./node_modules/@oclif/dev-cli/bin/run pack${{ matrix.pack-command }}
       - name: Create packages
         working-directory: ./packages/cli
+        run: ./node_modules/@oclif/dev-cli/bin/run pack -t="${{ matrix.package-targets }}"
```
olamothe, line 49: I think it would be better to do this only in Jenkins actually.

You should see in `Jenkinsfile` a step to `Download release assets`, which then copies the assets in an `artefacts` folder for the dpeloyment pipeline.

We could modify the existing `download-release-assets` script to make it so it also create a `latest` copy.

After all, we only need a copy for S3/CDN, not for github release assets.

louis-bompart: Initially, I wanted to keep the complexity out of Jenkins but yeah it kinda makes sense because it'd affect the user in the repo. (i.e. much files, so confused, wow)

y-lakhdar (approved):

PR #176, opened 2021-04-22
chore(cli): add a method on AuthenticatedClient to get user info

y-lakhdar:
Just to prevent redoing the same logic on ui:create commands

### DIFF
--- a/packages/cli/src/lib/platform/authenticatedClient.spec.ts
+++ b/packages/cli/src/lib/platform/authenticatedClient.spec.ts
@@ -29,6 +29,9 @@ describe('AuthenticatedClient', () => {
 
   const mockInitialize = jest.fn();
   const mockCreate = jest.fn();
+  const mockGetUser = jest
+    .fn()
+    .mockImplementation(() => ({providerUsername: 'bob@coveo.com'}));
 
   mockConfig.mockImplementation(
     () =>
@@ -42,6 +45,9 @@ describe('AuthenticatedClient', () => {
         initialize: mockInitialize,
         apiKey: {
           create: mockCreate as unknown,
+        },
+        user: {
+          get: mockGetUser,
         },
       } as unknown) as PlatformClient)
   );
@@ -120,5 +126,11 @@ describe('AuthenticatedClient', () => {
         ],
       })
     );
+  });
+
+  it('#getUserInfo should return the user info', async () => {
+    const userInfo = await new AuthenticatedClient().getUserInfo();
+    expect(mockGetUser).toHaveBeenCalled();
+    expect(userInfo).toStrictEqual({providerUsername: 'bob@coveo.com'});
   });
 });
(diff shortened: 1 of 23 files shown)


packages/cli/src/lib/platform/authenticatedClient.ts
```diff
@@ -65,6 +65,17 @@ export class AuthenticatedClient {
...
+  async platformUserCredentials(name: string) {
+    const authenticatedClient = new AuthenticatedClient();
+    const platformClient = await authenticatedClient.getClient();
+    await platformClient.initialize();
+
+    const userInfo = await platformClient.user.get();
+    const apiKey = await authenticatedClient.createImpersonateApiKey(name);
+
+    return {userInfo, apiKey};
+  }
```
louis-bompart, line 78: Do you think this UT-able? it'd be great tbh

louis-bompart (commented):
Can you write UT for the method while at it?
That'd make the code even more reliable :D

y-lakhdar:
I changed the function so it only does one thing : return the user info. 
I think it was weird for the function to return the user info and also create an API ([commit](https://github.com/coveo/cli/pull/176/commits/77e964e10be44b49d2bfc7e71a7d0b0add4f9d15)).

Also added UT

packages/cli/src/lib/platform/authenticatedClient.ts
```diff
@@ -65,6 +65,14 @@ export class AuthenticatedClient {
       ],
     });
   }
+
+  async getUserInfo() {
+    const authenticatedClient = new AuthenticatedClient();
+    const platformClient = await authenticatedClient.getClient();
+    await platformClient.initialize();
+
+    return await platformClient.user.get();
```
olamothe, line 74: the `await` is not necessary here since we return immediately anyway.

packages/cli/src/lib/platform/authenticatedClient.ts
```diff
@@ -65,6 +65,14 @@ export class AuthenticatedClient {
       ],
     });
   }
+
+  async getUserInfo() {
+    const authenticatedClient = new AuthenticatedClient();
+    const platformClient = await authenticatedClient.getClient();
+    await platformClient.initialize();
```
olamothe, line 72: Why do we need to initialize here ?

packages/cli/src/lib/platform/authenticatedClient.spec.ts
```diff
@@ -121,4 +127,10 @@ describe('AuthenticatedClient', () => {
       })
     );
   });
+
+  it('#getUserInfo should return the user info', async () => {
+    const userInfo = await new AuthenticatedClient().getUserInfo();
```
olamothe, line 132: Related to my comment below, if `initialize` is needed on the platform client in order to correctly receive the user info, then this is what the UT should focus on.

Because, otherwise, this test is kind of useless, we mock the return and just check that it returns what we just mocked ;)

olamothe (approved):

packages/cli/src/lib/platform/authenticatedClient.ts
y-lakhdar: Because, if I don't pass any argument to the `user.get()` method, it tries to load the currentUser (which does not exist at that moment)
https://github.com/coveo/platform-client/blob/master/src/resources/Users/User.ts#L8

PR #178, opened 2021-04-22
feat(cli): display "Happy Hacking" message at the end of the project creation

y-lakhdar:
image

* Making sure the "Happy Hacking" message is displayed at the complete end of the project creation
* Added a spinner for the token server creation. Otherwise there's a couple of seconds without any feedback

### DIFF
--- a/packages/cli/src/commands/ui/create/react.ts
+++ b/packages/cli/src/commands/ui/create/react.ts
@@ -78,7 +78,7 @@ export default class React extends Command {
     throw err;
   }
 
-  private createProject(name: string) {
+  private async createProject(name: string) {
     const flags = this.flags;
     const templateVersion =
       flags.version || getPackageVersion(React.templateName);
@@ -86,7 +86,13 @@ export default class React extends Command {
       name,
       '--template',
       `${React.templateName}@${templateVersion}`,
-    ]);
+    ]).catch((_e) =>
+      Promise.reject(
+        new Error(
+          'create-react-app is not able to create the project. See the logs above for more information.'
+        )
+      )
+    );
   }
 
   private async setupEnvironmentVariables(name: string) {
@@ -151,7 +157,7 @@ export default class React extends Command {
   }
 
   private async runReactCliCommand(commandArgs: string[], options = {}) {
-    return new Promise<string>((resolve) => {
+    return new Promise<string>((resolve, reject) => {
       const child = spawnProcessPTY(
         appendCmdIfWindows`npx`,
         ['create-react-app'].concat([...commandArgs]),
@@ -174,7 +180,10 @@ export default class React extends Command {
         }
       });
 
-      child.onExit(() => {
+      child.onExit(({exitCode}) => {
+        if (exitCode) {
+          reject();
+        }
         resolve(remainingString);
       });
     });
--- a/packages/cli/src/commands/ui/create/react.ts
+++ b/packages/cli/src/commands/ui/create/react.ts
@@ -154,7 +154,7 @@ export default class React extends Command {
     return new Promise<string>((resolve) => {
       const child = spawnProcessPTY(
         appendCmdIfWindows`npx`,
-        ['create-react-app'].concat([...commandArgs, '--color=always']),
+        ['create-react-app'].concat([...commandArgs]),
         options
       );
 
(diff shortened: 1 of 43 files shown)


packages/cli/src/commands/ui/create/react.ts
```diff
@@ -16,6 +17,7 @@ import {
   IsNpxInstalled,
 } from '../../../lib/decorators/preconditions';
 import {appendCmdIfWindows} from '../../../lib/utils/os';
+import {EOL} from 'os';
```
louis-bompart, line 20: -  [ ] Test on Windows. (Me)

packages/cli/src/commands/ui/create/react.ts
```diff
@@ -54,9 +56,16 @@ export default class React extends Command {
...
 
-    await this.createProject(args.name);
+    const finalOutput = await this.createProject(args.name);
+
+    cli.action.start('Creating search token server');
     await this.setupServer(args.name);
     await this.setupEnvironmentVariables(args.name);
+    cli.action.stop();
+
+    console.log(EOL);
```
louis-bompart, line 66: I'd use the Oclif log to be safe
```suggestion
    this.log(EOL);
```

packages/cli/src/commands/ui/create/react.ts
```diff
@@ -141,12 +150,34 @@ export default class React extends Command {
...
+    return new Promise<string>((resolve) => {
+      const child = spawnProcessPTY(
+        appendCmdIfWindows`npx`,
+        ['create-react-app'].concat([...commandArgs, '--color=always']),
+        options
+      );
+
+      const args = this.args;
+      let stopWritingInTerminal = false;
+      let remainingString = '';
```
louis-bompart, line 163: I think that if you create a buffer and write in it,  it might works without `--color=always` and be more efficient :)

packages/cli/src/lib/utils/process.ts
```diff
@@ -73,3 +74,18 @@ export async function spawnProcessOutput(
...
+export function spawnProcessPTY(
+  command: string,
+  args: string[],
+  options: SpawnOptions = {}
+) {
+  const ptyProcess = ptySpawn(command, args, {
+    name: 'xterm-color',
+    cols: process.stdout.columns,
+    rows: process.stdout.rows,
+    cwd: options.cwd || process.cwd(),
```
louis-bompart, line 87: Use `??` instead of `||`. `cwd:''` is valid IMO

louis-bompart (commented):
Can you try the buffer? I'm curious ;)
Also, I think we could improve the E2E for react in that sense to detect the 'Happy HAcking!' as a exit signal, as we did for the others.

Apart from that great job!

Tho, PSA y'all:
⚠️ microsoft/node-pty is a node native dependency ⚠️ 
We should build and test the binary.

packages/cli/src/commands/ui/create/react.ts
y-lakhdar: oups. The `--color-always` was added as an initial tentative to preserve the colors. Now that we use node-pty, I don't need it.

packages/cli/src/lib/utils/process.ts
y-lakhdar: good catch

y-lakhdar:
Looks like it broke the E2E tests.

- [ ] Update E2E tests
- [ ] Build and test node-pty native dependency

louis-bompart:
> Looks like it broke the E2E tests.
> 
> * [ ]  Update E2E tests
> * [ ]  Build and test node-pty native dependency

Just to be clear about the `node-pty` I'm not saying we should built it beforehand ourselves on our machine. This is the CD job.
However we ought to test it ;)

louis-bompart:
I'm building binaries from this branch to ensures all's well:
https://github.com/coveo/cli/actions/runs/785699814

louis-bompart:
image
Can you compare the behavior between current release and the new one?
`coveo ui:create:react reactTestPty`
made this one

louis-bompart:
When using a valid process name, the terminal/stdin is not released :/

louis-bompart (changes requested):
Required changes:
 - [ ] When create-react-app fails, we should promptly exit.
 - [x] When the command complete, it should release the prompt.

louis-bompart (approved):

olamothe (approved):

PR #179, opened 2021-04-22
ci(release): fix release commit message

louis-bompart:
Git commit does not interpret back-slashed escaped characters nor does it has an option for this.
However, you can use stdin for the message using `--file` flag with `-` for value

Echo allows back-slash interpretation with `-e`.
Thus: echo the message, pipe the echo in the git commit with `--file -` 🎉

Relevant links:
 - [Git-SCM commit documentation](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--Fltfilegt)
 - [Bash documentation](https://ss64.com/bash/echo.html)
-------

https://coveord.atlassian.net/browse/CDX-276

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,7 +45,7 @@ jobs:
         run: node ./scripts/get-version.js
       - name: Commit changes, tag and push
         run: |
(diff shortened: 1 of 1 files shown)


louis-bompart:
Tested on https://github.com/coveo/cli/tree/CDX-276-test with .github/release.yml

y-lakhdar:
nice

y-lakhdar (approved):

PR #180, opened 2021-04-23
doc(vue): review readme

jpmarceau:
I moved the readme to the root folder of the package, and added documentation on how to setup.


louis-bompart:
The content is good!
However, their two readme needs to consider:
 1. The one you wrote, for the npm package, which is visible on npmjs.com/@coveo/vue-cli-plugin-typescript for example
 2. The one you deleted, which is put in the generated project.
 
 Both are required, so you should delete none ;)

louis-bompart (changes requested):
Restore the deleted readme.

packages/vue-cli-plugin-typescript/generator/template/README.md
```diff
@@ -1,8 +1,8 @@
-# <%- rootOptions.projectName %>
+# @coveo/vue-cli-plugin-typescript
```
louis-bompart, line 1: You should keep the `rootOptions.projectName` it will be substituted by the end user's project name when the end-user uses the template.

louis-bompart (changes requested):
Almost there!

packages/vue-cli-plugin-typescript/generator/template/README.md
jpmarceau: ah that makes sense, thanks!

louis-bompart (approved):

PR #181, opened 2021-04-26
use winpty always + catch err

louis-bompart:
For Windows, it seems that `conPty` is not very stable yet, enforcing the usage of `winPty` fixes the console not being released.

We need to handle the rejection of the pty with the code (no stderr: () and process it properly for all OSes.


louis-bompart:
Uh. Looks the test went haywire. Checking...

louis-bompart:
fixed 🏁

PR #187, opened 2021-04-28
test(token-serves): add unit test

y-lakhdar:
I am using [supertest](https://www.npmjs.com/package/supertest) to test the express server. It works well with Jest.
However, it became a little bit trickier when I had to test the `ensureTokenGenerated` middleware ([code](https://github.com/coveo/cli/compare/CDX-236?expand=1#diff-5a35da0527d25e537cf650cda7af14d7ff8cad0e9284978e41cdd6bbcac09296R3)) . 

Since we need the server state to reset before some tests without mocking the middleware, I inserted a test middleware to clear the [app locals properties](http://expressjs.com/en/api.html#app.locals) when needed.

I also split the server.ts file into 2 files:
* `app.ts`: express server definition
* `server.ts`: contains the code to start the server

### DIFF
--- a/packages/search-token-server/app.ts
+++ b/packages/search-token-server/app.ts
@@ -1,18 +1,9 @@
-/* eslint-disable @typescript-eslint/no-namespace */
 import * as express from 'express';
 import {ensureTokenGenerated} from './middlewares/searchToken';
 import {errorHandler} from './middlewares/errorHandler';
 import {environmentCheck} from './middlewares/environmentCheck';
 
 const app = express();
-
-declare global {
-  namespace Express {
-    interface Request {
-      token: string;
-    }
-  }
-}
 
 app.use(express.json());
 
--- a/packages/search-token-server/app.ts
+++ b/packages/search-token-server/app.ts
@@ -1,0 +1,25 @@
+/* eslint-disable @typescript-eslint/no-namespace */
+import * as express from 'express';
+import {ensureTokenGenerated} from './middlewares/searchToken';
+import {errorHandler} from './middlewares/errorHandler';
+import {environmentCheck} from './middlewares/environmentCheck';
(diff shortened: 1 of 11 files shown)


packages/search-token-server/app.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const app = express();
+
+declare global {
+  namespace Express {
+    interface Request {
+      token: string;
+    }
+  }
+}
```
louis-bompart, line 15: I really don't like to mix the declaration with the code, can you split that into a declaration file?

louis-bompart, line 15: I really don't like to mix the declaration with the code, can you split that into a declaration file?

packages/search-token-server/app.ts
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable @typescript-eslint/no-namespace */
```
louis-bompart, line 1: If you apply my comment below, you should be able to remove that.

packages/search-token-server/server.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+
+  const triggerAMiddlewareError = () => {
+    mockedMiddleware.mockImplementationOnce(
+      (_req: Request, _res: Response, next: NextFunction) => {
+        const err: Error = new Error();
+        next(err);
+      }
+    );
+  };
+  const resetAppLocals = () => {
```
louis-bompart, line 52: I'm really splitting hair here, but stay consistent in your breakline :P 
```suggestion
  };

  const resetAppLocals = () => {
```

louis-bompart (approved):
👏 the tests are really crystal clear good job!

I would split the declaration tho from the actual script

PR #189, opened 2021-04-28
ci(release): extract fragment of release notes for github release

louis-bompart:
## Proposed changes

Split the changelog using a regex matching all `h1` and `h2` and extract the changeset on top.
Then, use the same process as before for publishing.

## Testing

Tested on my machine by altering the changelog.md and running the script (logging instead of publishing, ofc). I also checked what the RegExp was doing here: https://regex101.com/r/UitWby/1.

-----
CDX-185


olamothe (approved):

louis-bompart:
Blocked by #193 prolly

PR #191, opened 2021-04-29
test(cli): test 'bad environment' redirect

y-lakhdar:
## Testing

Adding E2E tests to handle cases where environment is invalid

-----
CDX-235

### DIFF
--- a/packages/cli-e2e/__tests__/angular.specs.ts
+++ b/packages/cli-e2e/__tests__/angular.specs.ts
@@ -13,92 +13,128 @@ import {isSearchRequest} from '../utils/platform';
 import {EOL} from 'os';
 import stripAnsi from 'strip-ansi';
 import {ProcessManager} from '../utils/processManager';
+import {join} from 'path';
+import {renameSync} from 'fs-extra';
 
-describe.skip('ui', () => {
-  describe('create:angular', () => {
-    let browser: Browser;
-    let processManager: ProcessManager;
-    // TODO: CDX-90: Assign a dynamic port for the search token server on all ui projects
-    const clientPort = '4200';
-    const projectName = `${process.env.GITHUB_ACTION}-angular-project`;
-    const searchPageEndpoint = `http://localhost:${clientPort}`;
-    const tokenProxyEndpoint = `http://localhost:${clientPort}/token`;
-    let interceptedRequests: HTTPRequest[] = [];
-    let page: Page;
+describe('ui:create:angular', () => {
+  let browser: Browser;
+  let buildProcessManager: ProcessManager;
+  let page: Page;
+  const oldEnv = process.env;
+  const projectName = `${process.env.GITHUB_ACTION}-angular-project`;
+  // TODO: CDX-90: Assign a dynamic port for the search token server on all ui projects
+  const clientPort = '4200';
+  const searchPageEndpoint = `http://localhost:${clientPort}`;
+  const tokenProxyEndpoint = `http://localhost:${clientPort}/token`;
 
-    const searchboxSelector = 'app-search-page app-search-box input';
+  const buildApplication = async (processManager: ProcessManager) => {
+    const proc = setupUIProject(
+      processManager,
+      'ui:create:angular',
+      projectName,
+      {
+        flags: ['--defaults'],
+      }
+    );
 
-    beforeAll(async () => {
-      processManager = new ProcessManager();
-      browser = await getNewBrowser();
-      const buildProcess = setupUIProject(
-        processManager,
-        'ui:create:angular',
-        projectName,
-        {
-          flags: ['--defaults'],
-        }
-      );
+    proc.stdout.on('data', async (data) => {
+      if (isGenericYesNoPrompt(data.toString())) {
+        await answerPrompt(`y${EOL}`, proc);
+      }
+    });
 
-      buildProcess.stdout.on('data', async (data) => {
-        if (isGenericYesNoPrompt(data.toString())) {
-          await answerPrompt(`y${EOL}`, buildProcess);
-        }
-      });
-
-      await Promise.race([
-        new Promise<void>((resolve) => {
-          buildProcess.on('exit', async () => {
-            resolve();
-          });
-        }),
-        new Promise<void>((resolve) => {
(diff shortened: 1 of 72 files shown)


packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -12,83 +12,121 @@ import {isSearchRequest} from '../utils/platform';
...
 import {EOL} from 'os';
 import {ProcessManager} from '../utils/processManager';
+import {join} from 'path';
+import {renameSync, existsSync} from 'fs-extra';
+
+describe('ui:create:vue', () => {
+  let processManager: ProcessManager;
+  let browser: Browser;
+  let page: Page;
+  const OLD_ENV = process.env;
```
olamothe, line 22: nit: Why the upper case naming here as opposed to other variables ?

olamothe (approved):

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -13,92 +13,128 @@ import {isSearchRequest} from '../utils/platform';
...
+      proc.stdout.on('data', async (data) => {
+        if (
+          /Compiled successfully/.test(
+            stripAnsi(data.toString()).replace(/\n/g, '')
+          )
+        ) {
+          resolve();
+        }
       });
-    }, 420e3);
```
louis-bompart, line 83: Beware of this one, default timeout is 5 seconds. Ain't sure that's gonna cut it.

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -13,92 +13,128 @@ import {isSearchRequest} from '../utils/platform';
...
+
+  const deactivateEnvironmentFile = () => {
+    const pathToEnv = getProjectPath(projectName);
+    renameSync(join(pathToEnv, '.env'), join(pathToEnv, '.env.disabled'));
+  };
+
+  const restoreEnvironmentFile = () => {
+    const pathToEnv = getProjectPath(projectName);
+    renameSync(join(pathToEnv, '.env.disabled'), join(pathToEnv, '.env'));
+  };
```
louis-bompart, line 92: Nice methods, can be refactored to be even neatier tho, see [that](https://www.typescriptlang.org/play?#code/PTAEFkHsGMGsEsB2BzU8C2AHSAnALqNAIaIDkBARgKagAWRAzgwDZUAmoA7vHraL0TwA6ALAAoNlWjMiOGtEiIGBOYiLoqAZQCeiaAC5QACknLDynEmQAaUADMckdObyWUASlABeAHygAbpDwbADc4pLSsvKKyqDIVHgACo4AVlJJgrSGRpip6QBy6lQubsievqAWVmESUjJyhDEEuZBp0HiFGiXV4XVRjUoEKUGI2ULjssgM3SgA2gC65X5VKOIgoACiiByQdvy0NOgwCKti4gqDoJJE7fD+guxb-gBi8KydNF6gpEJUiP5CNjwBhECisNikGoXWI3PB3KhPV6sNQabzfX7-SHic5NSqcIiYJ5oowOZyVVxWWx4SAzMrePwAb3EoBZA1imEyABVIESvvEknl2olMjlBR0iu4aqzQKoijo9EZhkgclyef9bKT3LYlYgVbxuU8qZB3JLxABfGo4y7XW73PAI-7wRyIDSIPBIz7GJZ4glPIyw+GIt5-Iq2G1wu2Pf4ej6ms5iaEqKjKXAO-xOxSu93B4nehj4wn-ExUAORthB96h0CltMelFUdxAA)

louis-bompart: Also, you should refactor that in an utils.

packages/cli-e2e/__tests__/react.specs.ts
```diff
@@ -7,80 +7,116 @@ import {captureScreenshots, getNewBrowser, openNewPage} from '../utils/browser';
...
+    page = await openNewPage(browser, page);
+  });
+
+  afterEach(async () => {
+    await captureScreenshots(browser);
+  });
+
+  afterAll(async () => {
+    process.env = oldEnv;
+    await buildProcessManager.killAllProcesses();
```
louis-bompart, line 93: I would:
 - Refactor `buildProcessManager` and `serverProcessManager` in a processManagers array.
 - Ensure that all processManagers are closed in the global afterAll
 - Kill all processes of the buildProcess at the end of the beforeAll. He doesn't make sense for it to be still running after the beforeAll.
 - Kill all processes of the server process in each afterAll sections (which I think you already do)
 
 The top-level afterAll is a catch 'em all clause in case something goes haywire, but at least we're sure of ourselves.
 The refactor into an array is to simplify adding other processes in the future.

louis-bompart (commented):
Very good job 👍
Some minor improvements to DRY the code but otherwise 💯

packages/cli-e2e/__tests__/react.specs.ts
y-lakhdar: I agree with the 2 last points. But not too much with the 2 firsts.
Killing multiple time the processes of the same ProcessManager will likely lead to `Error: kill ESRCH`.

louis-bompart: > Killing multiple times the processes of the same ProcessManager will likely lead to `Error: kill ESRCH`.

Hm, no actually.
The processManager unregister its registered processes as they exit, and the 'killing function' is awaitable (and should be awaited).
Therefore, if this happens, it means there's an issue either in the exiting of a processor in the processManager behaviour.

In either case, it would fail legitimately and should be 'fixed' and not ignored.

The most important thing in a test is to have a trustable setup phase (i.e. we know that the test environment is as we designed/thought it).
When you're not able to trust your setup phase, that means tests result could be due to unforeseen conditions that are not applicable for our end-user, leading to a false positive or false negative.

If that'd be the case, the tests would be worthless.

y-lakhdar: Fixed `killAllProcesses`.
It was not removing a process from the processes list once it was killed

louis-bompart (approved):

PR #192, opened 2021-04-30
test(e2e): terminal toolbelt for end-to-end test

louis-bompart:
## Proposed changes

Add a wrapper to handle the launch and handling of a 'CLI' process (so a process running in a console).
The wrapper goal is to simplify the specification of a condition using a 'When This Then That Until Thingy' (W3TUT... yea nah).

More concretely, the toolbelt allows command chaining to specify, in order:
 - the condition (regex or string)
 - the target of the condition (the process or one of its out-stream)
 - what to do when such condition is fulfilled.
 - when to stop checking for the condition (until).
 
This PR does not aim to be a complete toolbelt but the foundation to:
 - simplify the setup of our tests
 - build more tools to add to the toolbelts.

## Tests

Manually and CI, it's only test writing. A small bit of angular.specs.ts has been modified as proof of work.

-----
CDX-233


packages/cli-e2e/utils/terminal.ts
```diff
@@ -0,0 +1,176 @@
...
+class Action {
+  public parallelAction?: Action;
+  public sequentialAction?: Action;
+  public when?: When;
+  public on?: On;
+
+  constructor(public process: ChildProcessWithoutNullStreams) {}
+}
+
+type DoCallback = (
```
olamothe, line 39: personal pet peeves, but I generally prefer to see all the `types` / `interface` at the beginning of the file, grouped together.

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -65,20 +66,19 @@ describe('ui', () => {
...
-          if (
-            /Compiled successfully/.test(
-              stripAnsi(data.toString()).replace(/\n/g, '')
-            )
-          ) {
-            resolve();
-          }
-        });
+      await new Promise((resolve) => {
+        terminalServer
```
olamothe, line 78: The API is pretty clean/straightforward to understand 💯 

Proposition (feel free to reject):

Since the condition is always going to be performed `.on` a stream (stdout, stderr, unless I'm mistaken),
maybe we could just combine that and the `.on` becomes superfluous.

eg:

```
terminalserver
  .when('stdout', /Compiled successfully/)
  .do(resolve)
```

olamothe (approved):

packages/cli-e2e/__tests__/angular.specs.ts
louis-bompart: > The API is pretty clean/straightforward to understand 💯

Thanks! Hopefully this will ease the E2E testing process :D 

About your proposition, I thought about it and this is a conscious choice:
 - I think even if it's a tad bit more verbose, having one args per chaining func allow to have a verbose and clear understanding of what's going on.
 - A Condition can also be applied 'on metal', directly on the process.
 
 
 However, I agree that `on` will very often be `on('stdout')`. Which make it tedious. My plan in the future is that the grammar of the toolbelt (i.e. what can follow what in the func chaing) to be way more flexible.
Meaning that we'll have 'defaulting' on some part of the grammar: which will make this possible:

```typescript
// Will be the same...
terminalserver
  .when(/Compiled successfully/)
  .do(resolve);
// As this:
terminalserver
  .when(/Compiled successfully/)
  .on('stdout')
  .do(resolve);
```

packages/cli-e2e/utils/terminal.ts
louis-bompart: Yeah, for now I wanted to shoot the idea out there about the API etc, the code clearly need to be refined, I'm on it ;)

y-lakhdar (commented):
Well done!

y-lakhdar (approved):

olamothe (approved):

packages/cli-e2e/utils/terminal/do.ts
```diff
@@ -0,0 +1,165 @@
...
+        (data: Buffer) => {
+          if (stripAnsi(data.toString()).replace(/\n/g, '').match(condition)) {
+            callback(this.action.process, resolve);
+          }
+        },
+      ];
+    }
+
+    if (isConditionString(condition)) {
+      onParams = [condition, () => callback(this.action.process, resolve)];
```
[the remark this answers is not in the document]
louis-bompart: No because when a condition is a string, you listen to the event that is named as the condition, meaning that
```ts
new terminal(...spawnParams).when('potato').on('stdout').do(()=>console.log('foo!');
```

is the same as:
```ts
spawn(...spawnParams).stdout.on('potato',()=>console.log('foo!'));
```

and **not** the same as

```ts
spawn(...spawnParams)
  .stdout
  .on('data', () =>
     data.indexOf('potato')!==-1 ?
     console.log('foo!') :
     ()=>{}
  );
```

PR #193, opened 2021-05-02
fix(angular): fix Coveo headless error on application run

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-280

## Proposed changes

I updated the angular `tsconfig.json` file to include the `allowSyntheticDefaultImports` flag. Otherwise the `npm start` command leads to an error (see image). 
image

### Note:
All other templates (React and Vue) have the `allowSyntheticDefaultImports` flag in their `tsconfig.json`.

## Testing

This is already covered by current E2E tests.

-----
CDX-280

### DIFF
--- a/packages/angular/src/ng-add-setup-project/index.ts
+++ b/packages/angular/src/ng-add-setup-project/index.ts
@@ -10,6 +10,7 @@ import {
 } from '../common-rules/dependencies';
 import {updateNgModule} from './rules/ng-module';
 import {addMaterialAngular, addToPackageJson} from './rules/dependencies';
+import {updateTsConfig} from './rules/tsconfig';
 
 export default function (options: CoveoSchema): Rule {
   return async (tree: Tree) => {
@@ -20,6 +21,7 @@ export default function (options: CoveoSchema): Rule {
       addMaterialAngular(options),
       createFiles(options),
       updateNgModule(options, project),
+      updateTsConfig(options),
     ]);
   };
 }
--- a/packages/angular/src/ng-add-setup-project/rules/tsconfig.ts
+++ b/packages/angular/src/ng-add-setup-project/rules/tsconfig.ts
@@ -1,0 +1,32 @@
+import {Rule, SchematicContext, Tree} from '@angular-devkit/schematics';
+import {normalize} from '@angular-devkit/core';
+import {CoveoSchema} from '../../schema';
+import {EOL} from 'os';
+
+export function updateTsConfig(_options: CoveoSchema): Rule {
+  return (tree: Tree, _context: SchematicContext) => {
+    const tsconfigBuffer = tree.read(normalize('./tsconfig.json'));
+    if (tsconfigBuffer === null) {
+      return;
+    }
+    try {
+      const fileArray = tsconfigBuffer.toString().split(EOL);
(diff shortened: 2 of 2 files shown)


louis-bompart (commented):
The way we modify the tsconfig.json seems odd to me.
Can't we as a schematic author act upon the tsconfig in a less roundabout way?
I was wondering especially after reading this page from Angular documentation:


> Your library needs a custom Typescript configuration file with instructions on how to compile your schematics into your distributed library.

https://angular.io/guide/schematics-for-libraries#building-your-schematics

y-lakhdar:
> The way we modify the tsconfig.json seems odd to me.
> Can't we as a schematic author act upon the tsconfig in a less roundabout way?
> I was wondering especially after reading this page from Angular documentation:
> 
> > Your library needs a custom Typescript configuration file with instructions on how to compile your schematics into your distributed library.
> 
> https://angular.io/guide/schematics-for-libraries#building-your-schematics

mmh.. I think that custom tsconfig file is only used in the schematic compilation rather than being applied to the generated project.
Another option would be to have our own tsconfig file within the template. However, it means completely overriding the existing one.

olamothe:
Another approach might be to just `skipLibCheck=true` in ts config ?

If that works, I'd prefer this one. Otherwise these kind of problem will keep popping up from time to time.

y-lakhdar:
> > > The way we modify the tsconfig.json seems odd to me.
> > > Can't we as a schematic author act upon the tsconfig in a less roundabout way?
> > > I was wondering especially after reading this page from Angular documentation:
> > > > Your library needs a custom Typescript configuration file with instructions on how to compile your schematics into your distributed library.
> > > 
> > > 
> > > https://angular.io/guide/schematics-for-libraries#building-your-schematics
> > 
> > 
> > mmh.. I think that custom tsconfig file is only used in the schematic compilation rather than being applied to the generated project.
> > Another option would be to have our own tsconfig file within the template. However, it means completely overriding the existing one.
> 
> Do you know if we can detect if we're generating or adding the project?
> What I would do iff we're detecting that we're adding is not modify the tsconfig but indicate that our template either requires:
> 
> * skipLibCheck as @olamothe proposed
> * allowSyntheticDefaultImports, as currently proposed.
> 
> I think we should also try to see what would be the impact of both on the customer project, what does he win or lose.
> We should also check what are the good practices of the Angular schematics in general.
> 
> Tho, if the issue is pressing, we can also proceed with a fix-it twice logic: Fix it with the proposed changes here (it works fine really), and investigate afterwards to see what's the 'best' fix.
> 
> (Personnaly, I locked the headless version to 0.10.0 for my test)

Yes, we can detect if we are adding or updating a file. However in this case, I'm pretty sure we will always be updating the file. 
Angular schematic SDK offers plenty of util functions to dynamically insert code fragment into the project tree, but nothing specific to the tsconfig file.
We can go with a stronger approach and use a public method such as [insertAfterLastOccurrence](https://github.com/angular/angular-cli/blob/c1512e42742c17ace82e783e8e9c919ae925d269/packages/schematics/angular/utility/ast-utils.ts#L225) and prevent a hackish JSON parse.

@olamothe, `skipLibCheck=true` works

louis-bompart (approved):

PR #194, opened 2021-05-03
test(cli): check browser logs for errors

y-lakhdar:
## Proposed changes

Added additional E2E tests to ensure no errors and warnings are thrown in the developer console.
The tests currently check for `error` and `warning` message [ConsoleMessageType](https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-consolemessagetype), but we can add more types if needed.

-----
CDX-252


packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -137,6 +143,26 @@ describe('ui:create:angular', () => {
...
+        'warning',
+      ];
+
+      page.on('console', (msg) => interceptedMessages.push(msg));
+
+      await page.goto(searchPageEndpoint, {
+        waitUntil: 'networkidle0',
+      });
+
+      expect(
```
olamothe, line 159: When it fails, I'd log what is the actual error/warning message that we caught from the console, and log it in the test output.

Will be easier to troubleshoot. Or is it already the case with the way `jest` logs on test failure ?

olamothe (approved):

packages/cli-e2e/__tests__/angular.specs.ts
y-lakhdar: That's what I am trying to do actually.
So far I found 2 quick options (but neither one seems clean):
1. pipe browser process stdout and stderr into process.stdout and process.stderr ([info](https://www.puppeteersharp.com/api/PuppeteerSharp.LaunchOptions.html#PuppeteerSharp_LaunchOptions_DumpIO)).
2. Implement this [workaround](https://github.com/puppeteer/puppeteer/issues/3397#issuecomment-429325514) which use an undocumented method.

The third option is a little bit more complex but more robust
3. Create a CDP session to directly interact with the Chrome DevTools Protocole and create a listener on the [Runtime.consoleAPICalled event](Runtime.consoleAPICalled #)

y-lakhdar: nvm. `Runtime.consoleAPICalled` is awesome!

y-lakhdar:
I updated the E2E tests by creating a CDP session. That way, we have more control on what is logged in the dev console.

packages/cli-e2e/__tests__/angular.specs.ts
louis-bompart: You could probably reuse the Logger to log into files ;)
Like this, if we get a buttload of logs, we can still sieve through them.

packages/cli-e2e/__tests__/react.specs.ts
```diff
@@ -1,12 +1,19 @@
...
+  ConsoleMessageType,
+  CDPSession,
+} from 'puppeteer';
 import stripAnsi from 'strip-ansi';
 
 import {captureScreenshots, getNewBrowser, openNewPage} from '../utils/browser';
 import {getProjectPath, setupUIProject} from '../utils/cli';
 import {isSearchRequest} from '../utils/platform';
 import {ProcessManager} from '../utils/processManager';
+import {Runtime} from 'node:inspector';
```
louis-bompart, line 16:
```suggestion
import type {Runtime} from 'inspector';
```

packages/cli-e2e/__tests__/vue.specs.ts
```diff
@@ -13,6 +19,7 @@ import stripAnsi from 'strip-ansi';
 import {EOL} from 'os';
 import {ProcessManager} from '../utils/processManager';
 import {deactivateEnvironmentFile, restoreEnvironmentFile} from '../utils/file';
+import {Runtime} from 'node:inspector';
```
louis-bompart, line 22:
```suggestion
import type {Runtime} from 'inspector';
```

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -14,6 +20,7 @@ import {isSearchRequest} from '../utils/platform';
 import {EOL} from 'os';
 import stripAnsi from 'strip-ansi';
 import {ProcessManager} from '../utils/processManager';
+import {Runtime} from 'node:inspector';
```
louis-bompart, line 23:
```suggestion
import type {Runtime} from 'inspector';
```

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -113,20 +121,54 @@ describe('ui:create:angular', () => {
...
+              message.type as ConsoleMessageType
+            ) > -1
+          ) {
+            message.args.forEach((arg) => {
+              arg.description &&
+                interceptedConsoleMessages.push(arg.description);
+            });
+          }
+        }
+      );
```
louis-bompart, line 163: Refactor this bit in an utils and put it in `browser.ts`.
It'll be helpful.

louis-bompart (changes requested):
Really good finds with those API, good job! 🙌 
- I'd change the node: thingy to the true module and import type, it's simpler for ts (heck i don't know what those node:thingy are tbh)
- I'd refactor the CDP hooking in an utils.
- If you add logging, I'd do so in files, same as with the terminal.

packages/cli-e2e/__tests__/angular.specs.ts
y-lakhdar: we still need the `node:` prefix otherwise we get a eslint warning
image.

We can either use `node:` or disable the eslint rule

packages/cli-e2e/utils/consoleInterceptor.ts
```diff
@@ -0,0 +1,53 @@
+import {CDPSession, ConsoleMessageType, Page} from 'puppeteer';
+import {FileLogger} from './terminal/terminal';
+import type {Runtime} from 'node:inspector';
```
louis-bompart, line 3: You should not need interceptor but `devtools-protocol` types instead

louis-bompart (changes requested):
Some renaming, but most importantly target the actual interface being used by pupptr to avoid some 'close enough' casting

lerna.json
```diff
@@ -1,7 +1,5 @@
 {
-  "packages": [
-    "packages/*"
-  ],
+  "packages": ["packages/*"],
```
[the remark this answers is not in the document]
y-lakhdar: automatic lint I guess

louis-bompart (approved):
Really clean good job!

PR #196, opened 2021-05-03
build(cli): lock @coveo/headless on official build

olamothe:
## Proposed changes

Use 0.x version on official build; use alpha in dev/day to day mode.

## Breaking changes

--

## Testing

manual

https://coveord.atlassian.net/browse/CDX-281

### DIFF
--- a/packages/cra-template/package.json
+++ b/packages/cra-template/package.json
@@ -27,7 +27,7 @@
     "template.json"
   ],
   "devDependencies": {
-    "@coveo/headless": "latest",
+    "@coveo/headless": "alpha",
     "@material-ui/core": "^4.11.3",
     "@material-ui/lab": "^4.0.0-alpha.57",
     "@testing-library/jest-dom": "^5.11.9",
--- a/scripts/inject-headless-version.js
+++ b/scripts/inject-headless-version.js
@@ -1,11 +1,36 @@
-import {readFileSync} from 'fs';
+const {readFileSync, writeFileSync} = require('fs');
+
+const packages = ['angular', 'cra-template', 'vue-cli-plugin-typescript'];
+const headlessVersion = '0.x';
+
 async function main() {
-  const packages = ['angular', 'cra-template', 'vue-cli-plugin-typescript'];
-  //  const headlessVersion = '0.x';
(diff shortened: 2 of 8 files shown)


scripts/inject-headless-version.js
```diff
@@ -0,0 +1,36 @@
+const {readFileSync, writeFileSync} = require('fs');
+
+const packages = ['angular', 'cra-template', 'vue-cli-plugin-typescript'];
+const headlessVersion = '0.x';
```
louis-bompart, line 4: This will be changed I presume by `^1.0.0` when the v1 will be released?

louis-bompart:
It's a bit more complex for angular.
The dependency is installed in the `ng-add-project-setup` and is targeting by default `latest` 

I advise you to modify the addToPackageJson method as such:
```ts
export function addToPackageJson(
  packageName: string,
  version = 'latest'
): Rule {
  return (tree: Tree, _context: SchematicContext) => {
    const packageToAdd: NodeDependency = {
      type: NodeDependencyType.Default,
      name: packageName,
      version,
      overwrite: true,
    };

    addPackageJsonDependency(tree, packageToAdd);
  };
}```
After that, you'll probably have to 'fetch' the version somehow from the package.json of the schematic.

File: scripts/inject-headless-version.js (hunk 1 above)
Comment (olamothe) [reply scripts/inject-headless-version.js:4]: Yes, exactly.

File: packages/cra-template/package.json [hunk 2]
```diff
@@ -27,7 +27,7 @@
     "template.json"
   ],
   "devDependencies": {
-    "@coveo/headless": "latest",
+    "@coveo/headless": "alpha",
```
y-lakhdar, line 30: This will not affect the generated project.
You need to update `packages/cra-template/template.json`
[The template.json file](https://create-react-app.dev/docs/custom-templates/#the-templatejson-file).

Unless, you find a way to read from the package.json of the template...

y-lakhdar (changes requested):
Version locking will not work on React projects unless `packages/cra-template/template.json` is updated.

olamothe:
Alright, I was naive on this one ;D I'll revisit for each project.

PR #197, opened 2021-05-03
feat(cli): added a new org:create command

marcantoineveilleux:
## Proposed changes 

Add a new command to create new organizations

To do so, I needed to change the way the authorization work. You don't need to have an org to be able to authenticate anymore.

## Breaking changes

None

## Testing
Manual for now

-----
CDX-MAV :P

### DIFF
--- a/packages/cli/src/commands/org/create.ts
+++ b/packages/cli/src/commands/org/create.ts
@@ -1,0 +1,50 @@
+import {Command, flags} from '@oclif/command';
+import {cli} from 'cli-ux';
+import {Config} from '../../lib/config/config';
+import {
+  Preconditions,
+  HasAccessToken,
+} from '../../lib/decorators/preconditions/';
+import {AuthenticatedClient} from '../../lib/platform/authenticatedClient';
+
+export default class OrgCreate extends Command {
+  static description = 'Create a new trial Coveo organization';
+
+  private configuration!: Config;
+
+  static flags = {
+    setDefaultOrganization: flags.boolean({
+      char: 'D',
+      default: false,
+      description: 'Set the newly created organization as the default one',
+    }),
+  };
+
+  static args = [
+    {
+      name: 'name',
+      description: 'The name to assign to the new organization.',
+      required: true,
+    },
+  ];
+
+  @Preconditions(HasAccessToken())
+  async run() {
+    const {args, flags} = this.parse(OrgCreate);
+
+    this.configuration = new Config(this.config.configDir, this.error);
+    const authenticatedClient = new AuthenticatedClient();
+    const newCreatedOrg = await authenticatedClient.createOrganization(
+      args.name
+    );
+
+    if (flags.setDefaultOrganization) {
+      this.configuration.set('organization', newCreatedOrg.id);
+      cli.log('Organization created and set as default');
+    } else {
+      cli.log('Organization created');
+    }
+
+    cli.styledObject(newCreatedOrg);
+  }
+}
--- a/packages/cli/src/commands/org/list.ts
+++ b/packages/cli/src/commands/org/list.ts
@@ -7,7 +7,7 @@ import {
 } from '../../hooks/analytics/analytics';
 import {
   Preconditions,
-  IsAuthenticated,
+  HasAccessToken,
 } from '../../lib/decorators/preconditions/';
 
 export default class List extends Command {
@@ -17,7 +17,7 @@ export default class List extends Command {
     ...cli.table.flags(),
   };
 
-  @Preconditions(IsAuthenticated())
+  @Preconditions(HasAccessToken())
   async run() {
     const {flags} = this.parse(List);
     const orgs = await new AuthenticatedClient().getAllOrgsUserHasAccessTo();
--- a/packages/cli/src/lib/decorators/preconditions/authenticated.ts
+++ b/packages/cli/src/lib/decorators/preconditions/authenticated.ts
@@ -1,4 +1,5 @@
 import Command from '@oclif/command';
+import {Config} from '../../config/config';
 import {
   AuthenticationStatus,
   getAuthenticationStatus,
@@ -6,6 +7,17 @@ import {
 
 export function IsAuthenticated() {
   return async function (target: Command) {
+    const cfg = new Config(global.config.configDir);
+
+    const {organization} = await cfg.get();
+
+    if (!organization || organization.length === 0) {
+      target.warn(
+        'Not currently logged to a specific organization. Run coveo auth:login or config:set.'
+      );
+      return false;
+    }
+
     const status = await getAuthenticationStatus();
     if (status === AuthenticationStatus.LOGGED_OUT) {
       target.warn('Not currently logged in. Run coveo auth:login first.');
--- a/packages/cli/src/lib/decorators/preconditions/authenticated.ts
+++ b/packages/cli/src/lib/decorators/preconditions/authenticated.ts
@@ -11,7 +11,7 @@ export function IsAuthenticated() {
 
     const {organization} = await cfg.get();
 
-    if (!organization || organization.length === 0) {
+    if (organization?.length) {
       target.warn(
         'Not currently logged to a specific organization. Run coveo auth:login or config:set.'
       );
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -29,6 +29,16 @@
       "remoteRoot": "/home/notGroot/cli-copy",
       "sourceMaps": true,
       "autoAttachChildProcesses": false
+    },
+    {
+      "name": "Launch Program",
+      "program": "${workspaceFolder}/packages/cli/bin/run",
+      //"args": ["org:create", "-n", "testingOneTwo"],
+      //"args": ["auth:login", "-e", "dev"],
+      "args": ["ui:create:react", "test"],
+      "request": "launch",
+      "skipFiles": ["<node_internals>/**"],
+      "type": "pwa-node"
     }
   ]
 }
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -31,15 +31,19 @@ USAGE
 # Commands
 
 <!-- commands -->
-* [`coveo auth:login`](#coveo-authlogin)
-* [`coveo config:get`](#coveo-configget)
-* [`coveo config:set`](#coveo-configset)
-* [`coveo help [COMMAND]`](#coveo-help-command)
-* [`coveo org:list`](#coveo-orglist)
-* [`coveo ui:create:angular NAME`](#coveo-uicreateangular-name)
-* [`coveo ui:create:react NAME`](#coveo-uicreatereact-name)
-* [`coveo ui:create:vue NAME`](#coveo-uicreatevue-name)
-* [`coveo update [CHANNEL]`](#coveo-update-channel)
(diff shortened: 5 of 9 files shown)


packages/cli/src/lib/decorators/preconditions/authenticated.ts
```diff
@@ -1,11 +1,23 @@
...
   getAuthenticationStatus,
 } from '../../platform/authenticatedClient';
 
 export function IsAuthenticated() {
   return async function (target: Command) {
+    const cfg = new Config(global.config.configDir);
+
+    const {organization} = await cfg.get();
+
+    if (!organization || organization.length === 0) {
```
louis-bompart, line 14: Use optional chaining to simplify your code:
```suggestion
    if (organization?.length) {
```

louis-bompart (commented):
Smooth!
The code is very clean, You'll need to write:
 - Unit Test for the decorator & for the command
 - FT for the command. You'll need to be extra careful to ensure that the clean-up occurs properly otherwise we'll swarm dev with orgs (albeit with no index so its not that bad).

packages/cli/src/commands/org/list.ts
```diff
@@ -17,7 +17,7 @@ export default class List extends Command {
     ...cli.table.flags(),
   };
 
-  @Preconditions(IsAuthenticated())
+  @Preconditions(HasAccessToken())
```
olamothe, line 20: Why this change ? It's not because you have an access token that it is still valid/not expired/will actually be able to list orgs.

With this change, the command will now fail in a cryptic manner for the end user if the token is expired.

packages/cli/src/lib/decorators/preconditions/authenticated.ts
```diff
@@ -1,11 +1,23 @@
 import Command from '@oclif/command';
+import {Config} from '../../config/config';
 import {
   AuthenticationStatus,
   getAuthenticationStatus,
 } from '../../platform/authenticatedClient';
 
 export function IsAuthenticated() {
   return async function (target: Command) {
+    const cfg = new Config(global.config.configDir);
```
olamothe, line 10: Why this change ? 

It changes the meaning of the function: It's not about being "authenticated" in the platform anymore (as in: does the user have a valid access token as far as the platform is concerned), it's about verifying some value in a local configuration file.

If you really need to verify that "the user needs to have value X in his configuration file", then I would suggest creating a precondition function that does that, and to add the precondition it to all the commands that need it.

packages/cli/src/commands/org/create.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  static args = [
+    {
+      name: 'name',
+      description: 'The name to assign to the new organization.',
+      required: true,
+    },
+  ];
+
+  @Preconditions(HasAccessToken())
```
olamothe, line 31: Similar comment as above: It's not because you have an access token in a local configuration file that it will work.

You still need to verify that the token is valid.

If what you are trying to do is to `verify the validity of an access token without requiring the existence/knowledge of an organization ID`, then we need to change this in either platform-client ( https://github.com/coveo/platform-client ), or to add a function in authenticated client to call `platform.cloud.coveo.com/oauth/check_token`

olamothe (changes requested):
Unfortunately, the verification on access_token existence in configuration file will not work in a real life scenario, and will throw/crash commands whenever a end user just picks up the CLI after a couple day of inactivity.

packages/cli/src/commands/org/create.ts
```diff
@@ -0,0 +1,50 @@
...
+import {AuthenticatedClient} from '../../lib/platform/authenticatedClient';
+
+export default class OrgCreate extends Command {
+  static description = 'Create a new trial Coveo organization';
+
+  private configuration!: Config;
+
+  static flags = {
+    setDefaultOrganization: flags.boolean({
+      char: 'D',
```
olamothe, line 17: Lowercase `d`, similar to all other flags we use for other commands.

packages/cli/src/commands/org/create.ts
```diff
@@ -0,0 +1,50 @@
...
+import {cli} from 'cli-ux';
+import {Config} from '../../lib/config/config';
+import {
+  Preconditions,
+  HasAccessToken,
+} from '../../lib/decorators/preconditions/';
+import {AuthenticatedClient} from '../../lib/platform/authenticatedClient';
+
+export default class OrgCreate extends Command {
+  static description = 'Create a new trial Coveo organization';
```
olamothe, line 11: Thinking long term, this could be for "creating any coveo organization", and it being a trial/dev/sandbox/prod could be controlled by a flag.

Otherwise, we will essentially always be stuck in that mode for this particular command (since it uses the "namespace" `org:create`) and will need to create a new major version to break this behaviour if we need to.

The default value for the flag could be `trial`, however. The net result will be something very similar to what you have now, but with more room to grow in the future.

louis-bompart:
> Unfortunately, the verification on access_token existence in configuration file will not work in a real life scenario, and will throw/crash commands whenever a end user just picks up the CLI after a couple day of inactivity.

Good point @olamothe. I reckon we have an endpoint to ensure the validity of an `access_token`.

Or even if we don't a health check call to the platform ain't a bad idea either. (i.e. just do an authenticated call, if it faceplant, your validator return false, if it 2xx, validator return true)
I would probably call `GET /rest/privileges/me` if I were you @marcantoineveilleux

packages/cli/src/commands/org/create.ts
louis-bompart: I think that yes, we should consider it a default value.
Also (tho this is probably a discussion that needs to be done 'live'), we should not be afraid to do breaking changes.
What's important is for the breaking change to be well thought, motivated and documented with great precision.

Also IMO, even if we'd have no flag today if we add flags and they are optional, if the behaviour with no flags stays the same, then there no breaking change.

The breaking change definition for the CLI IMO should only be based on the 'command typed in' to 'end-result relation exclusively.

packages/cli/src/commands/org/list.ts
louis-bompart: _Technically_ we're not checking if the user is allowed to list org today.
However, it is true that were dropping on the token validity. Which is where your comment https://github.com/coveo/cli/pull/197#discussion_r625354148 should play its role.

packages/cli/src/commands/org/create.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    this.configuration = new Config(this.config.configDir, this.error);
+    const authenticatedClient = new AuthenticatedClient();
+    const newCreatedOrg = await authenticatedClient.createOrganization(
+      args.name
+    );
+
+    if (flags.setDefaultOrganization) {
+      this.configuration.set('organization', newCreatedOrg.id);
+      cli.log('Organization created and set as default');
```
y-lakhdar, line 43: Really not that important, but should we keep using `this.log` to stay consistent with other commands?

packages/cli/src/commands/org/create.ts
```diff
@@ -0,0 +1,50 @@
...
+    } else {
+      cli.log('Organization created');
+    }
+
+    cli.styledObject(newCreatedOrg);
```
y-lakhdar, line 48: And maybe log a url link to the newly created org?

packages/cli/src/commands/org/create.ts
marcantoineveilleux: so lower case `O` too?

packages/cli/src/commands/org/create.ts
marcantoineveilleux: I also thought of that and I have heard that some folks in the platform are looking to simplify the LicenseTemplate/OrgTemplate mechanism, which is why i decided to not expose it. Like louis said, in the future, we could easily add an optional flag to expose that behavior.

packages/cli/src/commands/org/create.ts
marcantoineveilleux: Loved to, but on a `Command`, there is no `cli` property. I am using exactly what we have in the `list` command (`cli.table`) via the import.

packages/cli/src/commands/org/create.ts
marcantoineveilleux: ish, i like the functionality, but hate the idea that the CLI craft links. Maybe it could be something that could be returned in the new created org call from the platform. I'll ask.

packages/cli/src/commands/org/create.ts
marcantoineveilleux: Also, this is our first flag with multiple words, did we look at the best practices?

packages/cli/src/commands/org/list.ts
marcantoineveilleux: Je crois que je vais revert tout ce qui a rapport avec à `IsAuthenticated` et `HasAccessToken `. Pour le changement en soi, j'en ai pas besoin. Je ne change pas le comportement en soi de ça; même si maintenant j'affiche qu'on s'est logged correctement lors du login. Autrefois, sans org, ça affichait un message de fail, mais après tu pouvais faire des commandes ne nécéssitant pas de specific org.

Je suggèrerais qu'on fasse la validation des orgs pour les commandes qui en ont besoin dans un autre changement. Qu'en dites-vous?

packages/cli/src/commands/org/create.ts
olamothe: Fair enough !

packages/cli/src/commands/org/create.ts
olamothe: You should be able to use `this.log` ? 🤔 .

packages/cli/src/commands/org/create.ts
```diff
@@ -0,0 +1,50 @@
...
+    },
+  ];
+
+  @Preconditions(HasAccessToken())
+  async run() {
+    const {args, flags} = this.parse(OrgCreate);
+
+    this.configuration = new Config(this.config.configDir, this.error);
+    const authenticatedClient = new AuthenticatedClient();
+    const newCreatedOrg = await authenticatedClient.createOrganization(
```
olamothe, line 37: Is this a long running operation ? Long running as in it takes multiple seconds for the end user ?

If so, you could look into using cli.action.start(), cli.action.pause() to provide nice visual feedback.

If it's "quick", then disregard this comment.

packages/cli/src/commands/org/create.ts
louis-bompart: Could you explain why you 'hate the idea that the CLI craft links'?

packages/cli/src/commands/org/list.ts
louis-bompart: I agree with the idea that the change between IsAuthenticated and HasToken should be dealt with in another PR.
However I would consider this change to be a prerequisite for this PR to go forward.

packages/cli/src/commands/org/create.ts
marcantoineveilleux: Herp derp, je viens de catcher.

packages/cli/src/commands/org/create.ts
marcantoineveilleux: In my experience, it's super fast.

olamothe:
@marcantoineveilleux : Please re-open when ready.

PR #198, opened 2021-05-04
test(e2e): refactor tests to uses Terminal + basic logger

louis-bompart:
## Proposed changes

This completes the refactor of the tests to uses the Terminal testing toolbelt from #192 and also address add a basic logger for the end-to-end tests.

The logger is just a collection of write stream and some basic piping to allow the text to flow.
The trade-off is that in case there multiple processes running at the same time and that one causes the other to crash, it's still a bit complicated to get the logs out. This will be refined in a further pull request to fully address our logging need (it'd still be simple, mostly adding timestamps and maybe a 'big garbage' log (i.e. everyone pipin' here prefixed by terminal alias or smthg.

image
Here's the dir hierarchy of the artifacts after this change, this folder will be what'll get from the artifact of a build (i.e. screenshots+logs)

## Testing

- CI + tested on my machine.

-----
CDX-284


packages/cli-e2e/utils/terminal/terminal.ts
```diff
@@ -26,21 +30,16 @@ export class Terminal {
...
-        console.log(`[STDOUT]${data.toString()}`)
-      );
-      this.childProcess.stderr.on('data', (data) =>
-        console.log(`[STDERR]${data.toString()}`)
-      );
-      this.childProcess.stdin.on('data', (data) =>
-        console.log(`[STDIN]${data.toString()}`)
-      );
-    }
+
```
louis-bompart, line 37: I chose to always log, it's dumb to re-run tests because you're missing logs, and it ain't costing much

packages/cli-e2e/utils/terminal/do.ts
```diff
@@ -153,6 +161,11 @@ export class Do implements Promise<void> {
       onParams = [condition, () => callback(this.action.process, resolve)];
     }
 
+    if (isConditionPromise(condition)) {
+      condition.then(() => callback(this.action.process, resolve));
+      return;
```
louis-bompart, line 166: A new type of 'Condition': Promises.

They replace the original 'function' promises but are much simpler to work with, especially combined with other terminal toolchains (see what's done in the spec refactor)

packages/cli-e2e/utils/terminal/do.ts
```diff
@@ -62,8 +67,8 @@ export class Do implements Promise<void> {
    * @returns {Promise} an awaitable promise that will resolve when the `untilCondition` is satisfied.
    */
   public until(untilCondition: Condition): Promise<void> {
-    let eventListener: EventListenerObject;
-    new Promise<void>((resolve) => {
+    let eventListener: EventListenerObject | undefined;
```
louis-bompart, line 70: Can now be undefined due to PromiseCondition

packages/cli-e2e/utils/cli.ts
```diff
@@ -3,38 +3,23 @@ import {resolve, join} from 'path';
...
-import stripAnsi from 'strip-ansi';
-
 import {ProcessManager} from './processManager';
 import {readJsonSync} from 'fs-extra';
 import type {AxiosResponse} from 'axios';
 import axios from 'axios';
-
-export function isYesNoPrompt(data: string) {
-  return data.trimEnd().toLowerCase().endsWith('(y/n):');
-}
```
louis-bompart, line 15: Was dup of `isGenericYesNoPrompt` so it got 🪓 .

packages/cli-e2e/utils/cli.ts
```diff
@@ -3,38 +3,23 @@ import {resolve, join} from 'path';
...
-
-export function isGenericYesNoPrompt(data: string) {
-  let stripedData = data;
-  try {
-    stripedData = stripAnsi(data.toString());
-  } catch (error) {
-    console.log('Unable to strip ansi from string', error);
-  }
-  return /\(y\/n\)[\s:]*$/i.test(stripedData);
-}
```
louis-bompart, line 25: Logic now handled by the Terminal Test Toolbelt, so we just need the regexp

packages/cli-e2e/utils/cli.ts
```diff
@@ -3,38 +3,23 @@ import {resolve, join} from 'path';
...
-  answer: string,
-  proc: ChildProcessWithoutNullStreams
-) {
-  return new Promise<void>((resolve) => {
-    if (!proc.stdin.write(answer)) {
-      proc.stdin.once('drain', () => resolve());
-    } else {
-      process.nextTick(() => resolve);
-    }
-  });
```
louis-bompart, line 37: got reworked to fit nicer with Terminal Test Toolbelt

louis-bompart:
Built on top of #192, it should be merged first to avoid having a big PR

packages/cli-e2e/__tests__/auth.specs.ts
```diff
@@ -23,20 +24,19 @@ describe('auth', () => {
...
-          await answerPrompt('n', cliProcess);
-        }
-      });
+      const cliTerminal = new Terminal(
+        CLI_EXEC_PATH,
+        ['auth:login', '-e=dev'],
+        undefined,
+        processManager,
+        'auth-login'
+      );
```
y-lakhdar, line 33: To avoid telescoping constructors, can we move some of these optional parameters into a single option?

olamothe (approved):

y-lakhdar (approved):

packages/cli-e2e/__tests__/auth.specs.ts
louis-bompart: I'm not sure to understand what you mean?
One thing I try to do tho is to avoid matryoshka arguments and keep the tests 'flat'. (i.e. avoid 42th level of indentation)

However, we certainly may do better processing of the args, I guess we could do that simply with the command chaining. I will do something akin to that in another PR ;)

PR #199, opened 2021-05-04
test(vue): ensure eslint is OK in the generated vue project

louis-bompart:
## Proposed changes

Based on #198.
Simply ensure that `npm start` in the Vue Project does not write a matcher specific to ESLint issues (from what I experienced).
It required some refactoring to 'split' the specs into two parts: one with a server started (and once to not waste too much time), another with no server running to be able to run it and check what's what.

## Testing
Checked in two parts:
 1. By adding bogus code into a Vue file and ensuring there was an error and the test tripped.
 2. By removing the aforementioned bogus code and ensuring the test passed.
 
This was done on my machine, I expect `#2` to pass also on the CI, obviously. 

-----
CDX-237


louis-bompart:
Third wagon of the PR train 🚂, needs #198 first.

olamothe (approved):

y-lakhdar (approved):

PR #205, opened 2021-05-06
test(cli): add timestamps in log files

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-300

## Proposed changes

Prepend a timestamp on every line that is added to the artifacts.
This will help knowing how long each spawn process takes
image

-----
CDX-300

### DIFF
--- a/packages/cli-e2e/utils/terminal/terminal.ts
+++ b/packages/cli-e2e/utils/terminal/terminal.ts
@@ -8,6 +8,7 @@ import type {When} from './when';
 import {ProcessManager} from '../processManager';
 import {Orchestrator} from './orchestrator';
 import {FileLogger} from '../filelogger';
+import {WriteStream} from 'node:fs';
 
 /**
  * An helper class to manipulate processes that interact with a TTY.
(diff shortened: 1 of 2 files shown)


y-lakhdar:
Just doing some tests with the github action before marking the PR ready for review

packages/cli-e2e/utils/terminal/terminal.ts
```diff
@@ -35,11 +36,22 @@ export class Terminal {
...
+    this.logIntoFile(fileLogger);
     this.orchestrator = new Orchestrator(this.childProcess);
   }
 
+  private logIntoFile(fileLogger: FileLogger) {
+    const appendTimestamp = (stream: WriteStream) => () =>
+      stream.write(fileLogger.getTimestamp());
+
+    this.childProcess.stdout.on('data', appendTimestamp(fileLogger.stdout));
+    this.childProcess.stderr.on('data', appendTimestamp(fileLogger.stderr));
```
louis-bompart, line 49: I think it'd be better to do something akin to:
chilldProcess.stdout pipe into a 'timestamper' and then the timestamper pipe into the file.

y-lakhdar:
> @y-lakhdar decline in favor of #207 ?

Yes, let's merge #207

PR #206, opened 2021-05-06
test(cli): increase E2E timetout for angular

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-299

## Proposed changes

The angular E2E tests sometime fail and I suspect a timeout issue. In this PR, we simply give a couple more minutes to the angular setup phase.

Once #205 is merged, we will be able to confirm if the errors are caused by an insufficient timeout by inspecting the log timestamps.

-----
CDX-299

### DIFF
--- a/packages/cli-e2e/__tests__/angular.specs.ts
+++ b/packages/cli-e2e/__tests__/angular.specs.ts
@@ -77,7 +77,7 @@ describe('ui:create:angular', () => {
     browser = await getNewBrowser();
     await buildApplication(buildProcessManager);
     await buildProcessManager.killAllProcesses();
-  }, 15 * 60e3);
+  }, 20 * 60e3);
 
   beforeEach(async () => {
     jest.resetModules();
@@ -107,7 +107,7 @@ describe('ui:create:angular', () => {
       serverProcessManager = new ProcessManager();
       processManagers.push(serverProcessManager);
       await startApplication(serverProcessManager);
-    }, 5 * 60e3);
+    }, 15 * 60e3);
 
     beforeEach(async () => {
       consoleInterceptor = new BrowserConsoleInterceptor(page, projectName);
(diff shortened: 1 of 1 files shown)


packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -130,7 +130,7 @@ describe('ui:create:angular', () => {
 
     it('should not contain console errors nor warnings', async () => {
       await page.goto(searchPageEndpoint, {
-        waitUntil: 'networkidle0',
+        waitUntil: 'networkidle2',
```
louis-bompart, line 133: What does motivate you to change from idle0 to idle2?

louis-bompart:
Looks good, overall.
However, like you mentioned in your description, I would wait for #205 and see it in action first.

If we want to be confident in our test, we need to be quite sure of the source of the problem when we do a fix.

I personally think that it's possible that it's just a timeout issue, but I'd prefer to be more certain.

packages/cli-e2e/__tests__/angular.specs.ts
y-lakhdar: I came across this [issue](https://github.com/puppeteer/puppeteer/issues/1908)

y-lakhdar:
> Looks good, overall.
> However, like you mentioned in your description, I would wait for #205 and see it in action first.
> 
> If we want to be confident in our test, we need to be quite sure of the source of the problem when we do a fix.
> 
> I personally think that it's possible that it's just a timeout issue, but I'd prefer to be more certain.

Sorry, I updated the PR a couple minutes before your comment (hoping no one would notice XD).
I ran the E2E tests locally with the timestamps and was able to reproduce the same timeout error. The thing is that the setup phase had plenty of time to finish... so the setup timeout has nothing to do with the test failing.

I also noticed that the test that was intermittently failing was the one testing the console errors... and that's when I found this [issue](https://github.com/puppeteer/puppeteer/issues/1908)

olamothe (approved):

PR #217, opened 2021-05-13
fix(vue): .eslintrc blocks user from commiting

y-lakhdar:
## Proposed changes

Removed `.eslintrc` file from `search-token-server` package as it does not have any additional effect. Furthermore, the `extends` property in that eslint file was pointing to an invalid path once added to a ui project.

The `/server` folder will be linted based on the project configuration (root eslint file)

## Testing

E2E tests to ensure the project can be commited without the `lint-staged` hook to fail.

-----
CDX-320


packages/search-token-server/.eslintrc
```diff
@@ -1,6 +0,0 @@
-{
-  "extends": "../../.eslintrc",
-  "rules": {
-    "node/no-unpublished-import": ["off"]
```
y-lakhdar, line 4: has no effect since all dependancies should be listed in the package.json

packages/search-token-server/.eslintrc
```diff
@@ -1,6 +0,0 @@
-{
-  "extends": "../../.eslintrc",
```
y-lakhdar, line 2: this line breaks point to an invalid path once added to a ui project (react, angular, vue)

packages/search-token-server/app.ts
```diff
@@ -7,7 +7,7 @@ const app = express();
 
 app.use(express.json());
 
-app.get<{}, any, {token: string}>(
+app.get<Record<string, never>, any, {token: string}>(
```
olamothe, line 10: Maybe: `Record<string, unknown>` ?

olamothe (approved):

packages/search-token-server/app.ts
y-lakhdar: `any` would work, but not `unknown` because of this condition: `Type 'unknown' is not assignable to type 'string'.`

packages/vue-cli-plugin-typescript/generator/index.js
```diff
@@ -12,6 +12,8 @@ module.exports = (api, options, rootOptions) => {
       '@coveo/search-token-server': '*',
       buefy: '^0.9.4',
       concurrently: '^5.3.0',
+      dotenv: '^9.0.2',
+      'fs-extra': '^10.0.0',
```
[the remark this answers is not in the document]
y-lakhdar: It's because the generated project depends on these 2 packages.
Currently we were lucky because these packages are available though other packages, hence why it never failed.

packages/search-token-server/app.ts
y-lakhdar: now that server/ has its own estlint, I will revert to the original type

PR #224, opened 2021-05-17
fix(vue): .eslintrc blocks user from commiting 

y-lakhdar:
An  exact copy of the PR #217
-----
CDX-320


packages/cli-e2e/utils/misc.ts
```diff
@@ -0,0 +1,56 @@
+import {ProcessManager} from './processManager';
```
louis-bompart, line 1: Rename the file git.ts

`misc` it's a large net :P

I know us developer, we're a lazy bunch and we'll end up dropping everything in there 😂  (or maybe it's just me 🤷 )

louis-bompart (approved):
lgtm!

olamothe (approved):

PR #229, opened 2021-05-19
feat: assign a dynamic port to both front and back ends on all ui projects

y-lakhdar:
## Proposed changes

Prevent the application to crash if the specified ports are already in use. Instead, the next available ports will be held for both the front-end application and search token server.
The idea is to run a script (`script/port-allocator.js`) upstream to allocate available ports to the front and back. The script checks if a port has already been specified by the user.
In all cases, a validation ensures that the port is actually available.

For React and Vue, it’s pretty straightforward: Run the port-allocator logic and hold the available port in a .env file. However, it is getting a little bit trickier with Angular.

Angular uses `environment.ts` files to store environment variables. It becomes very difficult (and against the Angular best practices) to read values from an `.env` file. In this case, the port-allocator service will directly read and update the `environment.ts` file with the appropriate port to the search token server.

There is a very tiny chance of a race condition if another process starts using the same port number as you in between the time you get the port number and you actually start using it... but seriously...

## Testing

- [ ] E2E Tests: Add a test in which the application tries to run on a busy port
- [x] Manual Tests:
    Tested with the 3 frameworks. Whenever the web application (or search token server) tries to run on a busy port, they get allocated available ports.

-----
CDX-99


y-lakhdar:
Sorry. I wanted to open the PR in draft mode
Actual PR is in #230

PR #233, opened 2021-05-21
feat(search): add a dump index command

olamothe:
https://coveord.atlassian.net/browse/CDX-341

## Proposed changes

Add a command to fetch the content of a whole source, and export it to a CSV format.

## Testing

- [X] Unit Tests:
- [X] Manual Tests:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,6 +21,7 @@
     "exponential-backoff": "^3.1.0",
     "fs-extra": "^9.1.0",
     "isomorphic-fetch": "^3.0.0",
+    "json2csv": "^5.0.6",
     "node-pty": "^0.10.0",
     "patch-package": "^6.2.2",
     "semver": "^7.3.4",
@@ -35,6 +36,7 @@
     "@oclif/test": "^1",
     "@types/fs-extra": "^9.0.6",
     "@types/jest": "^26.0.22",
+    "@types/json2csv": "^5.0.2",
     "@types/node": "^10",
     "chai": "^4.2.0",
     "eslint": "^7.23.0",
@@ -83,6 +85,9 @@
       },
       "ui:create": {
         "description": "create a user interface powered by different front end frameworks and Coveo Headless"
+      },
+      "search": {
+        "description": "perform queries against a coveo index"
       }
     },
     "hooks": {
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -86,7 +86,7 @@
       "ui:create": {
         "description": "create a user interface powered by different front end frameworks and Coveo Headless"
       },
-      "search": {
+      "org:search": {
         "description": "perform queries against a coveo index"
       }
     },
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -34,6 +34,7 @@
     "@coveo/vue-cli-plugin-typescript": "^1.2.1",
     "@oclif/dev-cli": "^1.26.0",
     "@oclif/test": "^1",
+    "@types/cli-progress": "^3.9.1",
     "@types/fs-extra": "^9.0.6",
     "@types/jest": "^26.0.22",
     "@types/json2csv": "^5.0.2",
--- a/packages/cli/src/commands/search/dump.ts
+++ b/packages/cli/src/commands/search/dump.ts
@@ -35,8 +35,10 @@ interface SearchResponse {
 interface FetchParameters {
   client: PlatformClient;
   organizationId: string;
-  source: string;
+  sources: string[];
   additionalFilter: string;
+  fieldsToExclude: string[] | undefined;
+  pipeline: string | undefined;
 }
 
 export default class Dump extends Command {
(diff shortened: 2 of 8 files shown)


packages/cli/src/commands/search/dump.ts
```diff
@@ -0,0 +1,169 @@
...
+  static description =
+    'Dump the whole content of a particular source in a CSV format.';
+
+  static flags = {
+    source: flags.string({
+      char: 's',
+      description:
+        'The identifier of the source for which to extract all documents.',
+      helpValue: 'mySourceName',
+      required: true,
```
y-lakhdar, line 52: I think that in some cases you will want to extract an aggregate of sources.
Will it be considered too dangerous to make the source flag optional?

Also, I think that a pipeline flag would be really useful.
**A past use case: "Extract all the documents from my Salesforce Agent pipeline."**

y-lakhdar: Also, if the "default" pipeline has filters, the dump will may not contain all the documents.

packages/cli/src/commands/search/dump.ts
```diff
@@ -0,0 +1,169 @@
...
+      additionalFilter: flags.additionalFilter,
+    });
+
+    if (allResults.length === 0) {
+      this.log(
+        'Found no results. Are you sure the source name and filter are valid ?'
+      );
+    } else {
+      const data = allResults.map((r) => r.raw);
+      const parser = new Parser({fields: Object.keys(allResults[0].raw)});
```
y-lakhdar, line 94: ouf. The 2 above 2 lines might take a lot of memory if the index has a couple 100K documents.
Maybe a writeStream would be more appropriate here... or just dumping into the CSV after each call.

packages/cli/package.json
```diff
@@ -83,6 +85,9 @@
       },
       "ui:create": {
         "description": "create a user interface powered by different front end frameworks and Coveo Headless"
+      },
+      "search": {
+        "description": "perform queries against a coveo index"
```
louis-bompart, line 90: How do you think in the future the topic subdivision with org etc?

One could argue that an index is something relative to an org, and therefore that it should be `org:search:dump`

packages/cli/src/commands/search/dump.ts
olamothe: Ive ran it on a source with 50k documents and it was fine. 
Took ~1min to query/download, but the file creation was pretty instant.

It took a while to open the CSV with excel though :D 

I could split on chunk, however.

Say, 10k documents maximum per file.

packages/cli/src/commands/search/dump.ts
olamothe: I can make source repeatable.

As for the pipeline, this is a great idea, Ill add it.

packages/cli/package.json
olamothe: Fair enough !

packages/cli/src/commands/search/dump.ts
y-lakhdar: correct. that's specially the case in commerce implementations with large spec fields...
or the infamous "all metadata fields".

packages/cli/src/commands/search/dump.ts
```diff
@@ -0,0 +1,169 @@
...
+import {writeFile} from 'fs-extra';
+import {Parser} from 'json2csv';
+import {cli} from 'cli-ux';
+import {
+  buildAnalyticsFailureHook,
+  buildAnalyticsSuccessHook,
+} from '../../hooks/analytics/analytics';
+
+interface SearchResult {
+  raw: {rowid: string};
```
louis-bompart, line 18: I think the `rowid` come from the backend, right?
If not, I'd say `rowId` (i.e. camelCase)

packages/cli/src/commands/search/dump.ts
louis-bompart: Chunk division is not a bad idea.
Also from a perf perspective, a writeStream is also a good idea, keeping an handle on the file while writting on it will be more efficient.
However you have to be absolutely sure to properly close it, even in case of failure. Otherwise it's a fd/socket that may stay open (although node does its best effort to close this kind of stuff in case of dirty closure, it's not perfect)

packages/cli/src/commands/search/dump.ts
```diff
@@ -0,0 +1,169 @@
...
+    if (response.results.length === 1000) {
+      const lastRowID = response.results[999].raw.rowid;
+      const nextPage = await this.fetchResults(
+        params,
+        progress,
+        lastRetrievedIndexToken,
+        lastRowID
+      );
+      return response.results.concat(nextPage);
+    } else {
```
louis-bompart, line 158: else is useless here ;)

packages/cli/src/commands/search/dump.ts
```diff
@@ -0,0 +1,169 @@
...
+import {
+  buildAnalyticsFailureHook,
+  buildAnalyticsSuccessHook,
+} from '../../hooks/analytics/analytics';
+
+interface SearchResult {
+  raw: {rowid: string};
+}
+
+interface ProgressBar {
```
louis-bompart, line 21: Why do we need a custom interface?

louis-bompart (commented):
Looks good, some🤨 and questions :D

packages/cli/src/commands/search/dump.ts
olamothe: Went with chunks, 10k results by default, possibility to exclude fields.

packages/cli/src/commands/search/dump.ts
olamothe: It's untyped from oclif (everything set as any). 

It's ultimately coming from another lib (cli-progress) that oclif uses. 

The alternative is to install @types/cli-progress, and tell eslint to remove warnings about extraneous import (we don't rely on cli-progress directly).

y-lakhdar (approved):
I think writing into the CSVs incrementally would be better in terms of memory consumption... instead of waiting for the all the results.
But really nice feature 👌

PR #235, opened 2021-05-28
feat(cli): create Zip from project

y-lakhdar:
## Proposed changes

Reads the project, compress the resources into a single zip and use it to create a snapshot.
The command uses a third party library to compress the resources. I initially tried using the native `zlib` API but it seems to only support single file compression. We can probably achieve the same result with some stream manipulation but, I am keeping it simple for now.

## Testing

- [x] Unit Tests:
    
- [x] Manual Tests: Created a project using this [structure](https://github.com/y-lakhdar/snapshot-automated-deployment/tree/master/Snapshot). 

-----
CDX-353


packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -0,0 +1,69 @@
...
+
+// TODO: CDX-357: platform-client should support zip file as stream.
+// In the meantime, we pass a custom object that contains all the require parameters expected by the createFromFile method.
+export interface CustomFile extends ReadStream {
+  type?: string;
+}
+
+export class Snapshot {
+  private client: AuthenticatedClient;
+  private lastSnapshot: ResourceSnapshotsModel | null = null;
```
louis-bompart, line 16: Why null and not undefined ? 🤔

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -0,0 +1,69 @@
...
+  private displayExpandedPreview() {
+    // TODO: CDX-347 Display Expanded preview
+  }
+
+  async deleteSnapshot() {
+    // TODO: CDX-359: Delete snapshot once previewed
+  }
+
+  private async getSnapshotClient() {
+    return (await this.client.getClient()).resourceSnapshot;
```
louis-bompart, line 67: I would probably do it as a prop
so `private _snapshotClient` and a getter. Like this you can cache it and avoid to call the lib again :)

louis-bompart (commented):
did an early sneak peak between two builds ;)

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: I split this PR into 2 PRs.
The snapshot class is now defined in the PR #236.

I decided to remove the `getSnapshotClient` method and put the client as an argument in the constructor.
**Reason behind that design:**
* If the user decides to deploy a snapshot to an org other than the one he is connected to, the `getSnapshotClient` will return an instance of the client tied to the wrong org.
* The Snapshot class does not need to know about the origin org. Only the target org matters.

packages/cli/src/commands/org/config/preview.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    const factory = new SnapshotFactory();
+    const snapshot = await factory.createFromZip(
+      pathToZip,
+      'cli-preview-snapshot'
+    );
+
+    cli.action.start('Creating snapshot');
+
+    const targetOrg = await this.getTargetOrg();
```
olamothe, line 53: Should there be a confirmation message here ?

`You are about to create and preview a snapshot for org "foo" in region "bar", do you wish to continue ?`

Or would that be too annoying ? Just to make sure that you are doing the operation in the correct org, like if you logged into an incorrect one...

packages/cli/src/lib/project/project.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  private get pathToTemporaryZip() {
+    return join(this.pathToProject, 'snapshot.zip');
+  }
+
+  private get pathToResources() {
+    return join(this.pathToProject, 'resources');
+  }
+
+  deleteTemporaryZipFile() {
```
olamothe, line 19: Similar to another comment in your other PR, I'd prefer to have the `public` keyword explicitly set, and to have all public methods grouped at the beginning of the class.

packages/cli/src/lib/project/project.ts
```diff
@@ -0,0 +1,57 @@
...
+    unlinkSync(this.pathToTemporaryZip);
+  }
+
+  ensureProjectCompliance() {
+    /*
+     * TODO: CDX-354: add checks to ensure the project is indeed a valid project
+     * e.g. * Check if path to resources is a folder
+     *      * Check if the root has a valid config file
```
olamothe, line 27: Possibly means we will need a `repair` command ? That's not 100% mandatory by the way... 

As end user, you could always get something valid from your git history anyway.

olamothe (approved):

packages/cli/src/commands/org/config/preview.ts
louis-bompart: I think we should not, too much of a hassle.

I would reserve such warning for
 - Destructive operations
 - Operations having impact on the queries of the org (e.g. if an operation change a query pipeline in the org). So when you apply snapshot

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+const mockedCreateSnapshotFromFile = jest.fn();
+
+const doMockReadStream = () => {
+  mockedCreateReadStream.mockImplementation(() => {
+    const buffer = Buffer.from('this is a tést');
+    const readable = new Readable();
+    readable._read = () => {};
+    readable.push(buffer);
+    readable.push(null);
+    return readable as unknown as ReadStream;
```
louis-bompart, line 22: You probably could do something like `createReadStream(Buffer.from('this is a tést'))`

You can use the callback of mockImplementation to alter it properly:
https://jestjs.io/docs/mock-function-api#mockfnmockimplementationfn

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+const doMockAuthenticatedClient = () => {
+  mockedAuthenticatedClient.mockImplementation(
+    () =>
+      ({
+        getClient: () =>
+          Promise.resolve({
+            resourceSnapshot: {createFromFile: mockedCreateSnapshotFromFile},
+          }),
+      } as unknown as AuthenticatedClient)
```
louis-bompart, line 34: I don't understand why you have to do such cast 🤔

louis-bompart (changes requested):
Some changes in the mocks :)

packages/cli/src/lib/snapshot/snapshot.spec.ts
y-lakhdar: So I don't have to implement the missing properties
image

packages/cli/src/commands/org/config/preview.ts
y-lakhdar: I agree with @louis-bompart Snapshot should not apply any modification to the destination org.
However, there will be a prompt if the snapshot requires a synchronization plan.

packages/cli/src/lib/snapshot/snapshot.spec.ts
y-lakhdar: Not sure I understand how I can mock the createReadStream by using it. Do you have an example?

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import {createReadStream, ReadStream} from 'fs';
+import {join} from 'path';
+import {Readable} from 'stream';
+import {mocked} from 'ts-jest/utils';
+import {AuthenticatedClient} from '../platform/authenticatedClient';
+import {SnapshotFactory} from './snapshotFactory';
+
+const mockedCreateReadStream = mocked(createReadStream);
+const mockedAuthenticatedClient = mocked(AuthenticatedClient);
```
louis-bompart, line 12:
```suggestion
const mockedAuthenticatedClient = mocked(AuthenticatedClient, true);
```

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+const doMockReadStream = () => {
+  mockedCreateReadStream.mockImplementation(() => {
+    const buffer = Buffer.from('this is a tést');
+    const readable = new Readable();
+    readable._read = () => {};
+    readable.push(buffer);
+    readable.push(null);
+    return readable as unknown as ReadStream;
+  });
+};
```
louis-bompart, line 24:
```suggestion
const doMockReadStream = () => {
  mockedCreateReadStream.mockImplementation(() =>
    jest
      .requireActual<{createReadStream: typeof createReadStream}>('fs')
      .createReadStream(Buffer.from('this is a tést'))
  );
};
```

You can also remove `ReadStream` and `Readable` from your imports.

louis-bompart (changes requested):
Suggestion to streamline and avoid casting in the mocks

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+  mockedAuthenticatedClient.mockImplementation(
+    () =>
+      ({
+        getClient: () =>
+          Promise.resolve({
+            resourceSnapshot: {createFromFile: mockedCreateSnapshotFromFile},
+          }),
+      } as unknown as AuthenticatedClient)
+  );
+};
```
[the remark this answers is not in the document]
y-lakhdar: Looks like it still causes the same error. Now the `resourceSnapshot` mock is missing all its methods.

packages/cli/src/lib/snapshot/snapshot.spec.ts
y-lakhdar: Wait a second, this will fail because no such file or directory named 'this is a tést' exist.
```

Error: ENOENT: no such file or directory, open 'this is a tést'
Emitted 'error' event on ReadStream instance at:
    at internal/fs/streams.js:147:14
    at FSReqCallback.oncomplete (fs.js:171:23) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'this is a tést'
}
events.js:292
```

louis-bompart: Hm true, I thought i would 'streamify' the buffer

packages/cli/src/lib/snapshot/snapshot.spec.ts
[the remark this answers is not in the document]
louis-bompart: Hm yeah, the `PlatformClient` was missing some stuff.
Double check but I think this would work tho
```suggestion
const doMockAuthenticatedClient = () =>
  mockedAuthenticatedClient.prototype.getClient.mockImplementation(() =>
    Promise.resolve<PlatformClient>({
      resourceSnapshot: {createFromFile: mockedCreateSnapshotFromFile},
    } as unknown as PlatformClient)
  );

```

louis-bompart (changes requested):

louis-bompart (approved):

packages/cli/src/lib/project/project.spec.ts
```diff
@@ -0,0 +1,113 @@
+jest.mock('@coveord/platform-client');
+jest.mock('fs');
+jest.mock('archiver');
+
+import {mocked} from 'ts-jest/utils';
+import {existsSync, createWriteStream, WriteStream, unlinkSync} from 'fs';
+import {Project} from './project';
+import {join} from 'path';
+import * as archiver from 'archiver';
```
louis-bompart, line 9: Mh, I would enable `esModuleInterop` in the tsconfig and do 
```suggestion
import archiver, {Archiver} from 'archiver';
```

packages/cli/src/lib/project/project.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+const mockedFinalize = jest.fn();
+
+mockedArchiver.mockImplementation(
+  () =>
+    ({
+      on: () => {},
+      pipe: mockedPipe,
+      directory: mockedPassDirectory,
+      finalize: mockedFinalize,
+    } as unknown as archiver.Archiver)
```
louis-bompart, line 27: If you do my suggestion above
```suggestion
    } as unknown as Archiver)
```

packages/cli/src/lib/project/project.ts
```diff
@@ -0,0 +1,54 @@
+import {createWriteStream, existsSync, unlinkSync} from 'fs';
+import {join} from 'path';
+import {cli} from 'cli-ux';
+import * as archiver from 'archiver';
```
louis-bompart, line 4:
```suggestion
import archiver from 'archiver';
```

louis-bompart (commented):
I think it'd be more syntactically correct to do a default import ;)

packages/cli/src/lib/project/project.spec.ts
y-lakhdar: I don't think it will work in this case since there are no default export in the archiver package

louis-bompart: image
Tested on my machine, it works ;)
UT green too
image

y-lakhdar: oups, I updated the wrong tsconfig 🤦

PR #236, opened 2021-05-30
feat: add snapshot class

y-lakhdar:
## Proposed changes

Creating the structure of the snapshot class. The class will be used to manipulate snapshots.

-----
CDX-360


packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -0,0 +1,37 @@
+import {ResourceSnapshotsModel} from '@coveord/platform-client';
+import ResourceSnapshots from '@coveord/platform-client/dist/definitions/resources/ResourceSnapshots/ResourceSnapshots';
```
olamothe, line 2: Is that the right import path ? It can't be imported from root of `@coveord/platform-client` ?

If so, would it make sense to open a PR/TODO for that in platform client project ?

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export class Snapshot {
+  constructor(
+    private model: ResourceSnapshotsModel,
+    private snapshotClient: ResourceSnapshots
```
olamothe, line 7: Instead of injecting the client in the ctor, could it be created internally ? 

However, it does ultimately rely on the parent command to make sure the platform-client is authenticated, so.... I think it's fine like this.

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -0,0 +1,37 @@
...
+  ) {}
+
+  async push(targetOrganizationId: string, developerNotes: string) {
```
olamothe, line 10: Personal pet peeve, but I prefer to have the `public` keyword explicitly set.

olamothe (approved):

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: Correct, it is not available. I will open a PR
image

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: Without the client passed in the constructor, the `Snapshot` class has to know all the org config variables (org, env and region) and create the appropriate client with that info.
However, I wanted the `SnapshotFactory` class to handle the client generation poutine since the `Snapshot` class does not need to know that information.

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: I'll do that correction in #240

packages/cli/src/lib/snapshot/snapshotFactory.ts
```diff
@@ -0,0 +1,14 @@
+import {Snapshot} from './snapshot';
+
+export class SnapshotFactory {
```
louis-bompart, line 3: If the class is stateless, then put every method in it as statics methods.

louis-bompart (changes requested):
Some changes, especially the good export for `@coveord/platform-client` (seems a quickwin, thus  no TODO, but a PR ;) )

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: [CDX-366](https://coveord.atlassian.net/browse/CDX-366)

y-lakhdar: @louis-bompart Here is the pull request : https://github.com/coveo/platform-client/pull/324

louis-bompart (approved):

PR #237, opened 2021-05-31
feat(cli): push snapshots to an organization

y-lakhdar:
## Proposed changes

Once the resources within the project are compressed into a Zip file, they are pushed to the destination org.
If the user does not specify the target org, the snapshot will be pushed to the org defined in the config file.

Since we need to support org-to-org deployment, I decided to add an optional parameter on the `AuthenticatedClient.getClient()` method, so we can instantiate a client that is not tied to the connected org.

## Testing

- [x] Unit Tests: Covered by unit tests
- [x] Manual Tests: Push a snapshot without the `-t` flag. Push a snapshot with the `-t` flag.

-----
CDX-356


packages/cli/src/lib/platform/authenticatedClient.ts
```diff
@@ -29,18 +29,18 @@ export class AuthenticatedClient {
...
 
-  async getClient() {
-    const {
-      environment,
-      region,
-      organization,
-      accessToken,
-    } = await this.cfg.get();
+  async getClient(customConfig?: Partial<Configuration>) {
+    const {environment, region, organization, accessToken} =
```
olamothe, line 33: You could merge the the 2 config at the beginning of the method, and then just reference the final config instead of multiple boolean check.

```
const configFromDisk = await this.cfg.get();
const resolvedConfig = {...configFromDisk, ...customConfig}

return new PlatformClient({environment: resolvedConfig.environment, [...etc...]})
```

packages/cli/src/lib/platform/authenticatedClient.ts
```diff
@@ -29,18 +29,18 @@ export class AuthenticatedClient {
...
   }
 
-  async getClient() {
-    const {
-      environment,
-      region,
-      organization,
-      accessToken,
-    } = await this.cfg.get();
+  async getClient(customConfig?: Partial<Configuration>) {
```
olamothe, line 32: Suggestion: `overrideConfig`

olamothe (approved):

packages/cli/src/commands/org/config/preview.ts
```diff
@@ -41,21 +41,16 @@ export default class Preview extends Command {
...
-      pathToZip,
-      'cli-preview-snapshot'
-    );
+    const targetOrg = await this.getTargetOrg();
 
     cli.action.start('Creating snapshot');
 
-    const targetOrg = await this.getTargetOrg();
-    await snapshot.push(targetOrg, 'cli-preview-snapshot');
+    const factory = new SnapshotFactory();
```
louis-bompart, line 48: would be impacted by my comment about the snapshotFactory, stateless class, no need for `new` ;)

y-lakhdar: Correct. I will be fix once I merge the corrections from CDX-353

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -6,17 +6,11 @@ export class Snapshot {
...
-    // snapshotClient.push(this.lastSnapshot?.targetId, {targetOrganisationId});
-  }
-
-  validate(targetOrganisationId: string) {
+  async validate() {
     // TODO: CDX-358: Validate snapshot
   }
 
-  preview() {
+  async preview() {
```
louis-bompart, line 13: why async?

packages/cli/src/lib/snapshot/snapshotFactory.ts
```diff
@@ -12,15 +12,15 @@ export interface CustomFile extends ReadStream {
 export class SnapshotFactory {
   static async createFromZip(
     pathToZip: string,
-    developerNotes: string
+    targetOrg: string
   ): Promise<Snapshot> {
-    const client = await new AuthenticatedClient().getClient();
+    const client = await this.getClient(targetOrg);
```
louis-bompart, line 17:
```suggestion
    const client = await SnapshotFactory.getClient(targetOrg);
```

I think?

PR #240, opened 2021-05-31
feat(cli): validate snapshots

y-lakhdar:
## Proposed changes

Validate a snapshot right after it is being pushed to an organization.
This PR does not handle all the invalid cases. It considers the snapshot valid if we get a `SUCESS` result code from the API. Otherwise, the snapshot is considered invalid. [CDX-362](https://coveord.atlassian.net/browse/CDX-362) will handle all the possible invalid states of a snapshot.

There is also a patch for the platform-client until the pull request [322](https://github.com/coveo/platform-client/pull/322) is merged. Without that change, it is not possible to get the status report of a snapshot.

## Testing
I split the snapshot tests into 2 files:
1. snapshot.spec.ts
2. snapshotFactory.spec.ts

- [x] Unit Tests:
- [x] Manual Tests: Manually tested valid and invalid snapshots.

## Acceptance Criteria 
- [x] Merge https://github.com/coveo/platform-client/pull/322

-----
CDX-358


packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -8,7 +15,19 @@ export class Snapshot {
...
   async validate() {
-    // TODO: CDX-358: Validate snapshot
+    await this.client.dryRun(this.snapshotId, {
+      deleteMissingResources: false, // TODO: CDX-361: Add flag to support missing resources deletion
+    });
+
+    await this.waitUntilIdle();
+  }
+
+  isValid() {
```
olamothe: The class interface here is a bit weird, because you have to call `validate()`, then you can call `isValid()` afterwards.

As a user of this class, you could very well try to call `isValid()` without first doing `validate()`, and then you'd get incorrect results.

I think we could refactor this so `validate()` returns true or false, and we could remove the `isValid()` method.

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -28,4 +47,53 @@ export class Snapshot {
   async delete() {
     // TODO: CDX-359: Delete snapshot once previewed
   }
+
+  get lastestReport(): ResourceSnapshotsReportModel {
```
olamothe: By refactoring it so `validate()` control the whole flow, instead of 2 method, this check would go away.

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -28,4 +47,53 @@ export class Snapshot {
...
+    }
+    return this.model.reports.slice(-1)[0];
+  }
+
+  private get snapshotId() {
+    return this.model.id;
+  }
+
+  private async refreshSnapshotData() {
+    this.model = await (this.client.get as any)(this.model.id, {
```
olamothe, line 65: For the `as any` cast, is this a bug in platform-client ? Should this be a TODO ?

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -28,4 +47,53 @@ export class Snapshot {
...
+      includeReports: false,
+    });
+  }
+
+  private async isIdle() {
+    await this.refreshSnapshotData();
+    const lastReport = this.model.reports?.slice(-1)[0];
```
olamothe: That seems very similar to `get lastestReport()` ? 🤔

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: yes, a  bug in the platform-client. I created a Pull request and added the link to the the jira in the comments

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: yeah, that was a bad commit. I replaced it with the get latestReport :P

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: Pull request : https://github.com/coveo/platform-client/pull/322

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -24,7 +55,52 @@ export class Snapshot {
     // TODO: CDX-347 Display Expanded preview
   }
 
-  async delete() {
-    // TODO: CDX-359: Delete snapshot once previewed
+  private get lastestReport(): ResourceSnapshotsReportModel {
```
olamothe, line 58:
```suggestion
  private get latestReport(): ResourceSnapshotsReportModel {
```

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -24,7 +55,52 @@ export class Snapshot {
...
+    if (this.model.reports === undefined) {
+      throw new Error(
+        `No detailed report found for the snapshot ${this.snapshotId}`
+      );
+    }
+    return this.model.reports.slice(-1)[0];
```
olamothe, line 64: Anything we might want to handle as far as reports array being empty ?
Is it even possible ?

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -24,7 +55,52 @@ export class Snapshot {
...
+
+  private async refreshSnapshotData() {
+    this.model = await this.snapshotClient.get(this.model.id, {
+      includeReports: true,
+    });
+  }
+
+  private async isIdle() {
+    await this.refreshSnapshotData();
+    return new Promise<void>((resolve) => {
```
olamothe, line 83: I'm not sure I understand why this is wrapped in a Promise ?

There's no rejection that would trigger the `waitUntilIdle` backoff mechanism, so... 🤔

olamothe (approved):

olamothe:
💯  Gj !

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: My bad. I forgot the `reject()`. I am not sure that is a good usage of the `exponential-backoff` package (as opposed to async-retry).

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: Probably a question to the backend team.
I included the `includeReport` query parameter on every GET request to the snapshot API. 

I'll ask for confirmation

packages/cli/patches/@coveord+platform-client+18.15.2.patch
```diff
@@ -0,0 +1,54 @@
+diff --git a/node_modules/@coveord/platform-client/dist/definitions/resources/ResourceSnapshots/ResourceSnapshots.d.ts b/node_modules/@coveord/platform-client/dist/definitions/resources/ResourceSnapshots/ResourceSnapshots.d.ts
```
louis-bompart, line 1: I'm really not in favor of doing a patch on dependencies where we have a bit of control (I mean, it's in the cie)
I can't approve this PR with this.

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -1,28 +1,75 @@
...
+    id: snapshotId,
+    createdBy: 'user@coveo.com',
+    createdDate: 1622555047116,
+    targetId: orgId,
+    developerNote: 'hello',
+    reports: reports,
+    synchronizationReports: [],
+    contentSummary: {EXTENSION: 1, FIELD: 11},
+  };
 };
```
louis-bompart, line 73: Not necessarly prettier compliant (i.e. might need a alt-shift-f in vscode)
But yeah, useless to do a full block function imo.

```suggestion
): ResourceSnapshotsModel => {
  {
    id: snapshotId,
    createdBy: 'user@coveo.com',
    createdDate: 1622555047116,
    targetId: orgId,
    developerNote: 'hello',
    reports: reports,
    synchronizationReports: [],
    contentSummary: {EXTENSION: 1, FIELD: 11},
  };
```

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -46,49 +95,101 @@ const doMockAuthenticatedClient = () => {
...
+        targetOrgId,
+        snapshotId,
+        [successReport]
+      );
+
+      mockedGetSnapshot.mockResolvedValueOnce(futureSnapshotState);
+    };
+
+    beforeEach(async () => {
+      doMockGetSnapshotWithoutError();
```
louis-bompart, line 178: await?
```suggestion
      await doMockGetSnapshotWithoutError();
```

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -46,49 +95,101 @@ const doMockAuthenticatedClient = () => {
...
+        [errorReport]
+      );
+
+      mockedGetSnapshot.mockResolvedValueOnce(futureSnapshotState);
+    };
 
     beforeEach(async () => {
-      await SnapshotFactory.createFromZip(pathToZip, 'my-target-org');
-    });
+      doMockGetSnapshotInError();
```
louis-bompart, line 119: await?
```suggestion
      await doMockGetSnapshotInError();
```

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -24,7 +55,54 @@ export class Snapshot {
...
+          ResourceSnapshotsReportStatus.Aborted,
+          ResourceSnapshotsReportStatus.Completed,
+        ].includes(this.latestReport.status)
+      ) {
+        resolve();
+      } else {
+        reject();
+      }
+    });
+  }
```
louis-bompart, line 95: Mhhh. Not sure here, what `isIdle` for you in layman terms?

louis-bompart (changes requested):
Some pet peeves, but mostly blocking approval due to the patch (tho it's fine for ongoing dev)

packages/cli/patches/@coveord+platform-client+18.15.2.patch
y-lakhdar: Blocked by https://github.com/coveo/cli/pull/240/

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: Meaning that snapshot is not in a pending state.
In the platform, that pending state is translated with a loading animation (see image).
image

The role of that `isIdle` function is to keep fetching the snapshot status until the report indicates a failure or a success.
But I agree, we can change that name :P

packages/cli/src/lib/snapshot/snapshot.spec.ts
y-lakhdar: mmh non. I think the doMockGetSnapshotInError should not return a promise

y-lakhdar: Good catch tho

packages/cli/src/lib/snapshot/snapshot.ts
louis-bompart: > Meaning that the snapshot is not being processed by some kind of operation (validation, deployment, being applied). That it had reached a finish state.

Is there a defined status for that in `ResourceSnapshotsReportStatus` ?

The thing that I found odd tbh is that `isIdle` resolves if the status is Aborted or Completed.

I interpret that as `isIdle` is 'true' if the status is Completed or Aborted. Which is strange imho

also resolve, reject, meh. I think a resolve w/ a boolean would be more semantically correct with the `isXX` name.

y-lakhdar: `isIdle` doesn't mean the snapshot is valid. It only means that it has settled from its last operation. We ca also go the other way around and have a method `isPending` that will check if the snapshot has one of the following values `Pending`, `InProgress`. 

I agree about the resolve reject. The other way would be to throw an error if the snapshot is still not idle

louis-bompart: > `isIdle` doesn't mean the snapshot is valid. It only means that it has settled from its last operation. We can also go the other way around and have a method `isPending` that will check if the snapshot is in one of the following states `PENDING`, `IN_PROGRESS`.
> 
> I agree about the resolve reject. The other way would be to throw an error if the snapshot is still not idle

Either way, I would name the method maybe `isComplete` or `isDone`
Idle I interpret this as "there's something that will go on, but not right now"

y-lakhdar: `isDone` it is.

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -1,92 +1,190 @@
...
+    id: snapshotId,
+    updatedDate: 1622555847000,
+    type: type,
+    status: status,
+    resultCode: resultCode,
+    resourcesProcessed: 12,
+    resourceOperations: {},
+    resourceOperationResults: {},
+  };
+};
```
louis-bompart, line 38:
```suggestion
): ResourceSnapshotsReportModel => {
  id: snapshotId,
  updatedDate: 1622555847000,
  type: type,
  status: status,
  resultCode: resultCode,
  resourcesProcessed: 12,
  resourceOperations: {},
  resourceOperationResults: {},
};
```

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -1,92 +1,190 @@
...
+
+const getSuccessReport = (snapshotId: string): ResourceSnapshotsReportModel => {
+  return getReport(
+    snapshotId,
+    ResourceSnapshotsReportType.DryRun,
+    ResourceSnapshotsReportStatus.Completed,
+    ResourceSnapshotsReportResultCode.Success
+  );
+};
```
louis-bompart, line 47:
```suggestion
const getSuccessReport = (snapshotId: string): ResourceSnapshotsReportModel =>
  getReport(
    snapshotId,
    ResourceSnapshotsReportType.DryRun,
    ResourceSnapshotsReportStatus.Completed,
    ResourceSnapshotsReportResultCode.Success
  );
```

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -1,92 +1,190 @@
...
+
+const getErrorReport = (snapshotId: string): ResourceSnapshotsReportModel => {
+  return getReport(
+    snapshotId,
+    ResourceSnapshotsReportType.DryRun,
+    ResourceSnapshotsReportStatus.Completed,
+    ResourceSnapshotsReportResultCode.ResourcesInError
+  );
```
louis-bompart, line 55:
```suggestion
const getErrorReport = (snapshotId: string): ResourceSnapshotsReportModel =>
  getReport(
    snapshotId,
    ResourceSnapshotsReportType.DryRun,
    ResourceSnapshotsReportStatus.Completed,
    ResourceSnapshotsReportResultCode.ResourcesInError
  );
```

louis-bompart (commented):
Small suggestions, it's flavouring tho, so feel free to ignore them!
Other than that, as mentioned earlier yeah there's the `isIdle`/`isDone` thing.
Rest is A-OK, good job! 👏

packages/cli/patches/@coveord+platform-client+18.15.2.patch
y-lakhdar: Not blocked anymore

louis-bompart (approved):

PR #241, opened 2021-05-31
style(eslint): enforce explicit member accessibility

louis-bompart:
## Proposed changes

Enforce explicit member accessibility, meaning that if a member declaration is not prefixed by public/protected/private, ESLint will throw a fit.

Put public everywhere where the prefix was missing, because that's the default.

## Testing

- Automated, ESLINT.

-----
CDX-363


louis-bompart:
Special mention for @olamothe pet-peeves :P (tho I fully agree on this one 😂)

olamothe (approved):

y-lakhdar (approved):
Now, I'll never forget a public/protected/private keyword 😂

PR #243, opened 2021-06-01
test(e2e): run PR workflow on Linux and Windows

louis-bompart:
## Proposed changes

Adapted tests to run on Windows and replicated setup for Windows.
Setup a run matrix to do stuff in parallel.

## Testing

- CI 

-----
CDX-273


.github/workflows/build.yml
```diff
@@ -7,18 +7,31 @@ on:
...
+    runs-on: ${{matrix.os}}
+    strategy:
+      fail-fast: true
+      matrix:
+        os: ['ubuntu-latest', 'windows-latest']
+        node: [14]
     env:
       PLATFORM_USER_NAME: ${{ secrets.PLATFORM_USER_NAME }}
       PLATFORM_USER_PASSWORD: ${{ secrets.PLATFORM_USER_PASSWORD }}
+      COVEO_DISABLE_AUTOUPDATE: true
```
olamothe, line 19: What is this for ? I searched in the current codebase and could not find anything.

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -228,15 +228,19 @@ describe('ui:create:angular', () => {
         projectName
       );
       await serverProcessManager.killAllProcesses();
-    }, 5e3);
+    }, 5 * 60e3);
```
olamothe, line 231: If it's important to bump the timeout for each, I'd just have a const somewhere at the top fo the file, used for all tests.

Also, some tests seems to have a different timeout. Is it important ? Could we just use the same value for each ?

packages/cli-e2e/__tests__/auth.specs.ts
```diff
@@ -24,9 +24,14 @@ describe('auth', () => {
 
     it('should open the platform page', async () => {
       // TODO CDX-98: Remove `-e=dev`.
+      const args: string[] = [CLI_EXEC_PATH, 'auth:login', '-e=dev'];
+      if (process.platform === 'win32') {
```
olamothe, line 28: Are those platform string typed in any manner ? 

If not, a constant from node that we could use ?

packages/cli-e2e/__tests__/react.specs.ts
```diff
@@ -87,35 +94,24 @@ describe('ui:create:react', () => {
 
   const startApplication = async (
     processManager: ProcessManager,
-    debugName = 'react-server',
-    errorCallback: (error: string) => void = () => {}
+    debugName = 'react-server'
   ) => {
+    const args = ['npm', 'run', 'start'];
+    if (process.platform === 'win32') {
```
olamothe, line 100: Similar comment as above about the platform constant.

packages/cli-e2e/__tests__/react.specs.ts
```diff
@@ -87,35 +94,24 @@ describe('ui:create:react', () => {
...
+      args[0] = npmJsPath();
+      args.unshift('node');
```
olamothe, line 102: This seems very similar to above, where we need to shift the "node" arg for windows.

Could benefit from a `makeArgsPlatformAgnostic(args)` kinda utility function ?

packages/cli-e2e/utils/cli.ts
```diff
@@ -48,9 +48,15 @@ export function setupUIProject(
     console.log('Testing with published version of the template');
   }
 
+  const args = [CLI_EXEC_PATH, ...command];
+
+  if (process.platform === 'win32') {
```
olamothe, line 53: Could also benefit from a make agnostic args function here

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -137,37 +136,33 @@ describe('ui:create:angular', () => {
...
+    const args = ['npm', 'run', 'start'];
+    if (process.platform === 'win32') {
+      args[0] = npmJsPath();
+      args.unshift('node');
+    }
+
     const serverTerminal = new Terminal(
-      'npm',
-      ['run', 'start'],
+      args.shift()!,
```
olamothe, line 148: I'm seeing that this is repeated for each invocation of the `Terminal` constructor.

Could that be handled internally by the class ? So that we don't have to "remember" to do it from a user POV.

olamothe (approved):
💯  Seems to have been very painful to do all of that, but it's awesome that you were able to tackle all the problems.

Job well done !

.github/workflows/build.yml
louis-bompart: It's a hidden flag in `oclif/plugin-update`:

packages/cli-e2e/__tests__/angular.specs.ts
louis-bompart: > If it's important to bump the timeout for each, I'd just have a const somewhere at the top fo the file, used for all tests.
That could make sense
> Also, some tests seem to have a different timeout. Is it important? Could we just use the same value for each?
Tests have different timeout depending on the 'complexity of the task at hand.

For example, everytimes we handle processes, we need to plan extra time because we ought to go through the PowerShell layer.

A more extreme example is the big beforeAll, where we instantiate the project (i.e. run the actual CLI method).
Because it often implies one or more `npm install` or other under the hood, it's quite lengthy (around 5 to 10 minutes).

We should not have 5-10' assertions time for our `it` therefore, a 'unique' timeout is not a good idea, because if doing npm start takes 5', there might be a problem.

This make me think we should not have one timeout fit for all, because it might hide some performance issues, in the test or in the framework.

packages/cli-e2e/__tests__/angular.specs.ts
louis-bompart: Alas no, because the arg shifting is only for `npm` related task.

packages/cli-e2e/__tests__/react.specs.ts
louis-bompart: Yeah, that's something I thought about.
I'd like to revisit this in the future because I'm unhappy with how I handle that (i.e. replicating what `npm.cmd` does in the terminal to remove the node layer, which lessens the 'end-to-end' aspect.)

I'll get a look tho to refactor this.

packages/cli-e2e/__tests__/auth.specs.ts
louis-bompart: the constants are typed, but not only typed (i.e. it's a `declare const` but it's not actually exposed afaik)

PR #248, opened 2021-06-06
feat: setup a synchronization plan if the resources are in error

y-lakhdar:
## Proposed changes

If the snapshot and the target organization are not in sync, the command will print an URL to the Admin Console so the user can synchronize his organization.
The CLI is not handling any synchronization logic at the moment.

## Testing

- [x] Unit Tests:
    
- [ ] Functionnal Tests:
    
- [ ] Manual Tests:
    

-----
CDX-370


packages/cli/src/commands/org/config/preview.ts
```diff
@@ -93,7 +75,49 @@ export default class Preview extends Command {
...
+
+    if (snapshot.requiresSynchronization()) {
+      cli.action.start('Synchronization');
+
+      const synchronizationPlanUrl = await this.getSynchronizationPage(
+        snapshot
+      );
+      this.warn(
+        dedent`Some conflicts were detected while comparing changes with the organization.
+        Click on the URL bellow to synchronize your organization before running the command again.
```
louis-bompart, line 92:
```suggestion
        dedent`Some conflicts were detected while comparing changes between the snapshot and the target organization.
        Click on the URL below to synchronize your organization before running the command again.
```

louis-bompart: Tho I think @jpmarceau expert eye would be welcome here :D

packages/cli/src/commands/org/config/preview.ts
```diff
@@ -93,7 +75,49 @@ export default class Preview extends Command {
     return cfg.organization;
   }
 
+  private async handleInvalidSnapshot(snapshot: Snapshot) {
+    // TODO: CDX-362: handle invalid snashot cases
```
louis-bompart, line 79: can be ignored, 50% suggestion, 50% trolling :P 
```suggestion
    // TODO: CDX-362: handle invalid snapshot cases
```

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -39,10 +41,41 @@ export class Snapshot {
...
+      'snapshot-reports',
+      `${this.latestReport.id}.json`
+    );
+    ensureFileSync(pathToReport);
+    writeJsonSync(pathToReport, this.latestReport, {spaces: 2});
+    return pathToReport;
+  }
+
+  public get latestReport(): ResourceSnapshotsReportModel {
+    if (this.model.reports === undefined || this.model.reports.length === 0) {
```
louis-bompart, line 65: Given what's on line 68:
```suggestion
    if (!Array.isArray(this.model.reports) || this.model.reports.length === 0) {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

louis-bompart (approved):
Code looks good!
I tagged @jpmarceau 'cause I ain't sure about the message tho

packages/cli/src/commands/org/config/preview.ts
jpmarceau: Building on Louis' suggestion:

jpmarceau, line 92:
```suggestion
        Click on the URL bellow to synchronize your snapshot with your organization before running the command again.
```

jpmarceau (approved):

olamothe (approved):
💯  !

PR #254, opened 2021-06-09
feat: add light preview

y-lakhdar:
## Proposed changes

Show snapshot preview in the console.

Example of a snapshot that is not in sync with the target org
image

Exemple of a valid snapshot
image

## Testing

- [x] Unit Tests:
    
- [x] Functionnal Tests:
    
- [x] Manual Tests:
    

## Blocked by 
- [x]  [CDX-382](https://github.com/coveo/platform-client/pull/331).

-----
CDX-346


packages/cli/src/lib/snapshot/reportViewer.spec.ts
```diff
@@ -140,15 +140,14 @@ describe('ReportViewer', () => {
...
-            '  (2 more errors)',
-          ].join(EOL)
+          dedent`
+          Fields
+            • RESOURCE_ALREADY_EXISTS: Field foo already exists.
+            • RESOURCE_ALREADY_EXISTS: Field bar already exists.
+            • RESOURCE_ALREADY_EXISTS: Field dsads already exists.
+            • RESOURCE_ALREADY_EXISTS: Field fdww already exists.
+            • RESOURCE_ALREADY_EXISTS: Field csad already exists.
+            (2 more errors)`
```
louis-bompart, line 150: image

packages/cli/src/lib/snapshot/reportViewer.spec.ts
```diff
@@ -160,25 +159,25 @@ describe('ReportViewer', () => {
...
       .do(() => {
         viewer.display();
       })
       .it('should print resource changes', (ctx) => {
-        const trimedStdout = ctx.stdout.replace(/\n\n/g, '');
+        // Remove padding added by cli-ux so we can test the text and not the padding on the line
+        const trimedStdout = ctx.stdout
+          .split('\n')
+          .map((s) => s.trimEnd())
+          .join('\n');
```
louis-bompart, line 171: You can save one operation and be 'OS safe' no matter fricking what with `$` instead of `\n`
image
Because the text still has the \n at the beginning, you can skip the join afterward ;)

```suggestion
        const trimedStdout = ctx.stdout
          .split(/$/m)
          .map((s) => s.trimEnd());
```

louis-bompart (changes requested):
I would simplify the trimedStdout, other than that 👍

packages/cli/src/commands/org/config/preview.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+  // TODO: use prototype
+  mockedConfig.mockImplementation(
+    () =>
+      ({
+        get: mockedConfigGet,
+      } as unknown as Config)
+  );
+};
+
+const mockSnapshotFactory = async (validResponse: any) => {
```
olamothe, line 62: If the type is not really important here, I would put `unknown` (it's a bit better than any)

packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -0,0 +1,185 @@
...
+  ResourceSnapshotsReportModel,
+  ResourceSnapshotsReportOperationModel,
+  ResourceSnapshotsReportOperationResults,
+  ResourceSnapshotsReportResultCode,
+} from '@coveord/platform-client';
+import {cli} from 'cli-ux';
+import {bgHex, green, yellow, red, bold, italic} from 'chalk';
+
+export class ReportViewer {
+  private style = {
```
olamothe, line 11: nit: `styles`, and could cleaner as a static method.

packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -0,0 +1,185 @@
...
+      },
+    });
+  }
+
+  // TODO: Change logic once SRC-4448 is complete
+  private printTableSection(row: {
+    resourceName: string;
+    operations: ResourceSnapshotsReportOperationModel;
+  }) {
+    const resourceType = this.prettyPrintResourceName(row.resourceName);
```
olamothe, line 48: This method works, but I think it's going to be somewhat tricky to change/evolve over time, because it relies on precise string alignment baked into each string output. And could be easy to end up with things un-aligned in the console when we do a change.

One potential refactor would be to create a class for `PreviewDisplay` (or PreviewFormatter).

Something like this (very crude pseudo code):
```

class PreviewDisplayRow() {
     contructor(text, type) {
        // where type is one of success/warning/error, etc.
     }

    public render() {
       // return string color formatted     
    }
}


class PreviewDisplay() {
   private rows: PreviewDisplayRow[];

   public addRow(text, type) {
         this.rows.push(new PreviewDisplayRow(text, type);
         return this;
    }

   render() {
       // call render on each row, and format them on individual line
   }
}
```

Do you think it would be helpful for the more "advanced" preview if we already have this created ?
Or that it would needlessly complicate things ?

packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  private get changedResources() {
+    type resourceEntries = [string, ResourceSnapshotsReportOperationModel];
+    const resourceHasAtLeastOneOperation = ([
+      _,
+      operations,
+    ]: resourceEntries) => {
+      return (
+        operations.resourcesCreated +
```
olamothe, line 88: Some weird alignment in here ? 🤔

packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -0,0 +1,185 @@
...
+      (count: number, current: ResourceSnapshotsReportOperationModel) =>
+        count + current[type],
+      0
+    );
+
+    return count;
+  }
+
+  private handleReportErrors() {
+    const maximumNumberOfErrorsToPrint = 5;
```
olamothe, line 134: nit: Extract this as a static const on `ReportViewer`

packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -0,0 +1,185 @@
...
+      const hasNoError = (op: ResourceSnapshotsReportOperationResults) =>
+        Object.keys(op).length === 0;
+
+      if (hasNoError(operationResult)) {
+        continue;
+      }
+
+      cli.log(`\n ${this.prettyPrintResourceName(resourceType)}`);
+
+      for (const resourceInError in operationResult) {
```
olamothe, line 159: Could this be simplified by splicing the array of errors ?

```
const toDisplay = operationResult.splice(0, maximumNumberOfErrorsToPrint);

// display each "toDisplay"
// display operationResult.length error left/unprinted
```

You can also clone operationResult first if you do not want to modify the original array.

olamothe (approved):

packages/cli/src/lib/snapshot/reportViewer.ts
y-lakhdar: @olamothe I get your point, but I don't think we need to worry about the alignment as it is already taken care of by the `cli-ux` package. The `cli.table` util will compute the length of the longest row and will apply the necessary adjustments. It also supports multi-line rows (as we have here).

packages/cli/src/lib/snapshot/reportViewer.ts
y-lakhdar: It's not that simple. `operationResult` is a record of record of array. There are 2 ways we can reach the maximum number of errors to print:

1. Having more than 5 entries within the same **resource type** (first-level record)
2. Having multiple errors within the same **resource** (second-level record)

That's why I am using nested for loops even if it's not pretty :S 

Here is an example:
```
  resourceOperationResults: {
    FIELD: {
      foo_4VNj5ds5: ['RESOURCE_ALREADY_EXISTS: Field foo already exists.'],
      bar_4VNj5ds5: ['RESOURCE_ALREADY_EXISTS: Field bar already exists.'],
      dsads_4VNj5ds5: ['RESOURCE_ALREADY_EXISTS: Field dsads already exists.'],
      fdww_4VNj5ds5: [
        'RESOURCE_ALREADY_EXISTS: Field fdww already exists.',
        'ERROR_CODE: another error',
        'ERROR_CODE: another another error',
        'ERROR_CODE: another another another error',
        'ERROR_CODE: another another another another error',
        'ERROR_CODE: another another another another another error',
      ],
      csad_4VNj5ds5: ['RESOURCE_ALREADY_EXISTS: Field csad already exists.'],
      hjkd_4VNj5ds5: ['RESOURCE_ALREADY_EXISTS: Field hjkd already exists.'],
      fdasf_4VNj5ds5: ['RESOURCE_ALREADY_EXISTS: Field fdasf already exists.'],
    },
  },
```

And it should print
```
* RESOURCE_ALREADY_EXISTS: Field foo already exists
* RESOURCE_ALREADY_EXISTS: Field bar already exists
* RESOURCE_ALREADY_EXISTS: Field dsads already exists
* RESOURCE_ALREADY_EXISTS: Field fdww already exists
* ERROR_CODE: another error
* ERROR_CODE: another another error
```

louis-bompart (approved):

PR #255, opened 2021-06-09
Update reportViewer.spec.ts

louis-bompart:
## Testing

- [ ] Unit Tests:
    
- [ ] Functionnal Tests:
    
- [ ] Manual Tests:


y-lakhdar:
Thanks, but I got rid of the regex and used template string instead. much simpler, much cleaner

PR #256, opened 2021-06-10
fix: fix race condition between create and validate

y-lakhdar:
## Proposed changes

There is an appropriate wait after the package is created. In these cases, the validation occurs to fast and the Snapshot class ends up returning the wrong report.

## Testing

- [x] Unit Tests:
    
- [ ] Functionnal Tests:
    
- [x] Manual Tests:
    

-----
CDX-387


packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -27,7 +28,7 @@ export class Snapshot {
       deleteMissingResources: false, // TODO: CDX-361: Add flag to support missing resources deletion
     });
 
-    await this.waitUntilDone();
+    await this.waitUntilDone(ResourceSnapshotsReportType.DryRun);
```
y-lakhdar, line 31: Making the wait more robust by specifying the operation to wait for

packages/cli/src/lib/snapshot/snapshotFactory.ts
```diff
@@ -27,7 +30,11 @@ export class SnapshotFactory {
       file,
       computedOptions
     );
-    return new Snapshot(model, client);
+    const snapshot = new Snapshot(model, client);
+
+    await snapshot.waitUntilDone(ResourceSnapshotsReportType.CreateSnapshot);
```
y-lakhdar, line 35: This missing line was the reason why it was failing from time to time

packages/cli/src/lib/snapshot/snapshot.ts
louis-bompart: In that case, I would rename the method accordingly to be explicit:
`waitUntilOperationIsDone(operationType: ResourceSnaphostReportType)`

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -116,7 +117,9 @@ export class Snapshot {
           ResourceSnapshotsReportStatus.InProgress,
         ].includes(this.latestReport.status);
 
-        if (isNotDone) {
+        const isRightOperationType = this.latestReport.type === type;
+
+        if (isNotDone || !isRightOperationType) {
```
louis-bompart, line 122: I would check for `isRightOperationType` before trying to initialize `isNotDone`, and throw if it ain't the rightOperationType (although I think tbh that having a variable here is really too much 😅 )

So something akin to
```ts
if (this.latestReport.type !== type) {
  throw new Error('Snapshot is still being processed'); // or a more appropriate message maybe?
}

const isNotDone = [
  ResourceSnapshotsReportStatus.Pending,
  ResourceSnapshotsReportStatus.InProgress,
].includes(this.latestReport.status);

if (isNotDone) {
  throw new Error('Snapshot is still being processed');
}
```

The boon here is that you save a declaration and an array manipulation.

If you want to really polish off this, you could even:
 - Declare a private static member containing the status.
 
 You would save the declaration and also, you would clean up the code:
 
```ts
private static ongoingReportStatuses = [
  ResourceSnapshotsReportStatus.Pending,
  ResourceSnapshotsReportStatus.InProgress,
];
/* .... */
/* .... */
/* .... */
public async waitUntilDone(type: ResourceSnapshotsReportType) {
  const waitPromise = backOff(
    async () => {
      await this.refreshSnapshotData();
      
      if (this.latestReport.type !== type) {
        throw new Error('Snapshot is still being processed'); // or a more appropriate message maybe?
      }

      if (Snapshot.ongoingReportStatutes.includes(this.latestReport.status)) {
        throw new Error('Snapshot is still being processed');
      }
    }
  )
}
```

louis-bompart (changes requested):
While at it, I would do some cleanup ;)

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: 👌

olamothe (approved):

louis-bompart (approved):

PR #258, opened 2021-06-11
feat: add flag to support delete changes

y-lakhdar:
## Proposed changes

Add a flag to show missing resources. By default the Snapshot API will not delete missing resources in the target org unless specified. Similarly, the preview command in the CLI will not show resources to delete unless the user specifically adds the flag to the command.

### Example
Without the flag
image

With the flag
image

> Notes: I also adde

## Testing

- [x] Unit Tests:
    
- [x] Functionnal Tests:
    
- [x] Manual Tests:
    

-----
CDX-361


packages/cli/src/commands/org/config/preview.ts
```diff
@@ -37,6 +37,12 @@ export default class Preview extends Command {
       default: cwd(),
       required: false,
     }),
+    showResourcesToDelete: flags.boolean({
+      char: 'd',
+      description: 'Whether or not to show resources to delete',
+      default: false,
```
olamothe, line 43: Instead of setting a default here, could we instead rely on "the default of the API" ?

So, if the flag is not specified, we just don't pass anything to the API, and it will decide if a deletion should happen or not.

packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -42,43 +66,62 @@ export class ReportViewer {
...
+    },
+    indentation = 3
+  ) {
     const resourceType = this.prettyPrintResourceName(row.resourceName);
-    let output = `   ${resourceType}\n`;
+    let output = `${''.padStart(indentation)}${resourceType}\n`;
 
-    if (row.operations.resourcesCreated > 0) {
+    if (
+      this.operationsToDisplay.resourcesCreated &&
```
olamothe, line 80: Suggestion: For readability sake, since it's all the same check but just on a different section of the report we could add a :

```
private shouldDisplayOperation(operationsToDisplaySection, operationSection): boolean {
  [...]
}

if(shouldDisplayOperation(this.operationsToDisplay.resourcesCreated, row.operations.resourcesCreated) {
   [...]
}
```

packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -42,43 +66,62 @@ export class ReportViewer {
...
+      row.operations.resourcesCreated > 0
+    ) {
       output += `${ReportViewer.styles.green(
-        '+'
-      )}   ${ReportViewer.styles.green(
+        '+'.padEnd(indentation + 1)
```
olamothe, line 84: Same as above, there's quite a bit of repetition to build the output.

Suggestion: Just a small util method that accept a color, a sign and an indentation could make this easier to read ?

olamothe (approved):

louis-bompart (commented):
Oli idea for the refactor, other than that lgtm!

packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -6,8 +6,22 @@ import {
...
 export class ReportViewer {
+  public static defaultOperationsToDisplay: ReportViewerOperationsToDisplay = {
+    resourcesCreated: true,
+    resourcesDeleted: true,
+    resourcesInError: true,
+    resourcesRecreated: true,
+    resourcesUnchanged: true,
+    resourcesUpdated: true,
+  };
+
```
louis-bompart, line 22: Here, I do not understand really the true/false 🤔 

I think the presence or lack thereof of the key (e.g. `resourcesCreated`) would work.

With this, you could do something like:
```suggestion
export type ReportViewerOperationsToDisplay = keyof ResourcesSnapshotsReportOperationModel //maybe not exactly that, but something similar.

export class ReportViewer {
  public static defaultOperationsToDisplay: ReportViewerOperationsToDisplay[] = [
    'resourcesCreated',
    'resourcesDeleted',
    'resourcesInError',
    'resourcesRecreated',
    'resourcesUnchanged',
    'resourcesUpdated',
  };
```

louis-bompart: This would of require to change your things to an array overall.

louis-bompart (changes requested):
I think there something to be made here :)

packages/cli/src/lib/snapshot/reportViewer.ts
y-lakhdar: I'm fine with that

louis-bompart (approved):

PR #259, opened 2021-06-11
feat: add unchanged resources

y-lakhdar:
## Proposed changes

There is a difference between an unchanged resource and a resource that is not showing in the report.
A unchanged resource means the resource from the snapshot has a match in the destination org but is identical.

Similar to other IaC tools, unchanged resources should be display in the preview so the user understand these resources were processed.

> Note: I did not know what color to use for unchanged resources. Maybe grey is not the right one...

## Testing

- [ ] Unit Tests:
    
- [ ] Functionnal Tests:
    
- [ ] Manual Tests:
    

-----
CDX-389


packages/cli/src/lib/snapshot/reportViewer.ts
```diff
@@ -27,6 +27,7 @@ export class ReportViewer {
     green: (txt: string) => green(txt),
     yellow: (txt: string) => yellow(txt),
     red: (txt: string) => red(txt),
+    gray: (txt: string) => gray(txt),
```
olamothe, line 30: I think gray is good 🤔 

I've asked people in another channel and that was their first instinct. Maybe just pick complimentary color for all of them ? It could be easier on the eye.

example: https://getbootstrap.com/docs/5.0/components/alerts/

Suggestion: Instead of pure `green/yellow/red` use those hex.

olamothe (approved):

louis-bompart (approved):

packages/cli/src/lib/snapshot/reportViewer.ts
y-lakhdar: I tried Bootstrap hex colors but it doesn't look too good in a dark terminal.
It's either too bright or too dark. 

Chalk default colors (green, yellow, red) seem to be widely used across other packages.

y-lakhdar:
Was added in #258 while refactoring reportViewer

PR #260, opened 2021-06-12
feat: add push command

y-lakhdar:
## Proposed changes

Add a command to apply a snapshot to an organization. It uses the same flow as the preview command and prompt a message to the user asking to apply the changes.

The user needs to approve for the snapshot to be applied.

Example:
image

## Testing

- [ ] Unit Tests:
    
- [ ] Functionnal Tests:
    
- [x] Manual Tests:
    

-----
CDX-348


packages/cli/src/commands/org/config/push.ts
```diff
@@ -0,0 +1,159 @@
...
+      await this.handleValidReport(snapshot);
+      await snapshot.delete();
+    } else {
+      await this.handleReportWithErrors(snapshot);
+    }
+
+    project.deleteTemporaryZipFile();
+  }
+
+  public async getTargetOrg() {
```
y-lakhdar, line 80: There are a lot of copy-pasted code from the preview command. I'm not sure if a util file would be necessary to prevent duplicating code in both `preview` and `push` command

packages/cli/src/commands/org/config/push.ts
```diff
@@ -0,0 +1,159 @@
...
+      description: 'The path to your Coveo project.',
+      helpValue: '/Users/Me/my-project',
+      default: cwd(),
+      required: false,
+    }),
+    deleteMissingResources: flags.boolean({
+      char: 'd',
+      description:
+        'Whether or not to delete resources missing in the destination',
+      default: false,
```
olamothe, line 45: Same comment as the other PR( https://github.com/coveo/cli/pull/258 ), should the default here be "the default of the API" ?

packages/cli/src/commands/org/config/push.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  private async handleValidReport(snapshot: Snapshot) {
+    if (!snapshot.hasChangedResources()) {
+      return;
+    }
+    const targetOrg = await this.getTargetOrg();
+    const canBeApplied = await cli.confirm(
+      `\nWould you like to apply these changes to the org ${bold(
+        targetOrg
+      )}? (yes/no)`
```
olamothe, line 97: Is the `yes/no` needed ? I think cli ux will already do a good job of indicating it's a y/n type of prompt ? 🤔

packages/cli/src/commands/org/config/push.ts
olamothe: Yeah I mean, if it's the same process with some small changes between them, it's a good idea to just extract the common stuff.

packages/cli/src/commands/org/config/push.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  private get configuration() {
+    return new Config(this.config.configDir, this.error);
+  }
+
+  private async getSnapshotPage(snapshot: Snapshot) {
+    const {environment} = await this.configuration.get();
+    const url = platformUrl({environment});
+    const targetOrg = snapshot.targetId;
+    return `${url}/admin/#${targetOrg}/organization/resource-snapshots/${snapshot.id}`;
```
olamothe, line 153: Regarding your comment about code duplication above, for those URL, we could just move them to the same module as `platformUrl()`.

eg:

```
platformSnapshotUrl({environment, orgId, snapshotId})
```

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -78,6 +92,9 @@ export class Snapshot {
   }
 
   public get targetId() {
+    if (this.model.targetId === undefined) {
```
olamothe, line 95: What are the scenario where this ID would be undefined ?

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -39,6 +39,20 @@ export class Snapshot {
     this.displayExpandedPreview();
   }
 
+  public hasChangedResources(): boolean {
+    throw new Error(
```
olamothe, line 43: When the TODO is processed, would that still be an error ? 🤔 

I think it should be still a "success" from the command/cli standpoint, just that nothing happens and we inform the end user.

olamothe (approved):

packages/cli/src/commands/org/config/push.ts
y-lakhdar: Unfortunately, the flag is mandatory when interacting with the platform client
```
export interface DryRunOptions {
    deleteMissingResources: boolean;
}
export interface ApplyOptions {
    deleteMissingResources: boolean;
}
```

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: Multiple parameters from the snapshotModel interface are optional as opposed to what is described by the Migration API. 
I'm checking with the backend team if a pull request on the platform-client is needed.

packages/cli/src/commands/org/config/orgConfigBase.ts
```diff
@@ -0,0 +1,137 @@
...
+import {platformUrl} from '../../../lib/platform/environment';
+import {Snapshot} from '../../../lib/snapshot/snapshot';
+import {red, green} from 'chalk';
+import {normalize} from 'path';
+
+export interface CustomFile extends ReadStream {
+  type?: string;
+}
+
+export default abstract class OrgConfigBase extends Command {
```
y-lakhdar: I created a custom base class so the same code can be reused in the preview and push commands.
Even if I am not a big fan of inheritance, this approach seems to be the most appropriate for this case.

Before that, I tried to simply call the preview methods from the Push command. It worked, but the Oclif team advises against calling [commands programatically](https://oclif.io/docs/running_programmatically).

I also tried moving the common code in a Facade kind of class, but it requires passing a bunch of arguments and flags specific to commands.

Anyhow, let me know if you think that is not the right approach.

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: https://github.com/coveo/platform-client/pull/339

packages/cli/src/lib/platform/environment.ts
```diff
@@ -28,6 +28,20 @@ export function platformUrl<E extends PlatformEnvironment = 'prod'>(options?: {
...
+  return `${url}/admin/#${options.targetOrgId}/organization/resource-snapshots/${options.snapshotId}`;
+}
+
+export function platformSnapshotSynchronizationUrl<
+  E extends PlatformEnvironment = 'prod'
+>(options: {environment: E; targetOrgId: string; snapshotId: string}) {
+  const url = platformSnapshotUrl(options);
+  return `${url}/synchronization`;
+}
+
```
louis-bompart, line 44: I would not put that in `environment.ts` but its in own file

louis-bompart (commented):
lgtm, would move the snapshot URL functions in another file, other than that gg!

packages/cli/src/lib/platform/url.ts
```diff
@@ -0,0 +1,17 @@
+import {PlatformEnvironment, platformUrl} from './environment';
+
+export interface SnapshotUrlOptionsArgs<> {
```
louis-bompart, line 3: You don't need the `<>` I reckon?
```suggestion
export interface SnapshotUrlOptionsArgs {
```

louis-bompart (approved):
LGTM 👍

louis-bompart:
As mentioned in #274, I would have the UT in this PR, even if it's a bit bulky.
The reasoning behind that is it's way easier to review a PR with the code and the test together ;)

PR #262, opened 2021-06-15
feat(ui): support for Headless v1

olamothe:
https://coveord.atlassian.net/browse/CDX-392

## Proposed changes

This is a pre-emptive PR for when v1 of Headless is released.

This week (or the next), we will do one last release of Headless 0.x, which contains all the new functionnality, but with the old function/names/types still present (marked as deprecated, however)

When this is released (eg: headless@0.25), I will be able to merge this PR and do a release of the CLI. 

Right now, the build should fail because all the breaking changes are not released as latest yet (which is what the CLI install by default).

When 1.x is released after that, this will be transparent, as all the necessary upgrade/changes will have already happened in the CLI.

Hopefully that's not too confusing :D

### DIFF
--- a/packages/angular/package-lock.json
+++ b/packages/angular/package-lock.json
@@ -149,31 +149,31 @@
       }
     },
     "@babel/runtime": {
-      "version": "7.14.5",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.14.5.tgz",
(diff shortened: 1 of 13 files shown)


y-lakhdar (approved):
Nice! It's a lot more simpler now

louis-bompart (commented):
Clean!

PR #263, opened 2021-06-15
ci(dependabot): add config

louis-bompart:
## Proposed changes

Added a config to ensure all commits message related to Dependabot includes a Jira number.

-----
CDX-396


y-lakhdar (approved):

olamothe (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
+    commit-message:
+      prefix: 'fix(CDX-227):'
```
olamothe, line 10: I think it should be :

`fix(J:CDX-227)`

for it to be picked up properly by dep pipeline as "disregard this commit + link to JIRA automatically".

Please don't ask my why the `J` is important 😂

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+# Please see the documentation for all configuration options:
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'daily'
```
olamothe, line 8: That's intense !..

louis-bompart: In my mind, it could be every millisecond that I would not care either.
Dependabot open PR when needed (i.e. security derps).
If we have dependabots PR  too often is that we have crappy dependencies and culling is required.

TL;DR: Key thing is that Dependabot updates for security purposes, which is quite rare, so the interval is irrelevant.
A dev in my mind must review code daily, so having it daily makes sense. I only want to have security alerts and PR open  ASAP, just to be 'aware' of them at the very least. Daily is the most frequent, so here's that.

.github/dependabot.yml
louis-bompart: I reckon the Jira should be compliant with what I have currently. I don't think we need dep pipeline to disregard the dependency commits, just to have the Jira tag (the HTTP link I reckon is unnecessary), be reviewed, and merged through a PR.

.github/dependabot.yml
louis-bompart: Found you J, ain't needed anymore https://coveord.atlassian.net/wiki/spaces/CM/pages/1786708612/Package+Rollout+Certifier

.github/dependabot.yml
louis-bompart: Found your J, and not needed according to the doc: https://coveord.atlassian.net/wiki/spaces/CM/pages/1786708612/Package+Rollout+Certifier

PR #264, opened 2021-06-15
fix(ui): use await/async

louis-bompart:
## Proposed changes

Some async functions were not awaited properly, which I think cause some (rare) issues in the tests, and possibly to our users. (i.e. code is not executed and the code after assume it is, oups).

Also, small QoL improvement (imho), I migrated the code from CommonJS modules to ECMA ones, which are I think a bit cleaner (I do prefer the import syntax over the require 😅 )

## Testing

 E2E, although it's a race condition so... Tough to 'reliably' test. It should eliminate the issue listed in the Jira tho.

-----
CDX-398


y-lakhdar (approved):
Oups! thanks for the catch!

PR #266, opened 2021-06-16
reportViewer refactor

y-lakhdar:
The ReportViewer class was messy so I refactored it into more comprehensible classes.


packages/cli/src/lib/snapshot/reportViewer/reportViewerInterfaces.ts
```diff
@@ -0,0 +1,9 @@
+import {ResourceSnapshotsReportOperationModel} from '@coveord/platform-client';
```
louis-bompart, line 1: I'd name the file `reportViewerDataModels`.

It's a bit antinomic to have an exported type in a file named fooInterfaces :D

packages/cli/src/lib/snapshot/reportViewer/reportViewerSection.ts
```diff
@@ -0,0 +1,86 @@
...
+    return this.operationsToDisplay.reduce((previous, current) => {
+      if (this.shouldPrintOperation(current)) {
+        const operationToLog =
+          ReportViewerSection.operationToLogFactoryDictionnary[
+            current
+          ].getString(this.resource.operations[current], indentation);
+        return previous + operationToLog;
+      } else {
+        return previous;
+      }
```
louis-bompart, line 83: I would maybe transfrom the shouldPrint usage in a precondition:
```suggestion
      if (!this.shouldPrintOperation(current)) {
        return previous;
      }
      const operationToLog =
        ReportViewerSection.operationToLogFactoryDictionnary[
          current
        ].getString(this.resource.operations[current], indentation);
      return previous + operationToLog;
```

PR #267, opened 2021-06-17
feat(cli): expanded preview

louis-bompart:
## Proposed changes

When previewing, being it thru the dedicated command or part of the `push` flow allow the user to analyze files to preview what's what.

To be precise:
 - We 'pull' the target organization configuration (i.e. create a new snapshot from org, pull it).
 - We compare each file using the resourceKeys.
   - Resource key present in the target but not in the source snapshot are added
   - Resource key present in the source snapshot but not in the target are kept unless using the `-d` flag.
   - If present in both, then the value of the resource from the source is applied on the target.
   

## Testing

The focus of the tests will be on UT for this, testing is still ongoing:
   - [x] ExpandedPreviewer
   - [x] FileDiffProcessor

-----
CDX-347


packages/cli/src/lib/snapshot/expandedPreviewer.ts
```diff
@@ -0,0 +1,85 @@
+import {
```
y-lakhdar, line 1: in that case, I would rename `reportViewer` to `Previewer`. That way, we will have the `Previewer` and the `ExpandedPreviewer`

packages/cli/tsconfig.json
```diff
@@ -5,10 +5,11 @@
     "declaration": true,
     "importHelpers": true,
     "module": "commonjs",
+    "lib": ["ESNext"],
     "outDir": "lib",
     "rootDir": "src",
     "strict": true,
-    "target": "es2017",
+    "target": "es2019",
```
louis-bompart, line 12: Checked with node.green, '19 is ok for Node 14

packages/cli/src/lib/snapshot/expandedPreviewer.ts
louis-bompart: Naming is really temporary, I'm mostly at the 'throw pasta on the wall and see what stick" stage.

And honestly, I don't think I would merge it till I can test it for real

louis-bompart: So @y-lakhdar, we go with `Previewer`?

olamothe (commented):
@louis-bompart : Are these changes ready to review and merge ?

I saw your last comment about the backend not being ready and you preferring to wait. If that's the case, can you close this temporarily and re-open when it's ready (or switch it to draft).

Thanks !

louis-bompart:
> @louis-bompart : Are these changes ready to review and merge ?

@olamothe, the PR description is up-to-date with the work that I think needs to be done (mainly, UT), but it is ready for review.

> I saw your last comment about the backend not being ready and you preferring to wait. If that's the case, can you close this temporarily and re-open when it's ready (or switch it to draft).

I don't know which comment you are talking about 😕?

olamothe:
> I don't know which comment you are talking about 😕?

This one: 
```
And honestly, I don't think I would merge it till I can test it for real. (the more I coded it, the more I realized that there are a lot of places where it can go haywire, so I don't feel comfortable going forward till the backend is done and that I'm able to see what my code does.
```

But if it's ready, then 👍 Ill do the review.

packages/cli/src/lib/snapshot/expandedPreviewer.ts
```diff
@@ -0,0 +1,221 @@
...
+import dedent from 'ts-dedent';
+import {Dirent} from 'fs';
+
+type ResourcesJSON = Object & {resourceName: string};
+
+type SnapshotFileJSON = Object & {
+  resources: Partial<{[key in ResourceSnapshotType]: ResourcesJSON[]}>;
+};
+
+export class ExpandedPreviewer {
```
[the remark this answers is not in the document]
louis-bompart: Agree. We have several loose ends on what is the 'coveo project' and how to manage it.

I'd say we could leave it as-is with a TODO, wdyt?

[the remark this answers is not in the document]
olamothe: I've created a PR https://github.com/coveo/cli/pull/379 which covers the topic a bit.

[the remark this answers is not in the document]
louis-bompart: Awesome thanks a lot 🙌.

#379 looks simple enough to merge.
Once it is, I will tweak out the code of ExpandedPreviewer to uses it and address this separation of concerns issue.

- [ ] Merge #379
- [ ] Refactor `.coveo/preview` check/creation to uses what's done in #379

packages/cli/src/lib/project/project.ts
```diff
@@ -6,18 +6,18 @@ import {InvalidProjectError} from '../errors';
 import extract from 'extract-zip';
 
 export class Project {
-  public constructor(private pathToProject: string) {}
+  public constructor(private _pathToProject: string) {}
```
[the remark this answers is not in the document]
louis-bompart: The idea with the `public get` is that the property can be read outside the class but not modified.
So, the getter is necessary I think to reflect that.

packages/cli/src/lib/project/project.ts
```diff
@@ -47,17 +46,21 @@ export class Project {
...
       cli.error(error);
     }
   }
 
-  private get pathToTemporaryZip() {
+  public get pathToProject() {
+    return this._pathToProject;
+  }
+
+  public get temporaryZipPath() {
```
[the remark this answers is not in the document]
louis-bompart: publicly readable, not modifiable, so I think not.

packages/cli/src/lib/snapshot/expandedPreviewer.ts
```diff
@@ -0,0 +1,219 @@
...
+
+        const currentFile = readJSONSync(currentFilePath);
+
+        const diffedJSON = this.buildDiffedJson(currentFile, previewFile);
+
+        writeJSONSync(join(currentDir, file.name), diffedJSON, {
+          spaces: 2,
+        });
+      }
+    });
```
y-lakhdar, line 134: What happens if a file exist in `currentDir` but not in `previewDir`?

louis-bompart: For now, nothing.

🔜:tm: 
 - if shouldDelete is true, and the file ain't in previewDir, then we delete the file.
 - if shouldDelete is false, do nothing.

louis-bompart: - [ ] handle deleted files with shouldDelete flag

louis-bompart:
Ready for full review. Incorporate part of #391 and needs it to be merged first
 - [ ] #391 merged

y-lakhdar (changes requested):
image

There is point in the terminal where nothing happens while the expanded preview is being computed. Maybe a spinner or any kind of feedback would let the user know this is normal

y-lakhdar (approved):
GG.
Maybe another to add a spinner while computing the expanded preview

PR #274, opened 2021-06-19
test: add `org:config:push` UT

y-lakhdar:
CDX-402


louis-bompart:
Personal opinion, but I would rather have the test -especially unit tests- and the functionality in the same PR, even if it's a bit bulgier

olamothe (approved):

y-lakhdar:
> Personal opinion, but I would rather have the test -especially unit tests- and the functionality in the same PR, even if it's a bit bulgier

Perfect, I'll adjust next time.

PR #275, opened 2021-06-20
feat: add flag `--skip-preview` on the push command

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-403

## Proposed changes

Add a flag to skip the confirmation prompt. Can be useful when the command is running in CI

## Testing

- [ ] Unit Tests:
    
- [x] Functionnal Tests:
    
- [x] Manual Tests:
    

-----
CDX-403


louis-bompart (approved):
👍

packages/cli/src/commands/org/config/push.ts
```diff
@@ -18,15 +19,25 @@ export default class Push extends SnapshotBase {
 
   public static flags = {
     ...SnapshotBase.flags,
+    skipPreview: flags.boolean({
+      char: 'f',
```
olamothe, line 23: Why the `f` for `skipPreview` ? 🤔

olamothe (approved):

packages/cli/src/commands/org/config/push.ts
y-lakhdar: My bad. I initially named the flag `--force` and forgot to rename the char.

y-lakhdar: Does `-s` makes sens?

PR #277, opened 2021-06-20
feat: remove `--projectPath` flag

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-401

## Proposed changes

`pull`, `preview` and `push` commands should only be executed inside a valid Coveo as Code project.

## Testing

- [x] Unit Tests:
    
- [x] Functionnal Tests:
    
- [x] Manual Tests:
    

-----
CDX-401


louis-bompart (approved):
Nice one! You're hammering the nail on the topic of the usefulness of a base class for the config methods 😄

packages/cli/src/commands/org/config/orgConfigBase.ts
```diff
@@ -47,7 +40,7 @@ export default abstract class SnapshotBase extends Command {
   public static hidden = true;
 
   protected async dryRun() {
-    const project = new Project(normalize(this.flags.projectPath));
+    const project = new Project(normalize(this.projectPath));
```
olamothe, line 43: So, if the command is ran inside a random folder that is not a valid project, what happens ?

What error will the end user see ?

The message is important to explain "why it's not working", and that it's not a cryptic `throw errror.

olamothe (approved):

packages/cli/src/commands/org/config/orgConfigBase.ts
y-lakhdar: Right now, there is a generic error message
image

CDX-354 will provide more information about the error

PR #284, opened 2021-06-22
test: add unit tests for `ui:create:*` commands

y-lakhdar:
Unit tests for `ui:create:*` commands


olamothe (approved):
I would personally have been fine to have no UT for these.

We see that the end result is that you need to mock 90% + of everything that gets executed, so the net result is that the value of the test is ... meh 🤷 . 

It's pretty much just checking that we end up spawning a process with some arguments. Which is borderline internal logic.

The "check on precondition" can be tested in a more general manner by just testing that the generic precondition function works well.

The e2e test that we have right now are what's going to really tell us that "this command works correctly".

PR #300, opened 2021-06-25
fix(release): put back windows signing with fix for github workflow

olamothe:
Bring back windows signing with fixed github workflow.

Was incorrectly injecting `${{{ env.tag }}` instead of `${{ env.tag }}`

https://coveord.atlassian.net/browse/CDX-416

## Testing

- [ ] Unit Tests:
    
- [ ] Functionnal Tests:
    
- [ ] Manual Tests:


louis-bompart:
How did you test it?

olamothe:
> How did you test it?

I pushed it to another branch, and ran the workflow with a manual dispatch.

Previously, the workflow would refuse to start because the YAML was invalid.

louis-bompart:
> > How did you test it?
> 
> I pushed it to another branch, and ran the workflow with a manual dispatch.
> 
> Previously, the workflow would refuse to start because the YAML was invalid.

Can you link the build?

Also, you should **never** do that. You alter the latest release like this.
As a end-user of the repo, I want to be able to trust that the binaries on a tag are made from the commit of the tag, not a random other commit :(

olamothe:
> binaries on a tag are made from the commit of the tag, not a random other commit :(

They were not actually uploaded... So, no release was altered.

louis-bompart (approved):
LGTM 👍 
As discussed earlier, we cannot consider this 'tested' because the tag acquisition process has not actually occurred because the manual dispatch does not carry the tag.

As you suggested, you should do a release process right after merging this PR for testing purposes 👍

PR #302, opened 2021-06-28
refactor(cli): move report logic to SnapshotReporter

y-lakhdar:
## Proposed changes

Encapsulate all operations related to snapshot reports into a single class called `SnapshotReporter`. `ReportViewer` is now only meant to render a report in the terminal... not to interact with in order to get report information (that's the `SnapshotReporter`'s job).

This will decouple the logic from the interface and prevent the `reportViewer` class to grow out of control.

## Testing

- [ ] Unit Tests:
  
- [ ] Functionnal Tests:
  
- [ ] Manual Tests:
  

-----
CDX-412


packages/cli/src/lib/snapshot/reportViewer/reportViewer.ts
```diff
@@ -27,7 +23,7 @@ export class ReportViewer {
   private operationsToDisplay: ReportViewerOperationName[];
 
   public constructor(
-    private readonly report: ResourceSnapshotsReportModel,
+    private reporter: SnapshotReporter,
```
louis-bompart, line 26: I don't think you're doing any re-assignation of this field, so I think you should keep the `readonly` ;) 

```suggestion
    private readonly reporter: SnapshotReporter,
```

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -0,0 +1,66 @@
...
+    return (
+      status === ResourceSnapshotsReportStatus.Completed &&
+      resultCode === ResourceSnapshotsReportResultCode.Success
+    );
+  }
+
+  public getChangedResources(
+    operationsToDisplay: ReportViewerOperationName[]
+  ): ReportViewerResourceReportModel[] {
+    type resourceEntries = [string, ResourceSnapshotsReportOperationModel];
```
louis-bompart, line 44: I would not declare a `type` in a scope (i.e function or class).
This is a hurdle to reusability IMHO.

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -0,0 +1,66 @@
...
+          (previous, current) => previous + operations[current],
+          0
+        ) > 0
+      );
+    };
+
+    const convertArrayToObject = ([name, operations]: resourceEntries) => ({
+      name,
+      operations,
+    });
```
louis-bompart, line 60: I would put those functions as static members of the SnapshotReporter class, it will allow `getChangedResources` to be shorter and thus more readable :)

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -0,0 +1,66 @@
...
+    ]: resourceEntries) => {
+      return (
+        operationsToDisplay.reduce(
+          (previous, current) => previous + operations[current],
+          0
+        ) > 0
+      );
+    };
+
+    const convertArrayToObject = ([name, operations]: resourceEntries) => ({
```
louis-bompart, line 57: Semantic, but I think this a tuple, not an array.

But regardless, because you have a named type for your input and output, I would use them for the function name:
`convertResourceEntriesToResourceSnapshotsReportOperationModel`

I agree that it's a mouthful, but I think it describes a bit better what the function does, given the param types and return type.

louis-bompart (commented):
Good refactor 👍 

I brought up some comments and suggestions, tho, none I would consider an "it must be that way" if there's a good reason not to behind, but could you please address all of them before I approve?

packages/cli/src/lib/snapshot/snapshotReporter.ts
y-lakhdar: I understand. However, this type will only be used in this function and I did not want to make it public outside that scope.
It simply to avoid copying the type twice

y-lakhdar: nvm. I just  saw your comment below :P

louis-bompart: > I understand. However, this type will only be used in this function and I did not want to make it public outside that scope.
> It simply to avoid copying the type twice

If you don't export it, it would only be available in the file. If you need to scope a type more narrowly than a file, that's kinda sus imho.

louis-bompart (approved):

packages/cli/src/commands/org/config/push.spec.ts
```diff
@@ -30,7 +37,44 @@ const mockedApplySnapshot = jest.fn();
 const mockedValidateSnapshot = jest.fn();
 const mockedPreviewSnapshot = jest.fn();
 const mockedLastReport = jest.fn();
-const mockedHasChangedResources = jest.fn();
+
+const getReport = (
```
olamothe, line 41: nit: This is very similar (identical?) to the mock function you are using in preview.spec.ts.

Suggestion: Export a common test utility function in `commands/org/config/mocks.ts` or something like that. 

Same comment for `getSuccessReport` and `getErrorReport`

olamothe (approved):
Good refactor !

PR #303, opened 2021-06-28
feat(cli): do not propose to apply a snapshot with no changes

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-390

## Proposed changes

Do not propose to the user to apply a snapshot if all resources are marked as unchanged
The execution should look like that:
image

Not sure about the message tho... What else can we add to make it more helpful?

## Testing

- [ ] Unit Tests:
  
- [ ] Functionnal Tests:
  
- [ ] Manual Tests:
  

-----
CDX-412


olamothe:
If we inspire ourselves from terraform: 

image

So, something similar would be:

```
No resources to change (in green).

${Target organization} matches the configuration. No changes to apply.
```

The green color makes it apparent that "Everything is good".
It's not an error or a "bad" thing that there's no changes to apply.

olamothe (approved):

PR #305, opened 2021-06-28
feat(cli): provide relevant info if snapshot operation is taking too long

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-411

## Proposed changes
If the operation is taking too much time, the process will stop and display a message to help the user monitor his operation.
image

## Testing

- [ ] Unit Tests:
  
- [ ] Functionnal Tests:
  
- [x] Manual Tests:
By reducing the polling time

-----
CDX-411


packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -135,7 +136,7 @@ export class Snapshot {
         );
 
         if (isNotDone) {
-          throw new Error('Snapshot is still being processed');
+          throw new Error(this.operationGettingTooMuchTimeMessage());
```
olamothe, line 139: Is this an error though ? 🤔 

 My first reaction when I saw the message/screenshot, is that it's not bugged, it's "normal" that sometimes it takes a long time. 

So more like an "info" ?

olamothe (approved):

packages/cli/src/lib/snapshot/snapshot.ts
olamothe: Related to that, what would be missing is something that explains: 

"After you run monitor command and it finally finishes, the next step would be to run XYZ".

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -135,7 +137,7 @@ export class Snapshot {
         );
 
         if (isNotDone) {
-          throw new Error('Snapshot is still being processed');
+          throw new Error(SnapshotErrors.operationTimedOut);
```
louis-bompart, line 140: I think exponentialBackoff here does not fit properly our needs compared to async-retry, I'll get you a suggestion built up 🔜:tm:

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: I guess this help message would only be shown within the context of a preview command. In this case, the next step would be to run `org:config:push`.

However, what would be the next step of a `org:config:push`?

louis-bompart (approved):

Issue #290, opened 2021-06-23
`coveo update` doesn't work when installed with `npm`


louis-bompart:
### Discussed in https://github.com/coveo/cli/discussions/285

<div type='discussions-op-text'>

<sup>Originally posted by **akaustav** June 23, 2021</sup>
Hi,
A few days ago, I had installed `@coveo/cli` on my work laptop (running Windows 10) using the `npm install -g @coveo/cli` command (I had installed `node-gyp` before that). During that installation, v1.3.0 of Coveo CLI was installed on my machine. I noticed that a [`v1.5.0`](https://github.com/coveo/cli/releases/tag/v1.5.0) was released yesterday - **June 22, 2021**. So, I tried to follow instructions documented in the [Usage section](https://github.com/coveo/cli/tree/v1.5.0#usage) to upgrade Coveo CLI:

> Afterwards, anytime you run `coveo update`, your CLI installation will also update to the latest version.

I ran the `coveo update` command in a command prompt and received a `not updatable` message. Is there a different recommended way to upgrade Coveo CLI for folks who installed the `npm` way?

```text
C:\>coveo version
@coveo/cli/1.3.0 win32-x64 node-v14.17.0

C:\>coveo update
@coveo/cli: Updating CLI... not updatable

C:\>coveo version
@coveo/cli/1.3.0 win32-x64 node-v14.17.0
```

Ameet</div>

## To Reproduce

1. Run `npm i -g @coveo/cli@1.3.0`
2. Check coveo is well installed with an appropriate version with `coveo -v`.
  - Expected output is `@coveo/cli/1.3.0` followed by your platform and node version.
3. Run `coveo update`
5. See the error in the terminal. `@coveo/cli: Updating CLI... not updatable`
6. Check the cover version again (see 2.), the version should still be `1.3.0`

## Expected behavior

The CLI should have updated and not display 'not updatable'

## Screenshots

![image](https://user-images.githubusercontent.com/12366410/123124002-4ad7f200-d415-11eb-8738-58a3f21341f7.png)

## Desktop (please complete the following information):

- OS: Windows
  - OS version: 10.0.19041
- Browser: N.A.
  - Browser version: N.A.
- Version of the CLI 1.3.0
- Local Node version: 14.17.0
- Local NPM version: 6.14.13

## Additional context

https://coveord.atlassian.net/browse/CDX-409

louis-bompart:
[...]

PR #306, opened 2021-06-28
feat(cli): add warning and update instructions for NPM installations

louis-bompart:
## Proposed changes

[oclif/plugin-update[(https://github.com/oclif/plugin-update) does not support NPM based installation.
I copied what SFDX does and added a warning message (literally the same, just modified the package name).

The PR used a bit of black magic by using the scopedVar of oclif (that's what SFDX does) & `UPDATE_INSTRUCTIONS` from plugin-update. You can see where the magic come from here: https://github.com/oclif/plugin-update/blob/42a65ebabb25b1af97337a52b43cc0b4bb2e7331/src/commands/update.ts#L203-L207

Here's the result:
image

## Testing

- ❌  Unit Tests: N.A.
- ❌  Functional Tests: N.A.
- [x] Manual Tests: tested with baguette-cli@1.9.0, see screenshot above.

-----
Fixes:
CDX-409
#290


louis-bompart:
@fbeaudoincoveo 👋 I added you just to correct my potentially horrendous English 😂

README.md
```diff
@@ -32,7 +32,10 @@ If you have [installed node-gyp](https://github.com/nodejs/node-gyp#installation
 npm install -g @coveo/cli
 ```
 
-Afterwards, anytime you run `coveo update`, your CLI installation will also update to the latest version.
```
fbeaudoincoveo, line 35: We already say:

> These executables install the latest available version. Afterwards, anytime you run `coveo update`, your CLI installation will update to the latest version.

on line 27.

I'd rephrase the above slightly:

> Running one of the above executables will install the latest available version of the CLI.
>
> Afterward, you can run `coveo update` at any time to update your CLI installation to the latest version.

And here (i.e., under the npm installation instructions), I'd simply write:

> Afterward, you can run `npm update -g @coveo/cli` at any time to update your npm-based CLI installation to the latest version.

packages/cli/src/hooks/init/set-global-config.ts
```diff
@@ -2,6 +2,8 @@ import {Hook} from '@oclif/config';
 
 const hook: Hook<'init'> = async function (opts) {
   global.config = opts.config;
+  process.env[this.config.scopedEnvVarKey('UPDATE_INSTRUCTIONS')] =
+    'Use "npm update --global @coveo/cli" to update npm-based installations.';
```
fbeaudoincoveo, line 6:
```suggestion
    'Use "npm update --global @coveo/cli" to update your npm-based CLI installation.';
```

fbeaudoincoveo (approved):

README.md
louis-bompart: Good idea!

louis-bompart: > We already say:
> 
> > These executables install the latest available version. Afterwards, anytime you run `coveo update`, your CLI installation will update to the latest version.
> 
> on line 27.
> 
> I'd rephrase the above slightly:
> 
> > Running one of the above executables will install the latest available version of the CLI.
> > Afterward, you can run `coveo update` at any time to update your CLI installation to the latest version.
> 
> And here (i.e., under the npm installation instructions), I'd simply write:
> 
> > Afterward, you can run `npm update -g @coveo/cli` at any time to update your npm-based CLI installation to the latest version.

@fbeaudoincoveo, following up on that, I restructured the readme a bit, wdyt?

README.md
```diff
@@ -6,42 +6,48 @@
 
 Coveo CLI is a command line interface to interact with the Coveo platform. It allows you to quickly create a [Coveo Headless](https://docs.coveo.com/headless)-powered search page for your Coveo organization, relying on [Angular](https://angular.io), [React](https://reactjs.org/) or [Vue.js](https://vuejs.org/).
 
-## Usage
+## Installation
```
fbeaudoincoveo, line 9: Putting "NPM/NPX" at the same level as "Mac", "Windows", and "Linux" is slightly confusing IMO, because it makes it seem like npm/npx is another OS.

I'd present the information as follows:

## Installation

### Option 1: Install via an Executable

Download and run the executable for your operating system to install the latest available version of the CLI.

Afterward, you can run `coveo update` at any time to update your CLI installation to the latest version.

- Mac
  ...

- Windows
  ...

- Linux
  ...

### Option 2: Install via npm

If you have [installed node-gyp](https://github.com/nodejs/node-gyp#installation), you can alternatively install the CLI globally via [npm](https://www.npmjs.com/package/@coveo/cli):

```sh
npm install -g @coveo/cli
```

Afterward, you can run `npm update -g @coveo/cli` at any time to update your npm-based CLI installation to the latest version.

fbeaudoincoveo (approved):

olamothe (approved):
Good find with the `UPDATE_INSTRUCTIONS` variable.

Have you however been able to test the "opposite" ? 

That is, if you have **not** installed coveo through NPM, is it still going to work correctly when you try to update ?

louis-bompart:
> Good find with the `UPDATE_INSTRUCTIONS` variable.
Thanks 😄  
> Have you however been able to test the "opposite"?
> 
> That is if you have **not** installed Coveo through NPM, is it still going to work correctly when you try to update?

No, I could not: testing the Oclif update process this way requires another S3 setup, etc.
However, the code that I implemented is almost a 1-to-1 copy-paste of SFDX-CLI, and I tested that their update process was working.

I concede that there's a risk that ours doesn't because I was not able to test it, but I do think that it's low enough to go forward.

However (pretty much as always), we should test it as-soon-as released (well, actually, as soon as we have a release after this changeset has been released)

PR #310, opened 2021-06-29
test(e2e): use axios for search token server assertions

louis-bompart:
## Proposed changes

Instead of opening the page in Chrome, do a HTTP GET on the server, check that the response ain't rotten.

## Testing

Changeset solely affect tests.

-----
CDX-426


y-lakhdar (approved):

packages/cli-e2e/__tests__/angular.specs.ts
```diff
@@ -459,9 +460,8 @@ describe('ui:create:angular', () => {
     });
 
     it('should run the server on a new port', async () => {
-      await expect(
-        page.goto(tokenServerEndpoint(), {waitUntil: 'load'})
-      ).resolves.not.toThrow();
+      const tokenRequest = await axios.get(tokenServerEndpoint());
+      expect(tokenRequest.data.token).toBeDefined;
```
olamothe, line 464: Could go a little bit further to check that it's a string that looks like a JWT token:

```suggestion
      expect(tokenRequest.data.token).toMatch(/^eyJh/)
```

olamothe (approved):
Good idea ! 💯

PR #317, opened 2021-06-30
test(e2e): sanitize github actions variable for angular

louis-bompart:
## Proposed changes

Angular wants URL-friendly text for the project name. Somehow GitHub Action switched from `run-#` to `__run_#`. Which 💥.

I implemented three regexes to counter-act this and sanitize the variable:
 - first one remove all leading underscore from the variable name.
 - second to replace all other underscores with a dash
 - third to remove the dash if the first character of the right side of the dash is a digit (angular ain't liking it either)

## Testing

- [x] Unit Tests: n.a.
- [x] Functional Tests: Covered. If the bug was here, no angular spec would go green.
- [x] Manual Tests: Regex101 to test the regex.

-----
CDX-431

### DIFF
--- a/packages/cli-e2e/setup.ts
+++ b/packages/cli-e2e/setup.ts
@@ -23,8 +23,11 @@ async function clearChromeBrowsingData(browser: Browser) {
 
 export default async function () {
   mkdirSync(SCREENSHOTS_PATH, {recursive: true});
-  process.env.GITHUB_ACTION =
-    process.env.GITHUB_ACTION || randomBytes(16).toString('hex');
+  process.env.GITHUB_ACTION = (
+    process.env.GITHUB_ACTION || randomBytes(16).toString('hex')
+  )
+    .replace(/^_*/g, '') // Remove all `-` prefix
+    .replace(/_/g, '-'); // Replace alls other '_' by a '-'
   const browser = await connectToChromeBrowser();
   await clearChromeBrowsingData(browser);
   await clearAccessTokenFromConfig();
--- a/packages/cli-e2e/setup.ts
+++ b/packages/cli-e2e/setup.ts
@@ -23,12 +23,15 @@ async function clearChromeBrowsingData(browser: Browser) {
 
 export default async function () {
   mkdirSync(SCREENSHOTS_PATH, {recursive: true});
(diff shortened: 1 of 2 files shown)


packages/cli-e2e/setup.ts
```diff
@@ -23,8 +23,12 @@ async function clearChromeBrowsingData(browser: Browser) {
 
 export default async function () {
   mkdirSync(SCREENSHOTS_PATH, {recursive: true});
-  process.env.GITHUB_ACTION =
-    process.env.GITHUB_ACTION || randomBytes(16).toString('hex');
+  process.env.GITHUB_ACTION = (
+    process.env.GITHUB_ACTION || randomBytes(16).toString('hex')
+  )
+    .replace(/^_*/g, '') // Remove all `-` prefix
```
olamothe, line 29: Instead of those comments, you can create the regex with a meaningful name.

```
const leadingDashPrefix = /^_*/g;
const allUnderscore = /_/g;
const allDashFollowedByADigit = /-*(?=\d)/g;

process.env.GITHUB_ACTION = (
    process.env.GITHUB_ACTION || randomBytes(16).toString('hex')
  )
  .replace(leadingDashPrefix, '')
  .replace(allUnderscore, '-')
  .replace(allDashFollowedByADigit, '')
```

And I would instead add a comment that explains why that's important/meaningful ( `Angular needs url friendly.... etc` )


Related to that, if angular really needs URL friendly text, can't we just do  
```
process.env.GITHUB_ACTION = encodeURIComponent(process.env.GITHUB_ACTION || randomBytes(16).toString('hex'))
```

louis-bompart: (not on PC anymore till monday, so as mentioned in Slack feel free to bring changes if need be)

But:
Sure for the var name, that actually is a good idea 👍 

For the encoding, no 'cause 
image
And also because `%20` and stuff like that is not really readable and can hurt the understanding of the logs imho.

y-lakhdar: Looks like the the replace is not doing what the description is saying
```
replace(/^_*/g, '') // Remove all `-` prefix
```
The regex replaces underscores... not dashes

y-lakhdar (approved):

PR #319, opened 2021-07-02
feat(cli): monitor command

y-lakhdar:
## Proposed changes

A command to monitor the operation of a snapshot. 

To achieve that, I had to do a few changes to existing classes:
* `SnapshotFactory` has a method to create a Snapshot object from a valid snapshotId.
* `Snapshot.waitUntilDone` method can either wait for a specific snapshot operation (if specified in the options) or any ongoing operations (if none specified).

## Testing

- [x] Unit Tests:
  
- [ ] Functionnal Tests:
  
- [x] Manual Tests:
  Created a number large snapshot that I monitor with the new `org:config:monitor` command

-----
CDX-423


packages/cli/src/commands/org/config/orgConfigBase.ts
```diff
@@ -23,7 +24,10 @@ export interface CustomFile extends ReadStream {
 export default abstract class SnapshotBase extends Command {
   public static description = 'Create and validate a snapshot';
 
-  public static flags = {
+  public static flags: {
+    target: IOptionFlag<string | undefined>;
+    deleteMissingResources?: IFlag<boolean>;
+  } = {
```
y-lakhdar, line 30: Making `deleteMissingResource` flag optional, so children classes (such a Monitor) can benefit from the base class without extending all of its flags.

It's kinda hackish. However, another solution would be to simply not inherit from the `SnapshotBase` class and simply rewrite the required methods directly in the `Monitor` command.

packages/cli/README.md
```diff
@@ -122,6 +123,22 @@ OPTIONS
 
 _See code: [@oclif/plugin-help](https://github.com/oclif/plugin-help/blob/v3.2.2/src/commands/help.ts)_
 
+## `coveo org:config:monitor [FILE]`
```
olamothe, line 126: I think this should be regenerated

packages/cli/src/commands/org/config/orgConfigBase.ts
olamothe: Yeah, that's the general problem with inheritance: you have to "nail" the abstraction in a very strict manner in order to not leak weird stuff into all children classes. 

And it's very hard to do it right on the first try and to never have to change it.

I would bite the bullet and refactor. Those kind of "add a hack in the parent class so that one of the child class is behave differently" is a sure sign that there's a flaw in the design, and they are prone to creating hard to debug problems.

If possible to use composition over inheritance, and to split the different "features" into separate class, and that you can then mix and match those classes in different command would make this much more robust.

For example, have different modules for the flag, another one for `dryRun`, `createSnapshotFromProject`, `validateSnapshot` etc.

And those different modules don't have to inherit from command: just simple functions/class.

packages/cli/README.md
louis-bompart: Mh, I was think about that: Should we document unreleased commands?

I would let the CI update the doc tbh. Like this we avoid playing ball with l23 and also ensure that the doc is updated whenever we release the command.

louis-bompart (commented):
As @olamothe said, the 'BaseClass' has made its time it seems.
Repeating code or hacking it is the lesser option and we'll prolly regret it later, so let's save future us from ourselves now.

olamothe (approved):
Assuming https://github.com/coveo/cli/pull/326 is getting merged, this is 👍 for me !

PR #321, opened 2021-07-05
chore(J:CDX-227) update husky to 7

louis-bompart:
Same as https://github.com/coveo/cli/pull/313, but with a 7. It allows to ditch the gitignore and whatnot and keep it clean.


y-lakhdar (approved):
works as expected

PR #326, opened 2021-07-07
chore: remove base class

y-lakhdar:
## Proposed changes

Remove abstract command and use modules instead.

### Other changes
* Removed `CustomFile` interface from push command since it was not used
* Create a SnapshotUrlBuilder class to generate urls to snapshot pages

-----
CDX-441


packages/cli/src/commands/org/config/monitor.ts
```diff
@@ -100,4 +100,17 @@ export default class Monitor extends SnapshotBase {
       },
     };
   }
+
+  private async getTargetOrg() {
```
olamothe, line 104: This is repeated across many command and could be extracted/shared.

The only difference seems to be the command itself (eg: `Monitor` + a reference to the command instance `this`), which could be passed as arguments to a shared function.

olamothe (approved):
💯 Good job ! It's a bit more code than before, but overall should be easier to maintain than inheritance.

packages/cli/src/commands/org/config/push.ts
```diff
@@ -1,25 +1,39 @@
...
-}
-
-export default class Push extends SnapshotBase {
+export default class Push extends Command {
   public static description =
     'Preview, validate and deploy your changes to the destination org';
 
   public static flags = {
-    ...SnapshotBase.flags,
+    target: flags.string({
```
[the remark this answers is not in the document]
y-lakhdar: I though about it but the description differs between commands
```
The unique identifier of the organization where to send the changes
```
VS
```
The unique identifier of the organization to preview
```

packages/cli/src/commands/org/config/monitor.ts
y-lakhdar: Honestly, I feel like a minimal base class just to return the target org would be a good idea for this case.
what do you think?

packages/cli/src/lib/snapshot/snapshotUrlBuilder.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('SnapshotUrlBuilder', () => {
+  let snapshotUrlBuilder: SnapshotUrlBuilder;
+  let snapshot: Snapshot;
+
+  beforeAll(async () => {
+    snapshot = await createSnapshot();
+  });
+
+  it('#createSnapshotPage should return the snapshot URL', () => {
+    snapshotUrlBuilder = new SnapshotUrlBuilder(getUSProdConfig());
```
louis-bompart, line 38: I think you could put that line in a beforeEach?

packages/cli/src/commands/org/config/push.ts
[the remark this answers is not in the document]
louis-bompart: It might be better to lose in specificity in the description of the flag names so to show that "it's the same thing"

e.g. if you check `-v` in SFDX-CLI, it's always(-ish) used to describe the dev hub organization:
https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_force_package.htm

However, in the case of more-than-one org commands, it is true that'd want to have a more specific command, 'cause "The unique identifier of the organization" ain't gonna cut it.

louis-bompart (commented):
lgtm overall, might be possible to refactor the flags a bit more (or maybe not!).

packages/cli/src/commands/org/config/monitor.ts
louis-bompart: > Honestly, I feel like a minimal base class just to return the target org would be a good idea for this case.
> what do you think?
> It will only be used by commands that actually connect to a target organization...
> 
> Or it can also be a util function in SnapshotCommon

I think it's better to stay in one paradigm here, the idea of Oli to have a pure function that given a specific Command, will return you the organization.

However, you could introduce specificity in the command that are allowed to call it.

i.e.
```ts
class Monitor extends Commands implements IWithTargetOrg
```

PR #328, opened 2021-07-08
feat(cli): create class for the `pull` command

y-lakhdar:
## Proposed changes

The Pull command placeholder as well as a method to create a snapshot from the organization

## Testing

- [x] Unit Tests:
  
- [ ] Functionnal Tests:
  
- [x] Manual Tests: created a snapshot from a trial organization by hard coding the resources to export (since the `--resourcesToExport` flag is not yet implemented)

-----
CDX-445


packages/cli/src/commands/org/config/pull.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  private async refreshProject(_snapshot: Snapshot) {
+    // TODO: CDX-446: refresh project
+  }
+
+  private async getSnapshot() {
+    const target = await this.getTargetOrg();
+
+    const resourceTypesToExport = this.resourceTypesToExport;
```
louis-bompart, line 42: I think this assignation is a bit superfluous

packages/cli/src/lib/snapshot/snapshotFactory.ts
```diff
@@ -50,8 +51,31 @@ export class SnapshotFactory {
...
+    targetOrg: string
+  ) {
+    const client = await this.getClient(targetOrg);
+    const resourcesToExport = resourceTypesToExport.reduce(
+      (resourceToExport, currentType) => {
+        resourceToExport[currentType] = ['*'];
+        return resourceToExport;
+      },
+      {} as Partial<Record<ResourceType, string[]>>
+    );
```
louis-bompart, line 65: in fine, for this one, I think a for each would be more readable
```suggestion
    let resourcesToExport: Partial<Record<ResourceType, string[]>> = {};
    resourceTypesToExport.forEach(
      (currentType) => {
        resourcesToExport[currentType] = ['*'];
      }
    );
```

louis-bompart (approved):
splitting hair really :P

packages/cli/src/commands/org/config/pull.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  public static flags = {
+    target: flags.string({
+      char: 't',
+      description:
+        'The unique identifier of the organization from which to pull the resources. If not specified, the organization you are connected to will be used.',
+      helpValue: 'destinationorganizationg7dg3gd',
+      required: false,
+    }),
+    // TODO: CDX-447: add flag to select resource types to export
```
olamothe, line 22: Question: How will we keep that list of ressources up to date ? If we hardcode them in the CLI, then it means whenever a new type of resource is added in the platform, we will be lagging behind.

Which is annoying for end users, and for us.

Any way to do this in an intelligent manner ? 🤔

packages/cli/src/commands/org/config/pull.ts
```diff
@@ -0,0 +1,62 @@
...
+  @Preconditions(IsAuthenticated())
+  public async run() {
+    const snapshot = await this.getSnapshot();
+
+    await this.refreshProject(snapshot);
+    await snapshot.delete();
+  }
+
+  private async refreshProject(_snapshot: Snapshot) {
+    // TODO: CDX-446: refresh project
```
olamothe, line 36: What is "refreshing a project" ?

olamothe (approved):

packages/cli/src/commands/org/config/pull.ts
y-lakhdar: mmh. It's probably not the right word.
I meant refresh/update the project directory.

packages/cli/src/commands/org/config/pull.ts
y-lakhdar: Don't worry. It will not be harcoded. 
All the available resource types will be directly coming from platform-client.

packages/cli/src/commands/org/config/pull.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  private get configuration() {
+    return new Config(this.config.configDir, this.error);
+  }
+
+  private get resourceTypesToExport() {
+    // TODO: CDX-447: pass resource types to export
+    return [ResourceType.field, ResourceType.extension];
+  }
```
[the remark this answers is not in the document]
y-lakhdar: Yes it will be be dynamic. I just hard coded these 2 resource types in the meantime... but I think I should have just thrown an error to prevent the confusion.
#341 contains the code that populates the resources to export

PR #331, opened 2021-07-09
feat(cli): refresh project with snapshot content

y-lakhdar:
## Proposed changes

* Added a method to download the snapshot ZIP and refresh the project with the snapshot content.
* Updated the `.tsconfig` file to support [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) and [File](https://developer.mozilla.org/en-US/docs/Web/API/File) types since the platform-client was mainly written to be used by web apps rather than Node.

This PR depends on 2 Platform-client PRs:
### Add method to export the snapshot as a ZIP file ([#347](https://github.com/coveo/platform-client/pull/347))
Added a method to support the new `/snapshots/{snapshotId}/content` request ([SRC-4434](https://coveord.atlassian.net/browse/SRC-4434)).

### Overload #createFromFile to support file buffer ([#349](https://github.com/coveo/platform-client/pull/349))
This PR is necessary for 2 reasons:
1. Avoid the [`CustomFile`](https://github.com/coveo/cli/blob/master/packages/cli/src/lib/snapshot/snapshotFactory.ts#L11) hack to simulate a File object.
2. Allow the `ResourceSnapshots#createFromFile` method to be used in NodeJs without the need of a polyfill for the [File API](https://developer.mozilla.org/en-US/docs/Web/API/File).

## Testing

- [x] Unit Tests:
  
- [ ] Functionnal Tests:
  
- [x] Manual Tests:
  

## Additional acceptance criteria
- [x] Merge https://github.com/coveo/platform-client/pull/349
- [x] Merge https://github.com/coveo/platform-client/pull/347
- [x] Merge https://github.com/coveo/platform-client/pull/351
- [x] Upgrade platform-client dependency and remove platform-client patch

-----
CDX-446


packages/cli/src/lib/project/projectError.ts
```diff
@@ -0,0 +1,6 @@
+export class InvalidProjectError extends Error {
+  public static readonly message = 'INVALID_PROJECT_ERROR';
+  public constructor() {
+    super(InvalidProjectError.message);
+  }
+}
```
louis-bompart, line 6: I reckon we have several errors now?
Maybe create a folder for them?

louis-bompart (approved):
pretty clean 👌

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -68,6 +69,12 @@ export class Snapshot {
     await this.client.resourceSnapshot.delete(this.model.id);
   }
 
+  public async download() {
```
olamothe, line 72: `async` is not needed here since there's no `await` keyword in the function body

packages/cli/src/lib/snapshot/snapshotFactory.spec.ts
```diff
@@ -82,17 +74,16 @@ describe('SnapshotFactory', () => {
 
     it('#createSnapshotFromZip should create a snapshot from Zip with appropriate parameters', () => {
       expect(mockedCreateSnapshotFromFile).toHaveBeenCalledWith(
-        expect.objectContaining({
-          type: 'application/zip',
-        }),
-        {developerNotes: 'cli-created-from-zip'}
+        Buffer.from('hello there'),
```
olamothe, line 77: General Kenobi !

packages/cli/tsconfig.json
```diff
@@ -12,7 +12,8 @@
     "noEmitOnError": false,
     "skipLibCheck": true,
     "types": ["jest"],
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "lib": ["WebWorker"]
```
olamothe, line 16: Why is this needed ? For which functionality ?

olamothe (approved):

packages/cli/tsconfig.json
y-lakhdar: Because platform-client is using APIs not available in NodeJs ([Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob), https://developer.mozilla.org/en-US/docs/Web/API/File).
We only need the typing and not the actual implementation.

y-lakhdar: Thanks to @louis-bompart for this one :P

PR #335, opened 2021-07-12
feat(push): add push source list command

olamothe:
## Proposed changes

Add a command that list the available push source in your organization. Will be needed for upcoming command that allows to push local files from disk, since you need to specify which source to use.

## Testing

- [X] Unit Tests:
- [ ] Functionnal Tests:
- [X] Manual Tests:

-----
https://coveord.atlassian.net/browse/CDX-453

### DIFF
--- a/packages/cli/src/commands/source/push/list.ts
+++ b/packages/cli/src/commands/source/push/list.ts
@@ -1,0 +1,65 @@
+import {FilterHostType, SourceModel} from '@coveord/platform-client';
+import {cli} from 'cli-ux';
+import {Command} from '@oclif/command';
+import {
+  IsAuthenticated,
+  Preconditions,
+} from '../../../lib/decorators/preconditions';
+import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
+import dedent from 'ts-dedent';
+
+export default class SourcePushList extends Command {
+  public static description =
+    'List all available push sources in your Coveo organization';
+
+  public static flags = {
+    ...cli.table.flags(),
+  };
+
+  @Preconditions(IsAuthenticated())
+  public async run() {
+    const {flags} = this.parse(SourcePushList);
+    const authenticatedClient = new AuthenticatedClient();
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/source/push/list.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  public static flags = {
+    ...cli.table.flags(),
+  };
+
+  @Preconditions(IsAuthenticated())
+  public async run() {
+    const {flags} = this.parse(SourcePushList);
+    const authenticatedClient = new AuthenticatedClient();
+    const org = (await authenticatedClient.cfg.get()).organization;
```
y-lakhdar, line 27: Is it a styling preference that the `org` variable is declared at the beginning even if it is only used in the `if` block?
Really not important

y-lakhdar (approved):

packages/cli/src/commands/source/push/list.ts
olamothe: Yeah, styling preference I guess :D

packages/cli/src/commands/source/push/list.ts
```diff
@@ -0,0 +1,80 @@
...
+      buildAnalyticsFailureHook(this, flags, err)
+    );
+    throw err;
+  }
+
+  private flattenSourceModels(sourceModels: SourceModel[]) {
+    return sourceModels.map((sourceModel) => ({
+      ...sourceModel,
+      status: sourceModel.information?.sourceStatus?.type,
+      numberOfDocuments: sourceModel.information?.numberOfDocuments,
```
louis-bompart, line 77: 🤔 how will `cli.table` digest those values if one of them is undefined or null

louis-bompart (approved):
lgtm

packages/cli/src/commands/source/push/list.ts
olamothe: You will get an empty row in the table.

PR #338, opened 2021-07-12
test(e2e): fix invalid path

louis-bompart:
## Proposed changes

This prevented E2E to run locally.

## Testing

- [x] Unit Tests: n.a.
- [x] Functionnal Tests: n.a.
- [x] Manual Tests: ran the test. could not before.

-----
CDX-458


y-lakhdar:
I don't understand how come the Github actions were passing before and after that change. 🤔

packages/cli-e2e/entrypoints/utils/common.js
```diff
@@ -5,7 +5,7 @@ const {existsSync, mkdirSync, writeFileSync} = require('fs');
 const DOCKER_IMAGE_NAME = 'coveo-cli-e2e-image';
 const composeProjectName = 'coveo-cli-e2e';
 const DOCKER_CONTAINER_NAME = 'coveo-cli-e2e-container';
-const repoHostPath = resolve(__dirname, ...new Array(3).fill('..'));
+const repoHostPath = resolve(__dirname, ...new Array(4).fill('..'));
```
olamothe, line 8: What does that 3->4 represent ? What does it reference ?

And second question, can we retrieve it "correctly" in another manner so that it does not happen ?

louis-bompart: > What does that 3->4 represent ? What does it reference ?
the `Array(n)` represent the number of `..` to add to the path resolving (i.e. how many time do we have to go one folder higher in the hierarchy). This is more readable IMO than `'..','..'` four times.

> And second question, can we retrieve it "correctly" in another manner so that it does not happen ?
No, we cannot afaik. It's pathing, not an import of any kind. If the files moves, we have to update the path. There's no 'automagic' correction possible being at the IDE level or anything afaik.

louis-bompart:
> I don't understand how come the Github actions were passing before and after that change. 🤔

Because they have their own 'circuits' since we implemented test run for Windows, we run 'on metal', no docker involved. More precisely, the entry.js is not used by the CI.

It's only their for us mere mortals :P

olamothe (approved):

PR #340, opened 2021-07-12
feat(angular-12): update all deps, tweak port-allocator

louis-bompart:
## Proposed changes

Updated all Angular dependencies up to the latest.
The field `options` is not here by default now in projectName.architect.serve🤷.
So I null-checked it and add it if missing.

## Testing

- [x] Unit Tests: covered
- [x] Functional Tests: covered
- [x] Manual Tests: Verdaccio publication of the template + test with the CLI.

-----
CDX-459


packages/angular/src/search-token-server/files/scripts/port-allocator.mjs
```diff
@@ -80,13 +80,16 @@ const isPortAvailable = async (port, options) => {
 };
 
 function updateAppPort(port) {
-  const angularJsonPath = resolve('angular.json');
+  const angularJsonPath = resolve("angular.json");
```
y-lakhdar, line 83: I think the build is failing because of the double quotes

olamothe (approved):

PR #341, opened 2021-07-13
feat(cli): add `--resourceTypes` flag

y-lakhdar:
## Proposed changes

Added a flag to specify resources to pull from organization (see image).
The resource types are coming directly from the platform-client and are shown in the help message.
image

## Testing

- [x] Unit Tests:
  
- [ ] Functionnal Tests:
  
- [x] Manual Tests:
  

-----
CDX-447


packages/cli/src/commands/org/config/pull.ts
```diff
@@ -17,20 +18,26 @@ export default class Pull extends Command {
...
       char: 't',
-      description:
-        'The unique identifier of the organization from which to pull the resources. If not specified, the organization you are connected to will be used.',
       helpValue: 'destinationorganizationg7dg3gd',
       required: false,
+      description:
+        'The unique identifier of the organization from which to pull the resources. If not specified, the organization you are connected to will be used.',
     }),
-    // TODO: CDX-447: add flag to select resource types to export
+    resourceTypes: flags.string({
```
olamothe, line 26: Is `ResourceType` a string enum in platform client ? 

If so, I think using `flags.enum` would be more appropriate, and would remove the need to cast on line 33. 

You also wouldn't need to do any mapping on line 62-63, since you would already have the correct value I think.

olamothe (approved):

packages/cli/src/commands/org/config/pull.ts
y-lakhdar: I tried with enum but looks like `flags.enum` only accept one value

olamothe: Like this here ? https://github.com/coveo/cli/blob/master/packages/cli/src/commands/source/push/new.ts#L19

olamothe: Ohh I understand, you can't repeat.

NVM me then.

packages/cli/src/commands/org/config/pull.ts
```diff
@@ -17,20 +18,26 @@ export default class Pull extends Command {
...
+      char: 'r',
+      helpValue: 'type1 type2',
```
louis-bompart, line 28: Did you test the space separator with an actual run (manual test)? I think this might cause issue (gut feeling)

louis-bompart (approved):
lgtm, just confirm that space aint causing issue

packages/cli/src/commands/org/config/pull.ts
y-lakhdar: Yes,
I just pulled my entire test org with the command

```
org:config:pull -r extension featuredResult field filter mapping mlModel mlModelAssociation queryParameter queryPipeline queryPipelineCondition rankingExpression  rankingWeight searchPage securityProvider source stopWord subscription thesaurus trigger
```

PR #342, opened 2021-07-13
doc(cli) Generate valid MD links & improve docs

fbeaudoincoveo:
## Proposed changes

Changed some descriptions, log messages, and error messages.

-----
DOC-9433


packages/cli/src/commands/org/search/dump.ts
```diff
@@ -36,49 +36,49 @@ interface FetchParameters {
...
       char: 'f',
       description:
-        'Additional search filter that should be applied while doing the extraction. See https://docs.coveo.com/en/1552 for more information',
+        'The additional search filter to apply while getting the content. See <https://docs.coveo.com/en/1552>.',
       default: '',
     }),
     chunkSize: flags.integer({
       char: 'c',
       description:
-        'The maximum number of results extract into each CSV file. Default is 10000',
```
fbeaudoincoveo, line 81: The current description seems to imply that more than one CSV file could be created if more results are returned than the specified chunkSize. However, other descriptions seemed to imply that only one CSV file would be created for the data dump.

Can more than one file be created? If so, I'll change the documentation a bit.

olamothe (approved):

packages/cli/src/commands/org/search/dump.ts
olamothe: Yes, it will create one file per chunk.

olamothe:
@fbeaudoincoveo If you can run `npm run prettier:fix`, this should fix some build problems. TY !

fbeaudoincoveo:
> @fbeaudoincoveo If you can run `npm run prettier:fix`, this should fix some build problems. TY !

Will do!

y-lakhdar (approved):

packages/cli/src/commands/ui/create/vue.ts
```diff
@@ -182,7 +186,7 @@ export default class Vue extends Command {
     npm run start
 
     See package.json for other available commands.
-    Happy hacking !
+    Happy hacking!
```
louis-bompart, line 189: This one is a bit special, we use it in our E2E test to detect when the command completes.

You can change it, no prob, but make sure to replace the regexp in `packages\cli-e2e\__tests__\vue.specs.ts` L92

packages/cli/src/commands/ui/create/angular.ts
```diff
@@ -136,7 +140,7 @@ export default class Angular extends Command {
     npm run start
 
     See package.json for other available commands.
-    Happy hacking !
+    Happy hacking!
```
louis-bompart, line 143: Check `packages\cli-e2e\__tests__\angular.specs.ts` L125 to update the regexp ;)

packages/cli/src/commands/auth/login.ts
```diff
@@ -30,18 +30,18 @@ export default class Login extends Command {
       ],
       default: 'us-east-1',
       description:
-        'The platform region to log into. See https://docs.coveo.com/en/2976.',
+        'The Coveo Platform region to log in to. See <https://docs.coveo.com/en/2976>.',
```
louis-bompart, line 33: 🤔 what's the point of the angle brackets?

louis-bompart (changes requested):
The RegExp from the E2E needs to be updated, other than that LGTM 👍

packages/cli/src/commands/auth/login.ts
fbeaudoincoveo: If I want to output these as hyperlinks on the docs site, our Markdown parser/converter will recognize `<https://example.com>` syntax as a link that doesn't have a label and convert it to `<a href="https://example.com"></a>`

Otherwise, if I just leave the link as is, it's going to be displayed as plain text on the docs site.

louis-bompart (approved):

PR #343, opened 2021-07-13
feat(push): add delete document push source command

olamothe:
https://coveord.atlassian.net/browse/CDX-461

## Proposed changes

Add command that allows to delete document in a push source, using either a list of document URI, or olderThan pattern.

## Testing

- [X] Unit Tests:
  
- [ ] Functionnal Tests:
  
- [X] Manual Tests:

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -7,6 +7,7 @@
   },
   "dependencies": {
     "@angular/cli": "^11.1.4",
+    "@coveo/push-api-client": "^1.0.0",
     "@coveord/platform-client": "^21.4.0",
     "@oclif/command": "^1",
     "@oclif/config": "^1",
--- a/packages/cli/src/commands/source/push/delete.spec.ts
+++ b/packages/cli/src/commands/source/push/delete.spec.ts
@@ -25,6 +25,7 @@ describe('source:push:delete', () => {
       },
     },
   });
+  
   const doMockAxiosSuccess = (status: number, text: string) => ({
     status,
     statusText: text,
--- a/packages/cli/src/commands/source/push/delete.spec.ts
+++ b/packages/cli/src/commands/source/push/delete.spec.ts
@@ -1,0 +1,259 @@
+jest.mock('../../../lib/config/config');
+jest.mock('../../../hooks/analytics/analytics');
+jest.mock('../../../hooks/prerun/prerun');
+jest.mock('../../../lib/platform/authenticatedClient');
+jest.mock('@coveo/push-api-client');
+
+import {mocked} from 'ts-jest/utils';
+import {test} from '@oclif/test';
+import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
+import {Source} from '@coveo/push-api-client';
+const mockedClient = mocked(AuthenticatedClient);
+const mockedSource = mocked(Source);
+
+describe('source:push:delete', () => {
+  const doMockAxiosError = (
+    status: number,
+    message: string,
+    errorCode: string
+  ) => ({
+    response: {
+      status,
+      data: {
+        errorCode,
+        message,
+      },
+    },
+  });
+  const doMockAxiosSuccess = (status: number, text: string) => ({
+    status,
+    statusText: text,
+  });
+
+  const mockDeleteOlderThan = jest
+    .fn()
+    .mockReturnValue(Promise.resolve(doMockAxiosSuccess(202, 'tiguidou')));
+
+  const mockeleteDocument = jest
+    .fn()
+    .mockReturnValue(Promise.resolve(doMockAxiosSuccess(202, 'right trou')));
+
+  mockedClient.mockImplementation(
+    () =>
+      ({
+        cfg: {
+          get: () =>
+            Promise.resolve({
+              accessToken: 'the_token',
+              organization: 'the_org',
+            }),
+        },
+      } as unknown as AuthenticatedClient)
+  );
+
+  mockedSource.mockImplementation(
+    () =>
+      ({
+        deleteDocumentsOlderThan: mockDeleteOlderThan,
+        deleteDocument: mockeleteDocument,
+      } as unknown as Source)
+  );
+
+  beforeEach(() => {
+    mockDeleteOlderThan.mockClear();
+    mockeleteDocument.mockClear();
+  });
+
+  test
+    .stdout()
+    .command(['source:push:delete', 'mysource'])
+    .catch(/You must provide either --delete= or --deleteOlderThan=/)
+    .it('throws when no flags are specified');
+
+  test
+    .stdout()
+    .command(['source:push:delete', 'mysource', '-x', 'foo', '-d', 'bar'])
+    .catch(/--delete= cannot also be provided when using --deleteOlderThan=/)
+    .it(
+      'throws when incompatible flags for olderThan and documentUri are passed'
+    );
+
+  test
+    .stdout()
+    .command(['source:push:delete', 'mysource', '-d', '2000/01/01'])
+    .it('pass correct configuration information to push-api-client', () => {
+      expect(mockedSource).toHaveBeenCalledWith('the_token', 'the_org');
+    });
+
+  test
+    .do(() => {
+      mockDeleteOlderThan.mockReturnValueOnce(
+        doMockAxiosSuccess(999, 'this document is gone')
+      );
+    })
+    .stdout()
+    .command(['source:push:delete', 'mysource', '-d', '12345'])
+    .it(
+      'returns an information message on successful deletion with older than',
+      (ctx) => {
+        expect(ctx.stdout).toContain(
+          'Successfully deleted document: older than 12345'
+        );
+        expect(ctx.stdout).toContain('Status code: 999 this document is gone');
+      }
+    );
+
+  test
+    .do(() => {
+      mockDeleteOlderThan.mockRejectedValueOnce(
+        doMockAxiosError(
+          412,
+          'this is a bad request and you should feel bad',
+          'BAD_REQUEST'
+        )
+      );
+    })
+    .stdout()
+    .stderr()
+    .command(['source:push:delete', 'mysource', '-d', '12345'])
+    .it(
+      'returns an information message on deletion failure with older than',
+      (ctx) => {
+        expect(ctx.stdout).toContain(
+          'Error while trying to delete document: older than 12345'
+        );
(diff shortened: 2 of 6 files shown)


packages/cli/package.json
```diff
@@ -7,6 +7,7 @@
   },
   "dependencies": {
     "@angular/cli": "^11.1.4",
+    "@coveo/push-api-client": "^1.0.0",
```
louis-bompart, line 10: Nice 😄

packages/cli/src/commands/source/push/delete.spec.ts
```diff
@@ -0,0 +1,259 @@
...
+  ) => ({
+    response: {
+      status,
+      data: {
+        errorCode,
+        message,
+      },
+    },
+  });
+  const doMockAxiosSuccess = (status: number, text: string) => ({
```
louis-bompart, line 28:
```suggestion
  });
  
  const doMockAxiosSuccess = (status: number, text: string) => ({
```

watchout the suggestion tho, done from windows, this brk line might be a crlf

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,165 @@
...
+  Preconditions,
+} from '../../../lib/decorators/preconditions';
+import dedent from 'ts-dedent';
+import {
+  buildAnalyticsFailureHook,
+  buildAnalyticsSuccessHook,
+} from '../../../hooks/analytics/analytics';
+import {green, red} from 'chalk';
+// eslint-disable-next-line node/no-extraneous-import
+import {AxiosResponse} from 'axios';
```
louis-bompart, line 15:
```suggestion
import type {AxiosResponse} from 'axios';
```

I reckon this should work.

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,165 @@
...
+
+interface ErrorFromAPI {
+  response: {
+    status: number;
+    data: {
+      errorCode: string;
+      message: string;
+    };
+  };
+}
```
louis-bompart, line 25: That seems pretty generic for an error, do you think we should extract it in an utils or will it be a one-use interface? 🤔

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export default class SourcePushDelete extends Command {
+  public static description =
+    'Delete one or multiple documents in a given push source. See https://docs.coveo.com/en/171 and https://docs.coveo.com/en/131';
```
louis-bompart, line 29: Let's check with @fbeaudoincoveo , I reckon he likes those `<>` around URIs :P

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  public static flags = {
+    deleteOlderThan: flags.string({
+      char: 'd',
+      exclusive: ['delete'],
+      description:
+        'Delete old items, using either an ISO 8601 date or a Unix timestamp',
+      helpValue: '2000-01-01T00:00:00-06:00 [ OR ] 946702800',
```
louis-bompart, line 37: Hm I don't get the [ ] around the OR.

At first I thought it might be some odd syntax like, `2000-01-01T00:00:00-06:00 [` being 'starting from 2000-01-01T00:00:00-06:00, excluding it' and `] 946702800` being until 946702800, excluding it.

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,165 @@
...
+      this.warn(
+        dedent(
+          'To delete large batch of documents, use source:push:batch command instead.'
+        )
+      );
+      return true;
+    }
+
+    return false;
+  }
```
louis-bompart, line 99: I would put the warn in the `run` because the warning goes against the name of the function IMO: The `isSomethingSomething` imo should not alter anything but just run a check and return true or false

louis-bompart (commented):
Some tweaks here and there

packages/cli/src/commands/source/push/delete.ts
olamothe: I'm trying to explain in the help value that multiple date format are accepted, either an ISO date or a timestamp.

The output when you `--help` is something like: 

```
--deleteOlderThan=2000-01-01T00:00:00-06:00 [ OR ] 946702800
```

Open to suggestion !

packages/cli/src/commands/source/push/delete.ts
olamothe: Good point.

I think this will also be used for `:push:add`.

However, I'll do the change/extract shared interface for both when I work on that command.

packages/cli/src/commands/source/push/delete.ts
louis-bompart: 👍 I would just put OR but not the `[` `]`

louis-bompart (approved):
LGTM 👌 
Lil doc review could be nice and :shipit:

y-lakhdar (approved):

olamothe:
@jpmarceau : I will go ahead and merge this PR since I need it for other things. If there's string revision you want to add, feel free to open them in this PR, and I will apply them at a later stage.

packages/cli/src/commands/source/push/delete.ts
fbeaudoincoveo: Haha yea

I'll open another PR to review this documentation :)

olamothe: @fbeaudoincoveo : You can just comment on this PR and I'll process it.

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,164 @@
...
+        'The identifier of the source on which to perform the delete operation. See source:push:list to obtain the identifier.',
+    },
+  ];
+
+  @Preconditions(IsAuthenticated())
+  public async run() {
+    const {flags} = this.parse(SourcePushDelete);
+    if (!flags.deleteOlderThan && !flags.delete) {
+      this.error(
+        'You must provide either --delete= or --deleteOlderThan=. See source:push:delete --help for more information.'
```
fbeaudoincoveo, line 70:
```
'You must minimally set the `delete` or the `deleteOlderThan` option. Use `source:push:delete --help` to get more information.'
```

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (flags.deleteOlderThan) {
+      this.doDeletionOlderThan(source);
+      return;
+    }
+
+    if (flags.delete) {
+      if (this.isNumberOfDeletionTooLarge) {
+        this.warn(
+          dedent(
+            'To delete large batch of documents, use source:push:batch command instead.'
```
fbeaudoincoveo, line 85:
```
'To delete large number of documents, use the `source:push:batch` command instead.'
```

But of course if we don't allow the delete option to be repeated, we won't even need this message anymore.

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,164 @@
...
+  Error code: ${red(e.response.data.errorCode)}
+  Message: ${red(e.response.data.message)}
+  `)
+    );
+  }
+
+  private successMessageOnDeletion(toDelete: string, res: AxiosResponse) {
+    this.log(
+      dedent(`
+    Successfully deleted document: ${green(toDelete)}
```
fbeaudoincoveo, line 150: If you're getting the success code from the Push API, then it doesn't necessarily mean that the operation was successful; it only means that the request was accepted by the Push API. The operation could still fail in the indexing pipeline.

Suggestion:

`The delete request for document: ${green(toDelete)} was accepted by the Push API`

Question: what will `toDelete` contain if only `deleteOlderThan` (not `delete`) was specified?

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,164 @@
...
+export default class SourcePushDelete extends Command {
+  public static description =
+    'Delete one or multiple documents in a given push source. See https://docs.coveo.com/en/171 and https://docs.coveo.com/en/131';
+
+  public static flags = {
+    deleteOlderThan: flags.string({
+      char: 'd',
+      exclusive: ['delete'],
+      description:
+        'Delete old items, using either an ISO 8601 date or a Unix timestamp',
```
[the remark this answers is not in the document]
olamothe: Exactly, it's one or the other (either deleteOlderThan or by documentId)

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,164 @@
...
+      helpValue: '2000-01-01T00:00:00-06:00 OR 946702800',
+    }),
+    delete: flags.string({
+      char: 'x',
+      multiple: true,
+      description:
+        'Document URI or identfier to delete. Can be repeated. If you want to delete a large batch of documents, use source:push:batch command instead.',
```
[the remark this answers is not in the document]
olamothe: This is handled in the code. If it goes above a number of documentId, the command won't process and will inform you to use batch instead.

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,164 @@
...
+    };
+  };
+}
+
+interface AxiosResponse {
+  status: number;
+  statusText: string;
+}
+
+export default class SourcePushDelete extends Command {
```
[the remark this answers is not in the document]
olamothe: We filter on host type: `PUSH`.
Crawling module should not be returned. For Sitecore.. I'm not sure.

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -0,0 +1,164 @@
...
+  public static description =
+    'Delete one or multiple documents in a given push source. See https://docs.coveo.com/en/171 and https://docs.coveo.com/en/131';
+
+  public static flags = {
+    deleteOlderThan: flags.string({
```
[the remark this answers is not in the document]
olamothe: Yes, using default.

packages/cli/src/commands/source/push/delete.ts
olamothe: The delete request for document: older than 12345 was accepted by the Push API

PR #349, opened 2021-07-15
feat(push): add support for push:add command 

olamothe:
## Proposed changes

Add a `source:push:add` command, which supports both folder and files containing JSON documents.

Parse each JSON document and tries to do some validation about missing properties.

After parsing documents, create a batch of update and upload to push API.

What's missing in this PR and that should come in upcoming PR: 

* UT
* Support for security on documents (currently ignored)
* Do some manual tests for with a * lot * of documents to see if it 💥 . I wanna try ~500k JSON documents or something like this. I'll working on this tomorrow most probably.
 

## Testing

- [X] Manual Tests:

Simple manual tests for now using the documents added in __stub__/documentsjson
UT will come on a later PR when I've had the chance to test a bigger batch of documents. 

But the PR was becoming pretty big already...

https://coveord.atlassian.net/browse/CDX-465


packages/cli/package.json
```diff
@@ -6,6 +6,7 @@
     "node": ">=14.15.1"
   },
   "dependencies": {
+    "@coveo/bueno": "^0.1.0",
```
olamothe, line 9: That's a generic "parameter validation" library we've built for headless.

I'm using it to validate documents properties in`requiredKeyValidator.ts`

packages/cli/src/lib/push/caseInsensitiveDocument.ts
```diff
@@ -0,0 +1,8 @@
+export class CaseInsensitiveDocument<T> {
```
olamothe, line 1: I'm using case insensitive check on "required keys" and "known keys", in order for  `documentid, documentId, DocumentID` etc to all be equal and "correct".

Less annoying for users IMO.

packages/cli/src/lib/push/requiredKeyValidator.ts
```diff
@@ -0,0 +1,60 @@
+import {SchemaValue} from '@coveo/bueno';
+import {CaseInsensitiveDocument} from './caseInsensitiveDocument';
+
+export class RequiredKeyValidator<T> {
```
olamothe, line 4: Used to validate that some properties are present(using bueno).

When they are not present, bueno will throw an error with a meaningful message (eg: document XYZ has no documentID, or documentXYZ has a documentID that is not a valid URL, or documentXYZ has a title that is not a string, etc.)

packages/cli/package.json
louis-bompart: Nice reusing!

The more I see it, the more I think we should extract it and put it in its own repo to promote its reusability in @coveo at least

packages/cli/src/commands/source/push/add.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+    const cfg = await new AuthenticatedClient().cfg.get();
+    const source = new Source(cfg.accessToken!, cfg.organization);
+
+    if (flags.file) {
+      await this.doPushFile(source);
+    }
+    if (flags.folder) {
+      await this.doPushFolder(source);
+    }
```
louis-bompart, line 69: It should be either file or folder, right?

If so I would do something like that:
```suggestion
    const cfg = await new AuthenticatedClient().cfg.get();
    const source = new Source(cfg.accessToken!, cfg.organization);
    let isFolderXORFileSet = false;
    if (flags.file) {
      await this.doPushFile(source);
      isFolderOrFileSet = !isFolderOrFileSet;
    }
    if (flags.folder) {
      await this.doPushFolder(source);
      isFolderOrFileSet = !isFolderOrFileSet;
    }
    if (!isFolderXORFileSet) {
      this.error(
        'You must provide either --file= or --folder=. See source:push:add --help for more information.'
      );
    }
```

packages/cli/src/commands/source/push/add.ts
```diff
@@ -0,0 +1,154 @@
...
+    });
+    this.successMessageOnAdd(flags.file, docBuilders.length, res);
+  }
+
+  private successMessageOnAdd(
+    files: string[],
+    numAdded: number,
+    res: AxiosResponse
+  ) {
+    let fileNames = files.slice(0, 5).join(', ');
```
louis-bompart, line 123: That's a bit cryptic, what the goal of this line?

packages/cli/src/lib/push/requiredKeyValidator.ts
```diff
@@ -0,0 +1,60 @@
...
+  private keys: string[];
+  private validations: {value: T; message: string | null}[] = [];
```
louis-bompart, line 6: null? 🤔 the API return null?

packages/cli/tsconfig.json
```diff
@@ -8,12 +8,12 @@
     "outDir": "lib",
     "rootDir": "src",
     "strict": true,
-    "target": "es2017",
+    "target": "es2019",
```
louis-bompart, line 11: Validated with https://node.green/ ✔️

louis-bompart (approved):
lgtm overall, some tweaks maybe

packages/cli/src/commands/source/push/add.ts
olamothe: It's the same end result though 🤔 Why do you prefer this approach ? Just curious.

packages/cli/package.json
olamothe: Why do you think putting it in it's own repo would make this more re-usable ? It is published as a standalone library on npm, like any other.. 🤔

packages/cli/src/commands/source/push/add.ts
olamothe: I can add a comment for explanation.

packages/cli/src/commands/source/push/add.ts
y-lakhdar: I find the initial way easier to read.

packages/cli/src/lib/push/requiredKeyValidator.ts
```diff
@@ -0,0 +1,60 @@
...
+    } else {
+      this.keys = [k];
+    }
+    this.doValidations();
+  }
+
+  private doValidations() {
+    this.validations = this.keys.map((k) => {
+      const value = this.doc.documentRecord[k.toLowerCase()];
+      const validationMessage = this.schema.validate(value);
```
y-lakhdar, line 24: Nice!

y-lakhdar (approved):

packages/cli/src/commands/source/push/add.ts
olamothe: I will leave it as is then. 

If there's any strong objection, please ping back.

PR #354, opened 2021-07-16
test: specify test org

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-468

## Proposed changes

Since `auth:login` connects to the first available org, we need to make sure some E2E tests always connect to the same org.

-----
CDX-468


.github/workflows/build.yml
```diff
@@ -16,6 +16,7 @@ jobs:
     env:
       PLATFORM_USER_NAME: ${{ secrets.PLATFORM_USER_NAME }}
       PLATFORM_USER_PASSWORD: ${{ secrets.PLATFORM_USER_PASSWORD }}
+      ORG_ID: ${{ secrets.ORG_ID }}
```
louis-bompart, line 19: Fine by me to use secrets as 'easy to update' config.

But let's not lure ourselves, this bit (the org id) ain't secure at all: we log it in the initial-login logs for e.g. :P

louis-bompart (approved):

.github/workflows/build.yml
y-lakhdar: haha. I know. it's just to avoid hardcoding it in the tests

louis-bompart (changes requested):
You outta update `packages\cli-e2e\__tests__\auth.specs.ts`

It's "less" critical, but ideally I'd prefer to keep a 1-to-1 behavior on the E2E tests

y-lakhdar:
> You outta update `packages\cli-e2e\__tests__\auth.specs.ts`
> 
> It's not critical, but ideally, I'd prefer to keep a 1-to-1 behaviour on the E2E tests (i.e. I'd prefer to keep creatin' API Keys in the default org, not a randy random test org)

good catch

louis-bompart (approved):

PR #358, opened 2021-07-19
feat(push): add support for very large batch of documents

olamothe:
## Proposed changes

Rework upload command to split by chunks manageable by the Push API. 
I was able to test and push ~500k documents in a source in a couple of seconds (have the documents accepted by the Push API).

The end result for the user is something like this: 

image

☝️ Pushing a folder with 2 files (both of them have the same number of documents because it is fake data).

Users don't have to worry about splitting the size of their files or how many documents there are, it should all be handled automatically for them.

## Testing

- [X] Manual Tests:

https://coveord.atlassian.net/browse/CDX-471


packages/cli/src/commands/source/push/add.ts
```diff
@@ -154,4 +172,61 @@ export default class SourcePushAdd extends Command {
     );
     throw err;
   }
+
+  private splitByChunkAndUpload(
```
olamothe, line 176: The main function of this PR: 

Create and return two function: 

* One to push documents (send) which accumulate documents until they reach a maxContentLength size. When the limit is reached, push the documents to Coveo (`uploadBatch` function).
* One to "close" when there's no more files to parse and convert to `DocumentBuilder`, so that we can flush the remaining queue (remaining documents in `accumulator.chunks`)

y-lakhdar: Nicely done

y-lakhdar (approved):

PR #360, opened 2021-07-19
test(login): fix login selectors

louis-bompart:
## Proposed changes

Use the recommended way of 'click and waiting for navigation' mostly, and did something similar for the 'view'. The latter should ensure that when trying to submit the password, it doesn't click on the submit button for the previous view.

## Testing

Local FT run + CI 

-----
CDX-475


y-lakhdar (approved):

louis-bompart:
See https://pptr.dev/#?product=Puppeteer&version=v10.1.0&show=api-pageclickselector-options

olamothe (approved):

PR #361, opened 2021-07-20
test(push): add ut for source:push:add

olamothe:
## Proposed changes

Adding UT for `:add` command. 

## Testing

- [X] Unit Tests:

https://coveord.atlassian.net/browse/CDX-473


packages/cli/src/commands/source/push/add.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        errorCode,
+        message,
+      },
+    },
+  });
+
+  const doMockAxiosSuccess = (status: number, text: string) => ({
+    status,
+    statusText: text,
+  });
```
louis-bompart, line 34: Let's refactor that I think ;)

It's already in `delete.spec.ts`

packages/cli/src/commands/source/push/add.ts
```diff
@@ -62,9 +62,10 @@ export default class SourcePushAdd extends Command {
     const {flags} = this.parse(SourcePushAdd);
     if (!flags.file && !flags.folder) {
       this.error(
-        'You must provide either --file= or --folder=. See source:push:add --help for more information.'
+        'You must minimally set the `file` or the `folder` flag. Use `source:push:add --help` to get more information.'
```
louis-bompart, line 65: Maybe me, but doesn't sounds... "right". Invoking @jpmarceau :P

louis-bompart (commented):
lgtm overall.

Needs a refactor, plus san-check on the string

packages/cli/src/commands/source/push/add.ts
olamothe: I copied the message from `delete`, which had been revised by Fred Beaudoin.

packages/cli/src/commands/source/push/add.spec.ts
olamothe: Indeed !

packages/cli/src/commands/source/push/add.ts
louis-bompart: minimally sounds odd, maybe it was just me okok, can be ignored 👍

louis-bompart (approved):

PR #363, opened 2021-07-20
fix(push): fix incorrect types for timestamp when using delete older than

olamothe:
## Proposed changes

Fix issue when doing deleteOlderThan using timestamp

## Testing

- [X] Unit Tests:
- [X] Manual Tests:

https://coveord.atlassian.net/browse/CDX-477

### DIFF
--- a/packages/cli/src/commands/source/push/delete.ts
+++ b/packages/cli/src/commands/source/push/delete.ts
@@ -104,20 +104,12 @@ export default class SourcePushDelete extends Command {
     const {flags, args} = this.parse(SourcePushDelete);
     const toDelete = `older than ${flags.deleteOlderThan}`;
     try {
-      const isNumber = flags.deleteOlderThan?.match(/^[0-9]+$/);
+      const isNumber = flags.deleteOlderThan?.match(/^\d+$/);
 
-      let res: AxiosResponse;
-      if (!isNumber) {
-        res = await source.deleteDocumentsOlderThan(
+      const res = await source.deleteDocumentsOlderThan(
           args.sourceId,
-          flags.deleteOlderThan!
+          isNumber ? parseInt(flags.deleteOlderThan!, 10) : flags.deleteOlderThan!
         );
-      } else {
-        res = await source.deleteDocumentsOlderThan(
-          args.sourceId,
-          parseInt(flags.deleteOlderThan!, 10)
-        );
-      }
       this.successMessageOnDeletion(toDelete, res);
     } catch (e) {
       this.errorMessageOnDeletion(toDelete, e);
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/source/push/delete.ts
```diff
@@ -104,10 +104,20 @@ export default class SourcePushDelete extends Command {
     const {flags, args} = this.parse(SourcePushDelete);
     const toDelete = `older than ${flags.deleteOlderThan}`;
     try {
-      const res = await source.deleteDocumentsOlderThan(
-        args.sourceId,
-        flags.deleteOlderThan!
-      );
+      const isNumber = flags.deleteOlderThan?.match(/^[0-9]+$/);
```
y-lakhdar, line 107: Why not use isNaN

y-lakhdar (approved):

packages/cli/src/commands/source/push/delete.ts
olamothe: Because it is "unreliable".

`new Number('2001/01/01')` -> NaN
`parseInt('2001/01/01')` -> 2000 
`isNaN('2001/01/01)` -> true, but that's not valid TypeScript. `isNan` accept a number, not a string.

I could do `isNaN(new Number(flags.deleteOlderThan))` if you guys prefer/find it more meaningful !

y-lakhdar: No. the regex is simpler. I was just wandering why the regex over the `isNaN`.
Thanks for the explication.

louis-bompart:
```suggestion
      const isNumber = flags.deleteOlderThan?.match(/^\d+$/);
```
I tend to prefer this, but y'know, taste and colors.

The `isNaN(new Number(flags.deleteOlderThan))` is nice too, no RegEx is the best regex often.

packages/cli/src/commands/source/push/delete.ts
```diff
@@ -104,10 +104,20 @@ export default class SourcePushDelete extends Command {
...
+        res = await source.deleteDocumentsOlderThan(
+          args.sourceId,
+          flags.deleteOlderThan!
+        );
+      } else {
+        res = await source.deleteDocumentsOlderThan(
+          args.sourceId,
+          parseInt(flags.deleteOlderThan!, 10)
+        );
+      }
```
louis-bompart, line 120: I would do:

```suggestion
      const res = await source.deleteDocumentsOlderThan(
          args.sourceId,
          isNumber ? parseInt(flags.deleteOlderThan!, 10) : flags.deleteOlderThan!
        );
```

louis-bompart (approved):
lgtm.
Added my two-cents on the regex thread.
I would prefer a ternary for the number/NaN branching, to avoid code repetition.

PR #365, opened 2021-07-20
fix(snapshot): pull should use new enum

louis-bompart:
## Proposed changes

Updated which enum to use, see https://github.com/coveo/platform-client/pull/351

## Testing

Manual test. The command goes kaboom without the changes.
- [ ] UT: to fix

-----
CDX-485

### DIFF
--- a/packages/angular/package-lock.json
+++ b/packages/angular/package-lock.json
@@ -147,9 +147,9 @@
       }
     },
     "@babel/runtime": {
-      "version": "7.14.6",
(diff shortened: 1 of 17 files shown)


y-lakhdar (approved):

y-lakhdar (changes requested):
You forgot to update the mocks on 
* snapshot.spec.ts
* snapshotFactory.spec.ts

y-lakhdar (approved):

olamothe (approved):

PR #366, opened 2021-07-20
feat(push): support security identities in JSON documents

olamothe:
https://coveord.atlassian.net/browse/CDX-481

## Proposed changes

Add support to parse security identities in JSON documents.

Goes hand in hand with this PR: https://github.com/coveo/push-api-client.js/pull/21

## Testing

- [X] Unit Tests:
- [X] Manual Tests:

### DIFF
--- a/packages/cli/src/lib/push/parseFile.ts
+++ b/packages/cli/src/lib/push/parseFile.ts
@@ -1,4 +1,10 @@
-import {ArrayValue, BooleanValue, RecordValue, StringValue} from '@coveo/bueno';
+import {
+  ArrayValue,
+  BooleanValue,
+  PrimitivesValues,
+  RecordValue,
+  StringValue,
+} from '@coveo/bueno';
 import {
   DocumentBuilder,
   MetadataValue,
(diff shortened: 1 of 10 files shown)


packages/cli/src/lib/push/parseFile.ts
```diff
@@ -26,17 +32,28 @@ export const parseAndGetDocumentBuilderFromJSONDocument = (
...
+  );
+  processSecurityIdentities(
+    caseInsensitiveDoc as CaseInsensitiveDocument<Document['permissions']>,
+    documentBuilder,
+    documentPath
+  );
+  processMetadata(
+    caseInsensitiveDoc as CaseInsensitiveDocument<MetadataValue>,
+    documentBuilder
+  );
```
louis-bompart, line 56: I think there's really too much casting.

That smells like a logic that reaches its ends and that you fight against it.

Maybe it's time to rethink this?

louis-bompart (commented):
Lot of casting, I think it's probably time to rethink certain classes here..

packages/cli/src/lib/push/parseFile.ts
olamothe: I'll make use of the `PrimitiveValues` type from Bueno a bit more.
Still, there's some casting that needs to happen since, inherently, all values in an arbitrary JSON documents are "unknown" upfront.

packages/cli/src/lib/push/requiredKeyValidator.ts
```diff
@@ -1,13 +1,13 @@
-import {SchemaValue} from '@coveo/bueno';
+import {PrimitivesValues, SchemaValue} from '@coveo/bueno';
 import {CaseInsensitiveDocument} from './caseInsensitiveDocument';
 
-export class RequiredKeyValidator<T> {
+export class RequiredKeyValidator<T extends PrimitivesValues> {
   private keys: string[];
-  private validations: {value: T; message: string | null}[] = [];
+  private validations: {value: T | unknown; message: string | null}[] = [];
```
louis-bompart, line 6: 👍 as you said in the PR, lot of value are unknown. I think it's really nice to reflect that semantically thru the typings.

louis-bompart (approved):

PR #371, opened 2021-07-21
test(e2e): `org:config:*` commands

y-lakhdar:
## Proposed changes

Testing:
* `org:config:preview`
* `org:config:pulsh`
* `org:config:pull`

Created a folder to contain the content of a dummy snapshot.

-----
CDX-460


packages/cli-e2e/__tests__/orgConfig.specs.ts
```diff
@@ -0,0 +1,158 @@
...
+    );
+  });
+
+  describe('org:config:push', () => {
+    beforeAll(async () => {
+      await pushToOrg(testOrgId, processManager);
+    }, defaultTimeout);
+
+    it('should have pushed fields', async () => {
+      const fields: unknown[] = (await listFields(testOrgId, accessToken))
```
olamothe, line 119: I'm pretty sure we could have better typings than `unknown []` here. 
A list of fields is a known type, that you could probably import from platform-client.

packages/cli-e2e/__tests__/orgConfig.specs.ts
```diff
@@ -0,0 +1,158 @@
...
+      expect(fields).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({name: 'firstfield'}),
+          expect.objectContaining({name: 'whereisbrian'}),
+        ])
+      );
+    });
+
+    it('should have pushed extensions', async () => {
+      const extensions: unknown[] = await listExtensions(
```
olamothe, line 130: Same as above related to better types.

packages/cli-e2e/utils/platform.ts
```diff
@@ -4,6 +4,28 @@ import {HTTPRequest} from 'puppeteer';
 // TODO: CDX-98: URL should vary in function of the target environment.
 export const platformHost = 'https://platformdev.cloud.coveo.com/rest/';
 
+export async function listPlatformResource(
```
olamothe, line 7: Any reason we are not using platform client here also to list resources from the platform ?

olamothe (approved):
💯 Nice, the e2e for this seems easier than for the `ui:*` commands :D

packages/cli-e2e/utils/platform.ts
y-lakhdar: Since we were already using axios for most of the HTTPs requests, I decided to stick with it.

But I also think it would be cleaner to use the platform-client.

packages/cli-e2e/__tests__/orgConfig.specs.ts
```diff
@@ -0,0 +1,157 @@
...
+        const previewTerminal = previewChange(testOrgId, processManager);
+
+        const expectedOutput = [
+          'Extensions',
+          '\\+   1 to create',
+          'Fields',
+          '\\+   2 to create',
+          'Filters',
+          '\\+   1 to create',
+        ].join('\\s*');
```
louis-bompart, line 107: What does the final regexp looks like?

y-lakhdar: That's not the output from the E2E test, but it looks something like that
image

louis-bompart (approved):

PR #373, opened 2021-07-22
ci(cdn): invalidate cache on prod release 

olamothe:
## Proposed changes

Added a new deployment phases (that runs only in prod) using the "ordered_phases" property of the deployment pipeline.

After binaries have been uploaded to S3, do a Cloudfront cache invalidation so that end users can immediately upgrade when it is released.

Without this, it can take up to 24 hours for new version to be detected by the update command.

Deployment pipeline can only trigger Jenkins job (for now?) so there's a small `JenkinsfileCloudfrontInvalidation ` file that is run through Jenkins job called `cli-cloudfront-invalidation` in the team Jenkins.

The AWS creds are also stored in Jenkins under the `CloudfrontCacheInvalidation` ID.

## Testing

- [X] Manual Tests:

https://coveord.atlassian.net/browse/CDX-495

### DIFF
--- a/.deployment.config.json
+++ b/.deployment.config.json
@@ -11,17 +11,30 @@
       "sequential": ["us-east-1"]
     }
   },
-  "phases": {
-    "terraform": {},
-    "s3": {
-      "bucket": "{terraform.infra.infra.bucket_binaries}",
-      "directory": "proda/StaticCDN/cli",
-      "source": "./artifacts",
-      "parameters": {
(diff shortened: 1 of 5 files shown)


louis-bompart (approved):
LGTM 👍 

If you're motivated, you could create a dedicated package.json for the `./scripts` folder, it'll save ya time by installing only the bare minimum.

No need for it to be lerna managed too IMO

PR #374, opened 2021-07-22
feat(config): add list snapshot command

olamothe:
https://coveord.atlassian.net/browse/CDX-498

## Proposed changes

Add a `org:config:list` command.
It can be used in conjunction/in tandem with `org:config:pull` which accept a snapshot id as a flag.

## Testing

- [X] Unit Tests:
- [X] Manual Tests:

https://coveord.atlassian.net/browse/CDX-498

### DIFF
--- a/packages/cli/src/commands/org/config/list.ts
+++ b/packages/cli/src/commands/org/config/list.ts
@@ -27,7 +27,7 @@ export default class List extends Command {
     }),
   };
 
-  public static args = [{name: 'file'}];
+  public static hidden = true;
 
   @Preconditions(IsAuthenticated())
   public async run() {
@@ -39,10 +39,9 @@ export default class List extends Command {
     const snapshots = await platformClient.resourceSnapshot.list();
     if (snapshots.length === 0) {
       this.log(
-        dedent(`
-      There is no configuration snapshot available in organization ${org}
-      You can create one using org:config:pull
-      `)
+        dedent(
+          `There is no configuration snapshot available in organization ${org}`
+        )
       );
       return;
     }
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/org/config/list.ts
```diff
@@ -0,0 +1,76 @@
...
+  IsAuthenticated,
+  Preconditions,
+} from '../../../lib/decorators/preconditions';
+import {
+  buildAnalyticsFailureHook,
+  buildAnalyticsSuccessHook,
+} from '../../../hooks/analytics/analytics';
+import dedent from 'ts-dedent';
+
+export default class List extends Command {
```
y-lakhdar, line 16: Should we hide all the `org:config:` commands before we get them all tested?
```
public static hidden = true;
```

packages/cli/src/commands/org/config/list.ts
```diff
@@ -0,0 +1,76 @@
...
+    target: flags.string({
+      char: 't',
+      description:
+        'The unique identifier of the organization containing the snapshots. If not specified, the organization you are connected to will be used.',
+      helpValue: 'destinationorganizationg7dg3gd',
+      required: false,
+    }),
+  };
+
+  public static args = [{name: 'file'}];
```
y-lakhdar, line 30: I don't think this is needed.

packages/cli/src/commands/org/config/list.ts
```diff
@@ -0,0 +1,76 @@
...
+    const org = await getTargetOrg(this.configuration, flags.target);
+    const platformClient = await new AuthenticatedClient().getClient({
+      organization: org,
+    });
+    const snapshots = await platformClient.resourceSnapshot.list();
+    if (snapshots.length === 0) {
+      this.log(
+        dedent(`
+      There is no configuration snapshot available in organization ${org}
+      You can create one using org:config:pull
```
y-lakhdar, line 44: I don't think this line is needed since the `pull` command will delete the snapshot once its content is added to the user's workspace.

y-lakhdar (changes requested):
Nice!
Just need to hide the command

y-lakhdar (approved):

PR #377, opened 2021-07-26
fix(cli): command timing out without showing info

y-lakhdar:
## Proposed changes

The analytics hook is ending the program execution, which makes the `handleSnapshotError` method unreachable. The issue has something to do with the `analyticsClient.sendCustomEvent` method.
Since the root cause is still unknown, I am moving all the CLI logic before sending analytics.

I also replaced `cli.info` with `cli.log` since `log` basically calls `info` ([link](https://github.com/oclif/cli-ux/blob/6dfa79da103cc608dc1b7b4790eb1a084c435c5b/src/index.ts#L82)).
## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:
I packed a version of the cli in which I moved the analytic hooks at the end.

-----
CDX-502


packages/cli/src/lib/snapshot/snapshotCommon.ts
```diff
@@ -79,14 +79,14 @@ export async function getTargetOrg(config: Config, target?: string) {
 
 export function handleSnapshotError(err?: Error) {
   if (err instanceof SnapshotOperationTimeoutError) {
-    cli.info(operationGettingTooMuchTimeMessage(err.snapshot));
```
louis-bompart, line 82: So did you find what's the differences between `info` and `log`?

louis-bompart (approved):

packages/cli/src/lib/snapshot/snapshotCommon.ts
y-lakhdar: yeah. `log` is just calling `info`.
I also did that switch since we are using `cli.log` everywhere else in the code.

olamothe (approved):

olamothe:
Did you open a ticket for the analytics failure (and add it to CDX-505)

packages/cli/src/lib/snapshot/snapshotCommon.ts
louis-bompart: Ok so switching for uniformity 👍 all's well

PR #378, opened 2021-07-26
fix(cli): wait for snapshot to be done

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-503

## Proposed changes

Pulling a snapshot by its ID will cause an error since the command tries to download the snapshot in a pending state.
The fix is simply to add a `waitUntilDone` promise to make sure the snapshot is idle before manipulating it.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:
1. Create a big snapshot.
2. Pull the snapshot while it is still pending.
3. It will show a spinner animation instead of breaking 

CDX-503


louis-bompart (approved):
lgtm!

olamothe (approved):

PR #379, opened 2021-07-26
feat(config): initialize git repo when doing initial pull

olamothe:
## Proposed changes

Adding support for `.git` project initialization, as well as `.coveo` project initialization when creating a `Project`. There is some overlap with https://github.com/coveo/cli/pull/267, specifically related to the creation of the hidden `.coveo` folder, which may contain multiple files eventually (for preview and for local config).

@louis-bompart Specifically looking for your input on if you think you would be able to leverage the `dotFolder` module to control the `preview` subfolder.
 

## Testing

- [X] Unit Tests:

https://coveord.atlassian.net/browse/CDX-504


packages/cli/src/commands/org/config/pull.ts
```diff
@@ -79,15 +87,19 @@ export default class Pull extends Command {
...
             `
       );
     }
   }
 
   private async refreshProject(snapshot: Snapshot) {
+    const {flags} = this.parse(Pull);
     const project = new Project(this.projectPath);
+    if (flags.git && !project.contains('.git')) {
+      exec(`git init ${this.projectPath}`);
```
louis-bompart, line 101: - [ ] Apply CodeQL recommendations

packages/cli/src/commands/org/config/pull.ts
```diff
@@ -45,11 +47,17 @@ export default class Pull extends Command {
       description:
         'The unique identifier of the snapshot to pull. If not specified, a new snapshot will be created.',
     }),
+    git: flags.boolean({
+      char: 'g',
+      description:
+        'Wether to create a git repository when creating a new project.',
```
louis-bompart, line 53: I do not think you meant ['castrated ram'](https://www.google.com/search?q=wether&oq=wether&aqs=chrome..69i57j0i271l3.252j0j7&sourceid=chrome&ie=UTF-8) 😂 

```suggestion
        'Whether to create a git repository when creating a new project.',
```

louis-bompart (changes requested):
Small fixes required, string correction and security advisory to answer to

packages/cli/src/lib/project/dotFolder.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  public static get hiddenFolderName() {
+    return '.coveo';
+  }
+
+  public get path() {
+    return join(this.owner.toString(), DotFolder.hiddenFolderName);
+  }
+
+  private ensureFolderExists() {
```
y-lakhdar, line 22: If you are using fs-extra, can you just use [ensureDir](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir.md)?

packages/cli/src/lib/project/dotFolder.ts
```diff
@@ -0,0 +1,53 @@
...
+    const exists = pathExistsSync(this.path);
+    if (!exists) {
+      createFileSync(this.path);
```
y-lakhdar, line 25: I think you meant `mkdirSync`

packages/cli/src/lib/project/dotFolder.ts
```diff
@@ -0,0 +1,53 @@
...
+  private get defaultConfig() {
+    return {
+      version: 1,
+      organization: '', // TODO: Needs to use that "local" config for all :config:* command that would rely on it.
+    };
+  }
+
+  private ensureFileExists() {
+    const path = join(this.ownerFolder.path, DotFolderConfig.configName);
+    if (!existsSync(path)) {
```
y-lakhdar, line 48: I think `writeJSONSync` already checks if the file already exists and creates it if that is not the case.

y-lakhdar (approved):

packages/cli/src/commands/org/config/pull.ts
louis-bompart, line 101: - Don't use non-synchronous methods from `child_process`. It's a mess to handle.
- Avoid to use `exec` to avoid [CWE-78](https://cwe.mitre.org/data/definitions/78.html) and [CWE-88](https://cwe.mitre.org/data/definitions/88.html). See CodeQL report for more information and remediation.
- Do use the existing utils for process Management, like `spawnSync`

louis-bompart (changes requested):

packages/cli/src/lib/decorators/preconditions/git.ts
```diff
@@ -0,0 +1,11 @@
+import {appendCmdIfWindows} from '../../utils/os';
+import {getBinInstalledPrecondition} from './binPreconditionsFactory';
+
+export const IsGitInstalled = getBinInstalledPrecondition(
+  appendCmdIfWindows`git`,
```
louis-bompart, line 5: git is a .exe on Windows, so you do not need `appendCmdIfWindows`
```suggestion
'git',
```
(iirc, based on what we do with Node, I don't think we need to add .exe either)

louis-bompart (approved):
just fix the git path for windows and :shipit: 👍

packages/cli/src/lib/project/project.spec.ts
```diff
@@ -55,18 +64,42 @@ describe('Project', () => {
...
+
+  it('should create .coveo project if absent', () => {
+    doMockFileDoesNotExists('.coveo');
+    mockedPathExistsSync.mockReturnValueOnce(false);
+    projectCreator();
+    expect(mockedCreateFileSync).toHaveBeenCalledWith(
+      expect.stringContaining('.coveo')
+    );
+    expect(mockedJSONSync).toHaveBeenCalledWith(
+      expect.stringContaining('.coveo/config.json'),
```
louis-bompart, line 100:
```suggestion
      expect.stringContaining(join('.coveo/config.json')),
```

You need this for your test to work on Windows because of the variations in path separator (`/` v.s `\\`)

louis-bompart (approved):
lgtm, fix the tests and :shipit: 🙌

PR #381, opened 2021-07-26
fix(config): fix race condition in config file

olamothe:
https://coveord.atlassian.net/browse/CDX-509

## Proposed changes

One time out of ~20, I would locally reproduce the problem and get errors manipulating the configuration while executing the analytics hook. I would end up with an empty config file.

In the analytics hook, there's some logic to access the config and to write the current visitorID (through the coveo.analytics library).

By moving to a sync method to write the new config content, everything seems 👍, and I can't reproduce by running the analytics hooks multiple times (~100) in quick succession.

## Testing

- [X] Manual Tests:


louis-bompart (commented):
The UT needs to be updated ;) 

Also, while at it from what I checked, I think the file could be put 'sync' because readJson etc as Sync alternative.

Why not do that while at it?

y-lakhdar (approved):
lgtm

PR #385, opened 2021-07-27
fix(react): remove console errors

y-lakhdar:
## Proposed changes

Fix all the issues causing an error in the browser console

## Testing
- [x] E2E Tests: 
- [x] Manual Tests: Created a React app

-----
CDX-513


packages/cra-template/template/src/Components/Pager.tsx
```diff
@@ -14,7 +14,10 @@ const PagerRenderer: FunctionComponent<PagerProps> = (props) => {
   const [state, setState] = useState(controller.state);
 
   useEffect(
-    () => controller.subscribe(() => setState(controller.state)),
+    () =>
+      controller.subscribe(() =>
+        process.nextTick(() => setState(controller.state))
+      ),
```
y-lakhdar, line 20: This is fixing the following error:  
```
Cannot update a component ("PagerRenderer") while rendering a different component ("ResultListRenderer")
```

In all the documentation I found, it says that the error can be prevented by using enclosing the `setState` method inside the `useEffect` hook (which is already the case here).
So, is it possible that the Pager and ResultList controllers are "dispatching" their new state at the same time, which causes the 2 components to trigger a page render at the same moment?

packages/cra-template/template/src/theme.tsx
```diff
@@ -1,8 +1,8 @@
 import red from '@material-ui/core/colors/red';
-import {createMuiTheme} from '@material-ui/core/styles';
+import {createTheme} from '@material-ui/core/styles';
```
y-lakhdar, line 2: `createMuiTheme` is deprecated

packages/cra-template/template/src/Components/SearchPage.tsx
```diff
@@ -25,7 +25,7 @@ const SearchPage: React.FunctionComponent<ISearchPageProps> = (props) => {
   return (
     <EngineProvider value={engine}>
       <Container maxWidth="lg">
-        <Grid container justify="center">
+        <Grid container justifyContent="center">
```
y-lakhdar, line 28: `justify` is deprecated

packages/cra-template/template/src/Components/Hero.tsx
```diff
@@ -62,16 +67,16 @@ const Hero: React.FunctionComponent<IHeroProps> = (props) => {
         Essential Links
       </Typography>
       <ul>
-        <li>
+        <li key="item-link-headless">
```
y-lakhdar, line 70: Each item in a list should have a key

packages/cra-template/template/src/Components/Hero.tsx
```diff
@@ -15,8 +15,13 @@ const Hero: React.FunctionComponent<IHeroProps> = (props) => {
   return (
     <header className="App-header">
       <Grid className="logo-container">
-        {props.logos.map((logo) => (
-          <img src={logo} className="App-logo" alt="logo" />
+        {props.logos.map((logo, idx) => (
+          <img
+            src={logo}
+            key={'image-' + idx}
```
y-lakhdar, line 21: Each item in a list should have a key

packages/cra-template/template/public/manifest.json
```diff
@@ -0,0 +1,15 @@
...
+      "src": "favicon.ico",
+      "sizes": "64x64 32x32 24x24 16x16",
+      "type": "image/x-icon"
+    }
+  ],
+  "start_url": ".",
+  "display": "standalone",
+  "theme_color": "#000000",
+  "background_color": "#ffffff"
+}
```
y-lakhdar, line 15: To fix the following error
```
Manifest: Line: 1, column: 1, Syntax error.
```

packages/cra-template/template/src/Components/Pager.tsx
louis-bompart: My React is rusty but... process.nextTick ? in a browser ?

louis-bompart (commented):
Really confused by the `process.nextTick`.
Other changes looks a-ok to me.

packages/cra-template/template/src/Components/Pager.tsx
y-lakhdar: Oh you're right. I'll switch to setTimeout(0).

louis-bompart: > Well, it is supported but I'll switch to setTimeout(0).

But... How 😕 When you break on this line, what is going on 🤯 what is process?

y-lakhdar: It's an implementation of process for browsers
https://www.npmjs.com/package/process

louis-bompart: Mh why not then, curious where does it come from?

louis-bompart (approved):

PR #387, opened 2021-07-28
feat(snapshot): add `--snapshotId` flag to `push` and `preview` commands

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-514

## Proposed changes

Pushing or previewing a large snapshot leads to a timeout with no possible action.
The idea is to add a `--snapshotId` to both `org:config:push` and `org:config:preview` commands in case the manipulated snapshot leads to a `SnapshotOperationTimeoutError` error.

If that is the case, the user will still be able to preview or push his snapshot once idle.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests: Pushed and preview a large snapshot causing a `SnapshotOperationTimeoutError` error

-----
CDX-514


packages/cli/src/commands/org/config/push.ts
```diff
@@ -47,6 +49,12 @@ export default class Push extends Command {
       default: false,
       required: false,
     }),
+    snapshotId: flags.string({
+      char: 's',
+      description:
+        'The unique identifier of the snapshot to push. If not specified, a new snapshot will be created from your local project. You can list available snapshot in your organization with org:config:list',
```
louis-bompart, line 55:
```suggestion
        'The unique identifier of the snapshot to push. If not specified, a new snapshot will be created from your local project. You can list available snapshots in your organization with org:config:list',
```

louis-bompart: I thought snapshot were not updatable 🤔 
What does pushing to a specific snapshotId does?

louis-bompart (commented):
not sure about push

packages/cli/src/commands/org/config/push.ts
y-lakhdar: I am not sure about this one. That's why I wanted your opinion on it.
Would it make sense to have a `--snapshotId` flag on all commands expect for the push?
But like you said, I don't see a lot of use cases expect for someone who wants to apply a snapshot that has been created in the Admin.

olamothe: I mean... if there's no use case that we see where someone would need to do that (and that we don't even know what happens when you do it), I would certainly not add the flag for push :D 

> Would it make sense to have a --snapshotId flag on all commands expect for the push?

I think that'd be fine, yes

packages/cli/src/commands/org/config/push.ts
```diff
@@ -47,6 +49,12 @@ export default class Push extends Command {
       default: false,
       required: false,
     }),
+    snapshotId: flags.string({
```
olamothe, line 52: As said in other comment, I guess we should not have this for `push`.

olamothe (approved):

PR #390, opened 2021-07-29
test(e2e): waitForNavigation will logging-in

louis-bompart:
## Proposed changes

Hypothesis: We try to click on the button while the JavaScript ain't fully loaded yet.
Proposed fix: Wait for everything to load (hopefully)

## Testing

FT/CI

-----
CDX-488


y-lakhdar (approved):
That would explain the intermittent fails

PR #394, opened 2021-07-30
fix(vue): support project creation with Node 16 and pnpm

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-520

## Proposed changes

Creating a VueJs project with Node 16 and [pnpm](https://pnpm.io/) causes the following error
image

CDX-520


packages/cli/src/commands/ui/create/vue.ts
```diff
@@ -143,7 +143,7 @@ export default class Vue extends Command {
         //   user: providerUsername,
         // },
       },
-      cssPreprocessor: 'node-sass',
+      cssPreprocessor: 'sass',
```
y-lakhdar, line 146: Node Sass causes some errors on the pnpm installation.
Node Sass is deprecated anyway.

packages/vue-cli-plugin-typescript/generator/index.js
```diff
@@ -12,11 +12,15 @@ module.exports = (api, options, rootOptions) => {
       '@coveo/headless': '*',
       '@coveo/search-token-server': '*',
       buefy: '^0.9.4',
+      bulma: '^0.9.3',
```
y-lakhdar, line 15: Even if [Buefy](https://buefy.org/) depends on [Bulma](https://bulma.io/), it has to be specified in the dependencies. Otherwise, the module is not found (when installed with pnpm).

packages/vue-cli-plugin-typescript/generator/index.js
```diff
@@ -12,11 +12,15 @@ module.exports = (api, options, rootOptions) => {
...
       concurrently: '^5.3.0',
       dotenv: '^9.0.2',
       'get-port': '^5.1.1',
       'fs-extra': '^10.0.0',
     },
+    devDependencies: {
+      tslib: '^2.3.0',
```
y-lakhdar, line 22: Same thing as above.
tslib should be added by [@vue/cli](https://cli.vuejs.org/), when using a typescript preset.

packages/vue-cli-plugin-typescript/generator/template/scripts/setup-server.js
```diff
@@ -27,6 +27,7 @@ function copySearchTokenServerToRoot() {
     resolve('server'),
     {
       filter: (src, dest) => !isNodeModule(dest) && !isEnvFile(dest),
+      dereference: true,
```
y-lakhdar, line 30: Because pnpm does some kind of symlink magic

louis-bompart (approved):

PR #398, opened 2021-08-02
use async-retry to not handle/redo the poll-timing logic

louis-bompart:
imho, this is a solved problem in the field, and we'd reinvent the wheel by not using a package to do it (it would be kinda an anti-pattern of how NPM works too)


packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -138,45 +139,45 @@ export class Snapshot {
     });
   }
 
-  public async waitUntilDone(
-    operationToWaitFor?: ResourceSnapshotsReportType | null,
-    options: WaitUntilDoneOptions = {},
-    iteratee = (_report: ResourceSnapshotsReportModel) => {}
+  public waitUntilDone(
+    operationToWaitFor: ResourceSnapshotsReportType | null,
+    options: Partial<WaitUntilDoneOptions> = {},
```
louis-bompart, line 144: Given the spread l145, we can provide a `Partial`. This allow the caller to only set the options he want

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -138,45 +139,45 @@ export class Snapshot {
...
-        timeout = setTimeout(() => {
-          const err = new SnapshotOperationTimeoutError(this);
-          clearInterval(interval);
-          reject(err);
-        }, toMilliseconds(opts.wait));
+
+    return retry(
+      this.waitUntilDoneRetryFunction(operationToWaitFor, onRetryCb),
+      {
+        retries: Infinity,
```
louis-bompart, line 153: We don't care about the number of attempt, so `Infinity`

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -138,45 +139,45 @@ export class Snapshot {
...
+        minTimeout: toMilliseconds(opts.waitInterval),
+        maxTimeout: toMilliseconds(opts.waitInterval),
```
louis-bompart, line 155: We want to retry every `opts.waitInterval` seconds.
Setting both limits of the Timeout ensure that this
`Math.min(random * minTimeout * Math.pow(factor, attempt), maxTimeout)` (see [node-retry](https://github.com/tim-kos/node-retry)) will always be eq to `opts.waitInterval`
given that: 
```
random=1 // i.e. no random
factor=2 // default value
attempt>=0 // 
```
 => Math.pow(factor,attempt)>=1
 => lower bound of the Math.min>=opts.waitInterval.
 => Math.min === opts.waitInterval

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -138,45 +139,45 @@ export class Snapshot {
...
-          clearInterval(interval);
-          resolve();
-        }
-      }, toMilliseconds(opts.waitInterval));
-    });
+      if (isOngoing || !isExpectedOperation) {
+        throw new SnapshotOperationTimeoutError(this);
+      }
+      return;
+    }).bind(this);
```
louis-bompart, line 181: extracted the function in a factory to keep things flat and simple tor ead

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -138,45 +139,45 @@ export class Snapshot {
...
+      onRetryCb(this.latestReport);
 
-        if (!isOngoing && isExpectedOperation) {
-          clearTimeout(timeout);
-          clearInterval(interval);
-          resolve();
-        }
-      }, toMilliseconds(opts.waitInterval));
-    });
+      if (isOngoing || !isExpectedOperation) {
```
louis-bompart, line 177: Reversed the condition using De Morgan Law to keep it simple. No return needed in the end

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -138,45 +139,45 @@ export class Snapshot {
...
-
-    await new Promise<void>((resolve, reject) => {
-      if (opts.wait > 0) {
-        timeout = setTimeout(() => {
-          const err = new SnapshotOperationTimeoutError(this);
-          clearInterval(interval);
-          reject(err);
-        }, toMilliseconds(opts.wait));
+
+    return retry(
```
louis-bompart, line 150: We can return the retry straight away and thus remove the async from waitUntilDone and just use the promise created by `async-retry`

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: Are you sure it supports Infinity? And that you don't have to use `Number.MAX_SAFE_INTEGER`

y-lakhdar (approved):
Clean

y-lakhdar (approved):

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -138,45 +139,44 @@ export class Snapshot {
...
-          clearTimeout(timeout);
-          clearInterval(interval);
-          resolve();
-        }
-      }, toMilliseconds(opts.waitInterval));
-    });
+      if (isOngoing || !isExpectedOperation) {
+        throw new SnapshotOperationTimeoutError(this);
+      }
+    }).bind(this);
```
y-lakhdar, line 180: why bind(this)

y-lakhdar (approved):

PR #404, opened 2021-08-04
fix: `org:config:monitor` doesn't show any feedback

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-526

## Proposed changes
The cli.action was not showing because it was at the wrong place.
image

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:
see image

-----
CDX-526

### DIFF
--- a/packages/cli/src/commands/org/config/monitor.ts
+++ b/packages/cli/src/commands/org/config/monitor.ts
@@ -45,9 +45,9 @@ export default class Monitor extends Command {
 
   @Preconditions(IsAuthenticated())
   public async run() {
-    const snapshot = await this.getSnapshot();
+    this.printHeader();
 
-    this.printHeader(snapshot.id);
+    const snapshot = await this.getSnapshot();
 
     await this.monitorSnapshot(snapshot);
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/org/config/monitor.ts
```diff
@@ -76,11 +76,14 @@ export default class Monitor extends Command {
...
-  private printHeader(snapshotId: string) {
+  private printHeader() {
+    const {args} = this.parse(Monitor);
+    const snapshotId = args.snapshotId;
     const header = ReportViewerStyles.header(
-      `\nMonitoring snapshot ${snapshotId}:`
+      `Monitoring snapshot ${snapshotId}`
     );
-    cli.log(header);
+    cli.log('');
```
louis-bompart, line 85: Why the log '' ?

y-lakhdar: Just to create some space between the command and the message.

louis-bompart (approved):

PR #405, opened 2021-08-04
feat(snapshot): enforce json format.

louis-bompart:
## Proposed changes

To ensures consistent file formating, let's enforce the format when pulling, previewing etc snapshots.

## Testing

- [X] Unit Tests: done
- [x] Manual Tests: Checked Expanded Preview, there are no more space changes in the commits

-----
CDX-491


packages/cli/src/lib/project/project.ts
```diff
@@ -1,13 +1,21 @@
...
 import {cli} from 'cli-ux';
 import archiver from 'archiver';
 import {InvalidProjectError} from '../errors';
 import extract from 'extract-zip';
 import {DotFolder, DotFolderConfig} from './dotFolder';
+import {readJsonSync, writeJsonSync, WriteOptions} from 'fs-extra';
 
 export class Project {
   private static readonly resourceFolderName = 'resources';
+  public static readonly jsonFormat: WriteOptions = {spaces: '\t'};
```
y-lakhdar, line 18: image

y-lakhdar: I don't think this will be enough to enforce a JSON format

y-lakhdar (commented):
Instead of replacing spaces with tabs, can't we just lint the file after pulling it?
It would be nice to add a linter in the Coveo project. 

I always format my JSON files (with the VScode extension) when I manipulate them. I'm sure I am not the only one who does that.

Here is a JSON file returned by the API and the formatted version (using the VScode JSON formatter)
image

packages/cli/src/lib/project/project.ts
louis-bompart: Why so, can you elaborate?

y-lakhdar: As shown in the screenshot above, the JSONs (the one from the project and the other coming from the API) could have more than just space and tabs differences.

louis-bompart: Hm 👍 to keep a trace of what we discuss, I think it's not a problem 'cause what is the true goal here is to enforce formating in the expanded preview, which should be done here I reckon

y-lakhdar (approved):

PR #409, opened 2021-08-05
fix(cli): handle npx 7

louis-bompart:
## Proposed changes

Starting with NPX 7, when you run an `npx` command that requires an installation, a prompt is given to the user to confirm the installation.
In our use-case, this occurs with `react` and the `create-react-app`. However, this should not be a concern for the user of the Coveo CLI, so we'll use the `--yes` flag with NPX iif its version is equal or greater than 7.

## Testing

- [x] Unit Tests: 💯 
- [x] Manual Tests

-----
CDX-530


y-lakhdar (approved):
Nice

PR #410, opened 2021-08-06
feat(auth): rework `auth:login` to work with `coveo://`

louis-bompart:
Still rough around the edges, but working ™️


louis-bompart:
Will require more work to support other usage for the custom scheme

PR #421, opened 2021-08-10
fix(vue): fix broken pager buttons

y-lakhdar:
## Proposed changes

The state was not correctly refreshed since the `total` value was only assigned on the component initialization and not on the subscribe method

## Before Merging
- [x] Merge #422

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/vue-cli-plugin-typescript/generator/template/src/components/Pager.vue
```diff
@@ -30,8 +29,6 @@ export default Vue.extend({
     return {
       pager: pager,
       state: {...pager.state, total: 0},
-      totalCount: (this.$root.$data.$engine as SearchEngine).state.search
```
y-lakhdar, line 33: `totalCount` was always only assigned once. Since the parameter is transmitted by value and not by reference, `totalCount` remains unchanged on future the state refreshes.

packages/vue-cli-plugin-typescript/generator/template/src/components/Pager.vue
```diff
@@ -43,7 +40,8 @@ export default Vue.extend({
     this.pager.subscribe(() => {
       this.state = {
         ...this.pager.state,
-        total: this.totalCount,
```
y-lakhdar, line 46: `this.totalCount` was always equal to zero

PR #431, opened 2021-08-16
feat(snapshot): enhance snapshot error messages

y-lakhdar:
## Proposed changes

* Clean snapshot error messages and make sure `cli.action` always displays the error message (instead of `... done`).
* Update the `#handleSnapshotError` method to make snapshot errors more consistent.

### Push Example
**Before** 
image

**After**
image

### Monitor Example
**Before**
image

**After**
image

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:

### DIFF
--- a/packages/cli/src/lib/errors/snapshotErrors.ts
+++ b/packages/cli/src/lib/errors/snapshotErrors.ts
@@ -1,13 +1,50 @@
 import {blueBright} from 'chalk';
+import {cli} from 'cli-ux';
 import dedent from 'ts-dedent';
 import {Configuration} from '../config/config';
 import {Snapshot} from '../snapshot/snapshot';
 import {SnapshotUrlBuilder} from '../snapshot/snapshotUrlBuilder';
 
-export class SnapshotOperationTimeoutError extends Error {
+// type severityLevels = 'info' | 'warn' | 'error';
+
+enum severityLevels {
+  'info' = 'info',
+  'warn' = 'warn',
+  'error' = 'error',
+}
+
+interface IDetailedReportable extends SnapshotError {
+  snapshot: Snapshot;
+  level: severityLevels;
+  projectPath?: string;
+}
+
+function trySavingDetailedReport(error: IDetailedReportable) {
+  if (error.projectPath) {
+    const reportPath = error.snapshot.saveDetailedReport(error.projectPath);
+    error.message += dedent`\n\n
+          Detailed report saved at ${reportPath}`;
+  }
+}
+
+export class SnapshotError extends Error {
+  public constructor(public level: severityLevels) {
+    super();
+  }
+
+  public print() {
+    cli.log();
+    cli[this.level]('\n' + this.message);
+  }
+}
+
+export class SnapshotOperationTimeoutError
+  extends SnapshotError
+  implements IDetailedReportable
+{
   public name = 'Snapshot Operation Timeout Error';
   public constructor(public snapshot: Snapshot) {
-    super();
+    super(severityLevels.info);
     this.message = dedent`${
       snapshot.latestReport.type
     } operation is taking a long time to complete.
@@ -17,14 +54,17 @@ export class SnapshotOperationTimeoutError extends Error {
   }
 }
 
-export class SnapshotSynchronizationError extends Error {
+export class SnapshotSynchronizationError
+  extends SnapshotError
+  implements IDetailedReportable
+{
   public name = 'Snapshot Synchronization Error';
   public constructor(
     public snapshot: Snapshot,
     public cfg: Configuration,
     public projectPath?: string
   ) {
-    super();
+    super(severityLevels.warn);
     const urlBuilder = new SnapshotUrlBuilder(cfg);
     const synchronizationPlanUrl = urlBuilder.getSynchronizationPage(snapshot);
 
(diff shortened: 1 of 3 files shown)


packages/cli/src/lib/errors/snapshotErrors.ts
```diff
@@ -26,24 +32,35 @@ export class SnapshotSynchronizationError extends Error {
...
-      Detailed report saved at ${pathToReport}.
-
       You can also use this link to view the snapshot in the Coveo Admin Console
       ${snapshotUrl}`;
+
+    if (projectPath) {
+      const reportPath = snapshot.saveDetailedReport(projectPath);
+      this.message += dedent`\n\n
+          Detailed report saved at ${reportPath}`;
+    }
```
louis-bompart, line 64: You could refactor this bit, see:
https://www.typescriptlang.org/play?#code/C4TwDgpgBAygdgQzAZwBYHthQLxQXEAbgChRIoBhdOAMwEsBzAVwCcFg7qc8CTi64wCCxoIAxtACSAEQjAEdADYQAJgCUIYdC3kAjZVAgAPIXBXIoAURYttUAN7EoUZIhQZgALlhu0mADROUGC2AFYQYsAACuyoAPzeyMAsAgzEAL7ExDRMcJGccFDJIACCKiqpsvJKqhpaOgAUwrYs3jJyCsrqmtp6ygCUDkF0NFBNNtoAdCHo4ZExwKiDjs7OYtRJUCw9Oguo3M1Trkh+wJPICABuEFWdtTvA4y3TYRHRsf0kq4YTLJMAthBkBcGNAANS4FSqCCCAAGAB04Iigt9vrcaiotg8XFdVHgsAASezbervRbpWFfKCZTLEMSKBDAnwnDwAcRhwjoYmsLUMJhh5isvygdH+YGUgMEFna1S6dV6CH00BWwSY+i5UEQgO4AHJ4CzMEKWjqqWA1YoNes4EkWExItoGijVeqxC5fB5vPr3AEnWaXVAxDQGN4qLRGKx2AVAt8-RbXTM5mT4olkqkgssncgmJAWA1Pk6rZsSb1uMdvWcGUIkvKdFS1hssKxFAAhJhKKEsbhwCAAd2Z5YAqiwW23FB2GoGGPnvoWsGXTkPFNwm6328JJqDgF7TjFQQ15x5807FnRkACgSDoJDoXDJHBLggLZiD4aALRQInFnSTbZZxTAKgoXSSYnWcABNdAmADfA8EUZB0CgJhkGgE8LAtOAAGsigQy46F7IpUGgF8sAEAjoCoa4ELKf5SNDeDlFAj97GIxcKT4b5ijKCo4AYdE5QeBpUP6IIaSAA

packages/cli/src/lib/snapshot/snapshotCommon.ts
```diff
@@ -69,24 +69,28 @@ export function cleanupProject(projectPath: string) {
...
   if (err instanceof SnapshotOperationTimeoutError) {
-    cli.action.stop('Incomplete');
-    cli.log(err.message);
+    cli.log(errorMessage);
   } else if (err instanceof SnapshotSynchronizationError) {
-    cli.warn(err.message);
+    cli.warn(errorMessage);
   } else if (err instanceof SnapshotGenericError) {
-    cli.error(err.message);
+    cli.error(errorMessage);
```
louis-bompart, line 93: For now, OK, if we add one more kind of Error, we should refactor the Error to rely on a severity system to avoid code repetition ;)

packages/cli/src/lib/errors/snapshotErrors.ts
```diff
@@ -26,24 +32,35 @@ export class SnapshotSynchronizationError extends Error {
       Some conflicts were detected while comparing changes between the snapshot and the target organization.
       Click on the URL below to synchronize your snapshot with your organization before running another push command.
       ${synchronizationPlanUrl}`;
+
+    if (projectPath) {
+      const reportPath = snapshot.saveDetailedReport(projectPath);
+      this.message += dedent`\n\n
+        Detailed report saved at ${reportPath}`;
+    }
```
louis-bompart, line 40: See comment below for refactor

louis-bompart (changes requested):
Refactor possible ;)

packages/cli/src/lib/snapshot/snapshotCommon.ts
olamothe: +1, just a simple `extends Error` class with a Severity enum, should be straightforward to add.

olamothe (approved):

packages/cli/src/lib/snapshot/snapshotCommon.ts
y-lakhdar: Would it make sense to have something like that
```
  if (err instanceof SnapshotError) {
    err.print();
  } else {
    throw err;
  }
```

packages/cli/src/lib/errors/snapshotErrors.ts
```diff
@@ -1,13 +1,50 @@
...
 import {SnapshotUrlBuilder} from '../snapshot/snapshotUrlBuilder';
 
-export class SnapshotOperationTimeoutError extends Error {
+// type severityLevels = 'info' | 'warn' | 'error';
+
+enum severityLevels {
+  'info' = 'info',
+  'warn' = 'warn',
+  'error' = 'error',
+}
```
louis-bompart, line 14: https://github.com/microsoft/TypeScript/wiki/Coding-guidelines
Please follow the guideline when possible ;)

All does not makes sense, but for enum here it kinda does (tho flavor, y'know), especially because we followed them so far (for enums)

y-lakhdar: In this case, the enum values need to match the log levels available under the `cli` object. That's because we are doing
I think `type severityLevels = 'info' | 'warn' 'error'` would be more appropriate than enum here.

What do you think?

louis-bompart: But... You can just do:
```ts
enum SeverityLevel {
  Info = 'info',
  Warm = 'warn',
  Error = 'error',
}
```
No?..

y-lakhdar: Oh I see. I thought you wanted the values to be in pascal case... my bad

PR #434, opened 2021-08-17
feat(cli): unbundle 3rd-party CLIs

louis-bompart:
## Proposed changes

Currently, a large part of the footprint of the CLI is caused by `@vue/cli` or `@angular/cli`, or `create-react-app`.
Unbundling them solves the issue and also ensures that the user will have a smooth ride moving forward when using either of them.

Removed some unused branching in the preconditions factories and add support for auto-fix in precondition

## Breaking changes

This is arguable. This could be considered a breaking change on `ui:create:angular` because we now have a new user interaction branching. However, I do not consider CI/CD usage of this command to be a use case we want to support. Therefore, I'll consider the breaking aspect on iff the input of the command changed (no), or if the output of the command changed (no)

## Testing

- [X] Unit Tests: 💯 unit-tested
- [X] Functional Tests: If the current tests pass (In the CI I trust 🎲), then it's covered, given that `@angular/cli` is the tricky bit with the global install, and is not installed in the setup phase (which mean it'll therefore be installed by the preconditions and the Terminal manipulation.
- [X] Manual Tests: Ran the three commands

-----
CDX-478

### DIFF
--- a/packages/cli/src/commands/ui/create/vue.ts
+++ b/packages/cli/src/commands/ui/create/vue.ts
@@ -14,9 +14,11 @@ import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
 import {platformUrl} from '../../../lib/platform/environment';
 import {spawnProcess} from '../../../lib/utils/process';
 import {getPackageVersion} from '../../../lib/utils/misc';
+import {appendCmdIfWindows} from '../../../lib/utils/os';
 
 export default class Vue extends Command {
   public static templateName = '@coveo/vue-cli-plugin-typescript';
+  public static cliPackage: '@vue/cli';
 
   /**
    * @see https://cli.vuejs.org/guide/installation.html for current requirements.
@@ -160,8 +162,11 @@ export default class Vue extends Command {
   }
 
   private runVueCliCommand(args: string[], options = {}) {
-    const executable = require.resolve('@vue/cli/bin/vue.js');
-    return spawnProcess('node', [executable, ...args], options);
+    return spawnProcess(
+      appendCmdIfWindows`npx`,
+      [`${Vue.cliPackage}@${getPackageVersion(Vue.cliPackage)}`, ...args],
+      options
+    );
   }
 
   private get configuration() {
--- a/packages/cli/src/commands/ui/create/vue.ts
+++ b/packages/cli/src/commands/ui/create/vue.ts
@@ -18,7 +18,7 @@ import {appendCmdIfWindows} from '../../../lib/utils/os';
 
 export default class Vue extends Command {
   public static templateName = '@coveo/vue-cli-plugin-typescript';
-  public static cliPackage: '@vue/cli';
+  public static cliPackage = '@vue/cli';
 
   /**
    * @see https://cli.vuejs.org/guide/installation.html for current requirements.
--- a/packages/cli/src/commands/ui/create/vue.ts
+++ b/packages/cli/src/commands/ui/create/vue.ts
@@ -9,6 +9,7 @@ import {
   Preconditions,
   IsAuthenticated,
   IsNodeVersionInRange,
+  IsNpxInstalled,
 } from '../../../lib/decorators/preconditions';
 import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
 import {platformUrl} from '../../../lib/platform/environment';
@@ -61,7 +62,8 @@ export default class Vue extends Command {
 
   @Preconditions(
     IsAuthenticated(),
-    IsNodeVersionInRange(Vue.requiredNodeVersion)
+    IsNodeVersionInRange(Vue.requiredNodeVersion),
+    IsNpxInstalled()
   )
   public async run() {
     const {args, flags} = this.parse(Vue);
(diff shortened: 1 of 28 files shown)


louis-bompart:
- [ ] #436

packages/cli/src/lib/decorators/preconditions/binPreconditionsFactory.ts
```diff
@@ -120,8 +133,5 @@ function warnHowToInstallBin(
 }
 
 function isBinFileMissing(output: SpawnProcessOutput) {
-  return (
-    output.exitCode !== null &&
-    Math.abs(output.exitCode) === constants.errno.ENOENT
-  );
+  return /^spawn \S* ENOENT$/.test(output.stderr);
```
y-lakhdar, line 136: Why? `constants.errno.ENOENT` was not working?

packages/cli/src/commands/ui/create/vue.ts
```diff
@@ -160,8 +162,11 @@ export default class Vue extends Command {
   }
 
   private runVueCliCommand(args: string[], options = {}) {
-    const executable = require.resolve('@vue/cli/bin/vue.js');
-    return spawnProcess('node', [executable, ...args], options);
+    return spawnProcess(
+      appendCmdIfWindows`npx`,
```
y-lakhdar, line 166: Do you need to add the `IsNpxInstalled` precondition in this case?

y-lakhdar (approved):
shia

packages/cli/src/lib/decorators/preconditions/binPreconditionsFactory.ts
olamothe, line 136: Why this change versus a check on the exit code ?

packages/cli/src/lib/decorators/preconditions/ng.ts
```diff
@@ -0,0 +1,43 @@
...
+    dedent`
+        You need to have the Angular-CLI installed on your computer before proceeding.
+        Do you want to install it now? (y/n)`
+  );
+  if (!shouldInstall) {
+    return false;
+  }
+  const version = getPackageVersion(angularCliPackage);
+  target.log(`Running npm install --global @angular/cli@${version}.`);
+  const output = await spawnProcess(appendCmdIfWindows`npm`, [
```
olamothe, line 21: It's starting to get a bit messy here, since it means we have a precondition that depends on one another (that npm is installed and new enough that you can actually install the angular cli with it).

Could we just print an info message about how to install it ?

It should be pretty easy for them to do it anyway, plus they have more control on the installation process, I guess...

Do you think there's a lot of value by us trying to do the installation "automatically" ?

olamothe (commented):
G

packages/cli/src/lib/utils/misc.ts
```diff
@@ -8,7 +8,10 @@ export function getPackageVersion(packageName: string) {
     'package.json'
   );
   const pkg = require(pathToPackageJson);
-  const dep = pkg.dependencies[packageName] || pkg.devDependencies[packageName];
+  const dep =
+    pkg.peerDependencies[packageName] ||
+    pkg.devDependencies[packageName] ||
+    pkg.peerDependencies[packageName];
```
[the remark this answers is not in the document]
louis-bompart: Good catch, should be dependencies, then fallback to dev, then fallback to peer

packages/cli/src/lib/decorators/preconditions/binPreconditionsFactory.ts
```diff
@@ -1,31 +1,41 @@
...
 
 export interface IBinPreconditionsOptions {
   params?: Array<string>;
-  prettyName?: string;
+  prettyName: string;
   installLink?: string;
   howToInstallBinText?: string;
 }
 
+export interface IBinPreconditionsOptionsWithAutoFix
```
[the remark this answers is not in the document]
louis-bompart: Ahahaah nice!
- [ ] rename interface

louis-bompart:
https://github.com/coveo/cli/pull/434#discussion_r690690043
https://github.com/coveo/cli/pull/434#discussion_r690681565

This is a funny story. I'll start with the short version: Windows. The long one now:

Node.JS rely on libuv/libuv for IO operations (which is fine!). However, libuv/libuv also does that:
https://github.com/libuv/libuv/blob/v1.x/include/uv/errno.h#L245-L249

What this does, is 'if ENOENT is defined, then I set UV__ENOENT to its value, else, I set it to -4058'
POSIX system has an ENOENT defined at a known and easy value: 2 (KISS).
Windows, you guessed it, has not.

So, when a file is missing on Windows, we got a -4058 for errorCode, and 2 for POSIX.
But the fun does not stop there, remember, 'If ENOENT is defined, then I set UV__ENOENT to its value'

Well, Windows user can define ENOENT (or so I read) if they so desire :feelsgood: . Which that the `errno` is worth nothing and can't be relied upon.

I'll double-check something tho on that given that 'code' is consistent it seems:
image

- [ ] Check if we could use `code`

packages/cli/src/lib/decorators/preconditions/ng.ts
louis-bompart: Honestly, not much value. I tried to match what was possible in `vue` and `react` but that's all. I do not mind axing the auto-update tbh

louis-bompart: - [ ] Remove autofix

olamothe (approved):

PR #441, opened 2021-08-23
fix(cli): use proper region suffix

louis-bompart:
## Proposed changes

Use `us` `eu` `au` instead of the AWS abv.

## Breaking changes

Existing user settings will ungracefully break with only this Jira.
I scoped the user setting in another Jira to keep PR tidy.

## Testing

- [X] Unit Tests: adapted
- [X] Functional Tests: existing should work, other regions added for future investigation on 'testing environment'.
- [X] Manual test: yep.

-----
CDX-544

### DIFF
--- a/packages/cli/jest.config.js
+++ b/packages/cli/jest.config.js
@@ -1,8 +1,15 @@
+const {defaults} = require('jest-config');
+
 module.exports = {
   preset: 'ts-jest',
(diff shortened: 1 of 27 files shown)


y-lakhdar (approved):

olamothe (approved):
💯 Great PR !

Simplifies a lot of test repetition

PR #442, opened 2021-08-23
feat(cli): add version on config.json

louis-bompart:
## Proposed changes

So, we reach the point where we need to invalidate the graceful config of existing users.
To ensure we can do this easily and without an entangling mess of if-else, I propose that we add a SemVer field in the config.json. 

The SemVer might be a bit overkill, but should give us some flexibility to handle 'old' config (i.e. config with a old schemaVersion) and 'future' config, in the case the user fallback to a previous version of the CLI.

## Breaking changes

All users config will be wiped.

## Testing

- [X] Unit Tests: Yep
- [x] Functional Tests: Nay
- [X] Manual Tests:
  - Run `config:set` or any command that relies on the command from the src. My settings were wiped.
  - Auth, and tada it works I have a version in my config. Yay 🎉 

-----
CDX-546


louis-bompart:
- [ ] Merge #441

packages/cli/src/lib/config/config.ts
```diff
@@ -22,16 +24,10 @@ export interface Configuration {
...
 
-export const DefaultConfig: Configuration = {
-  environment: DEFAULT_ENVIRONMENT,
-  region: DEFAULT_REGION,
-  organization: '',
-  analyticsEnabled: undefined,
-  accessToken: undefined,
-  anonymous: undefined,
-};
+class IncompatibleConfiguration extends Error {}
```
y-lakhdar, line 27: Maybe put this custom error in a different file?

y-lakhdar (approved):

packages/cli/src/lib/config/config.ts
```diff
@@ -44,9 +40,18 @@ export class Config {
...
       }
+      if (!this.isSettingVersionInRange(content)) {
+        throw new IncompatibleConfiguration();
+      }
       return content;
     } catch (e) {
-      this.error(`Error while reading configuration at ${this.configPath}`);
+      if (e instanceof IncompatibleConfiguration) {
+        this.error(
+          `The configuration at ${this.configPath} is not compatible with this version of the CLI`
```
olamothe, line 50: Maybe we could explain "why" its not compatible ? The `IncompatibleConfiguration` error could contain a message that explains it.

Something like "no version found", "mismatch on major version", "incompatible format", etc.

olamothe (approved):

packages/cli/src/lib/config/config.ts
```diff
@@ -57,6 +62,13 @@ export class Config {
     }
   }
 
+  private isSettingVersionInRange(content: Partial<Configuration>) {
+    if (!content.version) {
+      return false;
+    }
+    return satisfies(content.version, `^${Config.CurrentSchemaVersion}`);
```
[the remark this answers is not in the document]
louis-bompart: image
false :)

packages/cli/src/lib/config/config.ts
louis-bompart: - [ ] Better logging

packages/cli/src/lib/config/config.ts
louis-bompart: Because at first it was just a instanceof checker I was not convinced, but because I'll add a message, yeah, could make sense

PR #444, opened 2021-08-24
feat(react): commit /server directory on project creation **WIP**

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-545

## Proposed changes

Running `ui:create:react` performs an initial git commit, but does not include the `/server` directory.
Unfortunately, we do not control how and when the initial commit is being done. `create-react-app` is handling the initial commit and does not expose any flag to prevent that.

The solution is to perform a second commit to include the server directory
 
## Testing

- [x] Unit Tests:

- [x] Functionnal Tests:

- [ ] Manual Tests:

 

-----
CDX-545

### DIFF
--- a/packages/cli-e2e/__tests__/react.specs.ts
+++ b/packages/cli-e2e/__tests__/react.specs.ts
@@ -15,7 +15,7 @@ import {
   restoreEnvironmentFile,
 } from '../utils/file';
 import {BrowserConsoleInterceptor} from '../utils/browserConsoleInterceptor';
-import {commitProject, undoCommit} from '../utils/git';
+import {isDirectoryClean} from '../utils/git';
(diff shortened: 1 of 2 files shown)


packages/cli-e2e/__tests__/react.specs.ts
```diff
@@ -229,17 +224,17 @@ describe('ui:create:react', () => {
       });
     });
 
-    it('should be commited without lint-stage errors', async () => {
-      const eslintErrorSpy = jest.fn();
+    it('should have a clean working directory', async () => {
+      const gitDirtyWorkingTreeSpy = jest.fn();
 
-      commitProject(
```
y-lakhdar, line 235: There is no need to manually commit the /server directory since it is being done in the project creation.

PR #452, opened 2021-08-31
ci(pr): ensure pr name are semantic

louis-bompart:
## Proposed changes

Use https://github.com/github-actions/is-pr-title-semantic to ensure the title of a PR is semantic.

Like how `@coveo/merge-bot` do, it'll comment directly on the PR when there's a problem (and the process will also fail to mark the checks as invalid)

If all's well, it'll exit with code 0 (i.e. check validated) and delete its comment.

The linter runs on a selection of pull request event that should do the trick :tm: (I try to no go too trigger happy on the event to no spam GH Action)

## Testing

- [x] Manual Tests: on another repo

-----
CDX-561

### DIFF
--- a/.github/workflows/pr-title-semantic-lint.yml
+++ b/.github/workflows/pr-title-semantic-lint.yml
@@ -1,0 +1,16 @@
+name: PrTitleSemanticLint
+on:
+  pull_request:
+    branches: [ main ]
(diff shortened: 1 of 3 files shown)


olamothe (approved):

louis-bompart:
- [ ] Add `commitlint.config.js`

PR #455, opened 2021-09-01
feat(snapshot): handle snapshot synchronization

y-lakhdar:
## Proposed changes

If synchronization is required, generate a sync plan

If the plan has everything matched up with a 💯 confidence, propose to the user to apply said plan.

Else, redirect the user to the snapshot preview page

## Screenshots
image

image

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:

-----
CDX-551

### DIFF
--- a/packages/cli/src/commands/org/config/preview.spec.ts
+++ b/packages/cli/src/commands/org/config/preview.spec.ts
@@ -115,9 +115,6 @@ const mockSnapshotFactoryReturningInvalidSnapshot = async () => {
 const mockSnapshotFacade = () => {
   mockedSnapshotFacade.prototype.tryAutomaticSynchronization =
     mockedTryAutomaticSynchronization;
-  // .mockImplementation(() => ({
-  //   tryAutomaticSynchronization: jest.fn(),
-  // }));
 };
 
 describe('org:config:preview', () => {
--- a/packages/cli/src/commands/org/config/preview.spec.ts
+++ b/packages/cli/src/commands/org/config/preview.spec.ts
@@ -6,6 +6,7 @@ jest.mock('../../../lib/platform/authenticatedClient');
 jest.mock('../../../lib/snapshot/snapshot');
 jest.mock('../../../lib/snapshot/snapshotFactory');
 jest.mock('../../../lib/project/project');
+jest.mock('../../../lib/snapshot/snapshotFacade');
 jest.mock('@oclif/errors');
 
 import {mocked} from 'ts-jest/utils';
@@ -16,7 +17,7 @@ import {cwd} from 'process';
 import {Config} from '../../../lib/config/config';
 import {SnapshotFactory} from '../../../lib/snapshot/snapshotFactory';
 import {Snapshot} from '../../../lib/snapshot/snapshot';
-import {warn, error} from '@oclif/errors';
+import {error} from '@oclif/errors';
 import {SnapshotReporter} from '../../../lib/snapshot/snapshotReporter';
 import {ResourceSnapshotsReportType} from '@coveord/platform-client';
 import {
@@ -25,11 +26,11 @@ import {
 } from '../../../__stub__/resourceSnapshotsReportModel';
 import {Command} from '@oclif/command';
 import {IsGitInstalled} from '../../../lib/decorators/preconditions';
+import {SnapshotFacade} from '../../../lib/snapshot/snapshotFacade';
 
 const mockedSnapshotFactory = mocked(SnapshotFactory, true);
 const mockedConfig = mocked(Config);
 const mockedProject = mocked(Project);
-const mockedWarn = mocked(warn);
 const mockedError = mocked(error);
 const mockedConfigGet = jest.fn();
 const mockedDeleteTemporaryZipFile = jest.fn();
@@ -39,7 +40,12 @@ const mockedRequiresSynchronization = jest.fn();
 const mockedValidateSnapshot = jest.fn();
 const mockedPreviewSnapshot = jest.fn();
 const mockedLastReport = jest.fn();
+const mockedCreateSynchronizationPlan = jest.fn();
+const mockedApplySynchronizationPlan = jest.fn();
+const mockedTryAutomaticSynchronization = jest.fn();
 const mockedIsGitInstalled = mocked(IsGitInstalled, true);
+const mockedSnapshotFacade = mocked(SnapshotFacade, true);
+
 const mockProject = () => {
   mockedProject.mockImplementation(
     () =>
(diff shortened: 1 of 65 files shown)


packages/cli/src/lib/snapshot/synchronization/synchronizationPlan.ts
```diff
@@ -0,0 +1,29 @@
...
+  ) {}
+
+  public containsUnambiguousMatches(): boolean {
+    const synchronizationOperations =
+      this.model.resourceSynchronizationOperations;
+    if (!synchronizationOperations) {
+      return false;
+    }
+
+    for (const [_, operations] of Object.entries(synchronizationOperations)) {
```
louis-bompart, line 15:
```suggestion
    for (const operations of Object.values(synchronizationOperations)) {
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/values

y-lakhdar: yeah, I kept it like this because I intended to use the keys in the synchronization preview... but I can change it later

packages/cli/src/lib/snapshot/synchronization/synchronizationPlan.ts
```diff
@@ -0,0 +1,29 @@
...
+      for (let i = 0; i < operations.length; i++) {
+        const matches = operations[i].matches;
+        for (let j = 0; j < matches.length; j++) {
+          const match = matches[j];
+          if (match.associationScore && match.associationScore < 1) {
+            return false;
+          }
+        }
+      }
+    }
```
louis-bompart, line 25:
```suggestion
    for (const synchronizationOperation of Object.values(synchronizationOperations)) {
      for (const operations of synchronizationOperation) {
        for (const matches of operation.matches) {
          return matches.some(match=> match.associationScore || 0 <1)
        }
      }
    }    
```

packages/cli/src/lib/snapshot/snapshotFacade.ts
```diff
@@ -0,0 +1,58 @@
...
+    const reporter = await this.snapshot.applySynchronizationPlan(
+      plan.model.id
+    );
+    const success = reporter.isSuccessReport();
+
+    if (!success) {
+      throw new SnapshotSynchronizationUnknownError(this.snapshot, this.cfg);
+    }
+
+    cli.action.stop(success ? green('✔') : red.bold('!'));
```
louis-bompart, line 56: Unreachable code: red.bold('!) I reckon.

y-lakhdar: haha good catch

packages/cli/src/lib/snapshot/reportPreviewer/reportPreviewer.ts
```diff
@@ -132,4 +133,8 @@ export class ReportViewer {
       );
     }
   }
+
+  private printNewLine() {
+    cli.log('');
+  }
```
louis-bompart, line 139: 👍

packages/cli/src/lib/snapshot/synchronization/synchronizationPlan.ts
y-lakhdar: clean 👍

packages/cli/src/commands/org/config/preview.spec.ts
```diff
@@ -104,6 +112,14 @@ const mockSnapshotFactoryReturningInvalidSnapshot = async () => {
   await mockSnapshotFactory();
 };
 
+const mockSnapshotFacade = () => {
+  mockedSnapshotFacade.prototype.tryAutomaticSynchronization =
+    mockedTryAutomaticSynchronization;
+  // .mockImplementation(() => ({
```
olamothe, line 118: Cleanup

packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -152,6 +199,17 @@ export class Snapshot {
...
 
+  private sortReportsByDate<T extends SnapshotReport>(report: T[]): T[] {
+    return report.sort((a, b) => b.updatedDate - a.updatedDate);
+  }
+
+  private hasBegunSynchronization(
+    _synchronizationReports?: ResourceSnapshotsSynchronizationReportModel[]
+  ): _synchronizationReports is ResourceSnapshotsSynchronizationReportModel[] {
+    const reports = this.model.synchronizationReports;
+    return Array.isArray(reports) && reports.length > 0;
```
olamothe, line 210: What else could `reports` be ?

 I'm guessing it's more of a "is not null or undefined"  check ? 

If so, just optional chaining should do the trick.

packages/cli/src/lib/snapshot/snapshotFacade.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  public async tryAutomaticSynchronization() {
+    const plan = await this.createSynchronizationPlan();
+    await this.waitForConfirmation();
+    await this.applySynchronizationPlan(plan);
+  }
+
+  private async waitForConfirmation() {
+    const canApplySynchronizationPlan = await cli.confirm(
+      'Synchronization plan matched all resources with great confidence, do you want to proceed? (y/n)'
```
olamothe, line 23: Maybe I'd loop in @jpmarceau  for the wording... 

I think I could ask myself, as a end user, what "great confidence" means... Like, is it necessary ? Could we just say "Synchronization plan matched all resources, do you want to proceed?

olamothe (approved):
Nice work !

packages/cli/src/lib/errors/snapshotErrors.ts
```diff
@@ -52,23 +52,75 @@ export class SnapshotOperationTimeoutError
...
+export class SnapshotNoSynchronizationReportFoundError
+  extends SnapshotError
+  implements IDetailedReportable
+{
+  public name = 'No Synchronization Report Found Error';
+  public constructor(public snapshot: Snapshot) {
+    super(SeverityLevel.Error);
+    this.message = dedent`
+    No synchronization report found for the snapshot ${this.snapshot.id}.
+    The snapshot should first be synchronized`;
```
jpmarceau, line 76:
```suggestion
    The snapshot should first be synchronized.`;
```

packages/cli/src/lib/snapshot/snapshotFacade.ts
jpmarceau: That's a good question. I don't completely remember how the migration api evaluates this, but Olivier's recommendation sounds good to me :)

packages/cli/src/lib/snapshot/snapshot.ts
y-lakhdar: Yeah it's only a undefined check

I tried but `reports?.length` cannot be compared to a number since it might be `undefined`
image

I think the shorter alternative would be something like 
```
const reports = this.model.synchronizationReports || [];
return reports.length > 0;
```

jpmarceau (approved):
👍

PR #460, opened 2021-09-07
feat(snapshot): add `--sync` flag to automatically apply sync plan (if 100% confidence)

y-lakhdar:
## Proposed changes

If the `--sync` flag is set to true and synchronization plan has only 100% matches, then the plan will automatically be applied.

## Screenshots
image

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/lib/snapshot/snapshotFacade.ts
```diff
@@ -12,9 +12,11 @@ import {SynchronizationPlan} from './synchronization/synchronizationPlan';
 export class SnapshotFacade {
   public constructor(private snapshot: Snapshot, private cfg: Configuration) {}
 
-  public async tryAutomaticSynchronization() {
+  public async tryAutomaticSynchronization(skipPrompt = false) {
```
olamothe, line 15: Nitpick, but I would just reverse that to prompt = true, so there would be no "double negative".

And in the code: 

```
if(prompt) {
   await this.waitForConfirmation();
}
[...]
```

olamothe (approved):

packages/cli/src/lib/flags/snapshotCommonFlags.ts
```diff
@@ -0,0 +1,23 @@
...
+    helpValue: 'seconds',
+    required: false,
+    description:
+      'The maximum number of seconds to wait before the commands exits with a timeout error. A value of zero means that the command will wait indefinitely.',
+  }),
+});
+
+export const sync = () => ({
+  sync: flags.boolean({
+    char: 'y',
```
louis-bompart, line 17: why `y`?

y-lakhdar: 's' is already used for `--snapshotId`.

louis-bompart (approved):

PR #461, opened 2021-09-07
feat(react): use transform instead of pty

louis-bompart:
## Proposed changes

Removed the burden of the environment and server initialization from the CLI and shift it to the React Template.
This follows more closely what we do in the other templates/schematics (or whatever you wanna name 'em).

The big boon in our case here is that because this is 'part' of the install, therefore, it's run when create-react-app actually invoke our template using `npm init`/`npm exec`

### Detailed list of changes:
 - Shift server & environment initialization from the CLI to the templates
 - Implemented cross-package-manager support for the commands.
 - Removed `node-pty`

## Testing

- [x] Unit Tests: rewrite of UT based on Vue's one (they both now behave almost identically)
- [x] Functional Tests: no new features, existing does the trick
- [x] Manual Tests: ran a `packages/cli/bin/run ui:create:react` command after shifting the template to `baguette` (because of course I own [cra-template-baguette](https://www.npmjs.com/package/cra-template-baguette) 🥖 )

-----
CDX-571


louis-bompart:
👋 @jpmarceau I reckon the bit that I updated in the ReadMe is imported into `doc.coveo.com` so I'd prefer having you checking it before I bring this to the trunk ;)

jpmarceau (approved):
👌 
Very nice, I recall that node-gyp was a bit annoying on my mac 😅

y-lakhdar:
The downside is that you loose all the feedback animation and colors in the terminal.
It used to be like that
image

Now you get stuck at that line without info on what is going on.
image

louis-bompart:
> The downside is that you loose all the feedback animation and colors in the terminal.
> It used to be like that
> image
> 
> Now you get stuck at that line without info on what is going on.
> image

You're right dang. Well, I have a few more aces up my sleeve.

louis-bompart:
@y-lakhdar, aces on table ♠️. We're inheriting from the main process TTY proper, the fancy bars and 'mojis should still be here.
And the process should work A-OK.

packages/cli/src/commands/ui/create/react.spec.ts
```diff
@@ -191,18 +216,23 @@ describe('ui:create:react', () => {
...
+        expect(mockedSpawn).nthCalledWith(
+          1,
+          expect.stringContaining('npx'),
+          [
+            'create-react-app@1.0.0',
+            'myapp',
+            '--template',
+            '@coveo/cra-template@1.0.0',
+          ],
+          expect.objectContaining({})
```
olamothe, line 231: Nitpick, but I think this could probably be transformed with `expect.any(Object)`

And is it really an empty object that we are receiving there ?

olamothe (approved):

packages/cli/src/commands/ui/create/react.spec.ts
louis-bompart: ✔️ I redid the test proper

louis-bompart:
will fix #464

PR #463, opened 2021-09-08
ci: add clean-up CRON for API Keys

y-lakhdar:
## Proposed changes

* Removed key deletion logic from jest teardown
* Added additional job to the **build workflow**.
* Added new CRON job that will run weekly.

The same JS script is used for both jobs to delete old API keys. The only difference is the presence of the `TEST_RUN_ID` environment variable.
When the variable is defined (meaning the job is following the E2E tests), only the API key associated to that job will be deleted.
Otherwise, all the keys older than a day will be deleted.

## Screenshots
image

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests: Created a [fork](https://github.com/y-lakhdar/cli/runs/3558681877?check_suite_focus=true) to test the Github action.


scripts/delete-api-keys.js
```diff
@@ -0,0 +1,61 @@
...
+
+async function main() {
+  const testOrgId = process.env.TEST_ORG_ID;
+  const accessToken = process.env.ACCESS_TOKEN;
+  const testRunId = process.env.TEST_RUN_ID;
+  const apiKeys = await getAllApiKeys(testOrgId, accessToken);
+
+  let cliApiKeys = apiKeys.filter(wasCreatedByTheCli(testRunId));
+  if (!testRunId) {
+    cliApiKeys = cliApiKeys.filter(isAtLeastOneDayOld);
```
olamothe, line 55: The check on `testRunId` means that it will only delete `isAtLeastOneDayOld` only when running the cron job ?

If so, I would just this an explicit script parameter (an argument to the script), instead of relying on a somewhat unclear environment variable injected by github.

Suggestion:
```
node ./scripts/delete-api-keys.js --onlyOlderThan 1d
```

olamothe (approved):

louis-bompart (changes requested):
I would use coveo/platform-client instead.
If the APIs are not here yet, let's make 'em :D

louis-bompart (approved):

PR #473, opened 2021-09-16
test(e2e): remove timeout on push command

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-588

## Proposed changes

Remove timeout when applying a snapshot. Otherwise, the push command is sometime timing out after 60 seconds.
There is already a timeout within the `beforeAll` scope.
 

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


louis-bompart (approved):

y-lakhdar:
> looks good, still failing tho

I still see the push command timing out... I'll check if the wait flag correctly works when the value is set to `0`

PR #474, opened 2021-09-16
fix(snapshot): wait flag should support the value `0`

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-589

## Proposed changes

There is a bug in the retry logic which is causing the `#waitUntilDone` method to automatically exit when the `wait` option equals `0`.

Reason is that `retries  =  Math.ceil(opts.wait / opts.waitInterval)  =  Math.ceil(0/1)  =  0`

The solution is to use the `forever` option.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


louis-bompart (changes requested):
Add a UT for wait=0 ;)

louis-bompart (approved):

PR #475, opened 2021-09-16
ci: use deploy key

louis-bompart:
## Proposed changes

Deploy Keys are SSH keys that are used for Git authentication.
Where they differ from a traditional SSH Key is that the SSH Key is attached to a repository instead of a user.

The Deploy Key, if given write access, can act with the same right as an administrator of the repository, but only on the Git side of GitHub (so, just commits, no 'comments', releases, or anything like that.

## Testing

Implemented on [coveo/github-actions](https://github.com/coveo/github-actions)

-----
CDX-592


.github/workflows/release.yml
```diff
@@ -29,6 +30,22 @@ jobs:
           git config --global user.email action@github.com
           git config --global user.name GitHub Action
           echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
+      - name: Setup git SSH remote
+        env:
+          DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
+        run: |
+          eval "$(ssh-agent -s)"
+          ssh-add - <<< "${DEPLOY_KEY}"
+          mkdir -p ~/.ssh
```
y-lakhdar, line 39: Why are you creating `~/.ssh/id_rsa` after adding the private key to the SSH auth agent?

louis-bompart: Because Git is not using the SSH Agent somehow 😅

y-lakhdar (approved):

.github/workflows/release.yml
louis-bompart: Good catch @y-lakhdar, removed the auth agent in 86b20f6

PR #477, opened 2021-09-17
feat: control the granularity of the preview with a `--previewLevel`

y-lakhdar:
## Proposed changes

The expanded preview can take several minutes to compute. The idea is to let the user choose the level of granularity when it comes to previewing the changes.

The user can choose 2 levels:
1. `light`: will only display the snapshot summary in the terminal
2. `detailed` (default): will display the light in the terminal as well as computing the expanded preview

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:
Ran preview and push commands using different level of preview


packages/cli/src/lib/flags/snapshotCommonFlags.ts
```diff
@@ -21,3 +21,14 @@ export const sync = () => ({
       'Apply synchronization when there is a 100% match between organization and snapshot resources.',
   }),
 });
+
+export const previewLevel = () => ({
+  previewLevel: flags.enum({
+    char: 'p',
+    description:
+      'Whether to only display a light or detailed preview of the overall changes',
+    options: ['light', 'detailed'],
```
louis-bompart, line 30: Store the value of the enum in an actual enum and use Object.keys ;) (or value I don't remember tbh ;)

You have examples of that I think with the resources coming from `platform-client` :)

louis-bompart (commented):
LGTM, enum tho ;)

packages/cli/src/lib/flags/snapshotCommonFlags.ts
```diff
@@ -21,3 +21,14 @@ export const sync = () => ({
       'Apply synchronization when there is a 100% match between organization and snapshot resources.',
   }),
 });
+
+export const previewLevel = () => ({
+  previewLevel: flags.enum({
+    char: 'p',
+    description:
+      'Whether to only display a light or detailed preview of the overall changes',
```
olamothe, line 29: I would add some explanation about why someone might want to use a `light` preview instead of always a `detailed` one.

Something that explains why it can save time.

olamothe (approved):

packages/cli/src/lib/flags/snapshotCommonFlags.ts
```diff
@@ -21,3 +21,14 @@ export const sync = () => ({
       'Apply synchronization when there is a 100% match between organization and snapshot resources.',
   }),
 });
+
+export const previewLevel = () => ({
+  previewLevel: flags.enum({
+    char: 'p',
+    description:
+      'The verbosity of the preview. The `light` preview is faster to generate but only contains a limited amount of information, as opposed to the `detailed` preview which takes more time to generate, but returns a diff representation of all the changes to apply.',
```
y-lakhdar, line 29: @jpmarceau can you let me know if this description makes sense?

jpmarceau, line 29:
```suggestion
      'The verbosity of the preview. The `light` preview is faster to generate but only contains a limited amount of information, as opposed to the `detailed` preview that takes more time to generate, but returns a diff representation of all the changes to apply.',
```

jpmarceau: Yep I think that's quite good 👍

jpmarceau: (just a grammar thing. 'which' needs to be preceded by a comma, whereas 'that' doesn't)

jpmarceau (approved):

packages/cli/src/lib/flags/snapshotCommonFlags.ts
y-lakhdar: Thanks! I'll remember that.

packages/cli/src/lib/flags/snapshotCommonFlags.ts
```diff
@@ -21,3 +26,14 @@ export const sync = () => ({
       'Apply synchronization when there is a 100% match between organization and snapshot resources.',
   }),
 });
+
+export const previewLevel = () => ({
+  previewLevel: flags.enum({
+    char: 'p',
+    description:
+      'The verbosity of the preview. The `light` preview is faster to generate but only contains a limited amount of information, as opposed to the `detailed` preview that takes more time to generate, but returns a diff representation of all the changes to apply.',
+    options: [PreviewLevel.Light, PreviewLevel.Detailed],
```
louis-bompart, line 35:
```suggestion
    options: Object.keys(PreviewLevelValue),
```

Like this, if for whatever reason we add a new PreviewLevel, we can keep this value 🎉 .

packages/cli/src/lib/flags/snapshotCommonFlags.ts
```diff
@@ -1,6 +1,11 @@
 import {flags} from '@oclif/command';
 import {Snapshot} from '../snapshot/snapshot';
 
+export enum PreviewLevel {
```
louis-bompart, line 4:
```suggestion
export enum PreviewLevelValue {
```

IMO, it's a pitfall waiting for us to fall in to have two variables that changes by a simple capitalization*

*: except for classes and instances, but here, for e.g. it become fuzzy in the imports

louis-bompart: Could be `PreviewLevelOption` too if you prefer

louis-bompart (commented):
LGTM, some tweaks in the flag & naming things 😅 !

packages/cli/src/lib/flags/snapshotCommonFlags.ts
y-lakhdar: Good point. but I think you meant `Object.values`.

PR #478, opened 2021-09-20
refactor: rename `org:config` commands to `org:resources`

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-594

## Proposed changes

Replacing the `config` keyword with `resources`.
This name change will will be significative once we implement the granularity on the pull command.

## Breaking changes

Since the command has not been released at this point, I don't think it should be considered as a breaking change

## Testing

- [x] Unit Tests: same UTs

- [x] Functionnal Tests:

- [x] Manual Tests: Ran a couple commands with the syntax


louis-bompart (approved):
smooth :D

olamothe (approved):

PR #479, opened 2021-09-20
fix: remove line break from url

y-lakhdar:
## Proposed changes

`cli.warn` formats the string given in parameter and injects line breaks based on the terminal width. That way it can prepend symbols at every lines (see [source code](https://github.com/oclif/errors/blob/master/src/errors/pretty-print.ts#L54)).

The proposed solution is to print the url with the `cli.log` instead of `cli.warn`. `cli.log` does not manipulate the string and ensures the uri is clickable even when it is wrapped.
image

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/lib/errors/snapshotErrors.ts
```diff
@@ -93,38 +93,50 @@ export class SnapshotSynchronizationAmbiguousMatchesError
   extends SnapshotError
   implements IDetailedReportable
 {
+  private synchronizationPlanUrl: string;
```
olamothe, line 96: I'd suggest not adding stateful class member for something that can be re-computed very easily.

Suggestion: 
```
private get synchronizationPlanUrl() {
return urlBuilder.getSynchronizationPage(this.snapshot);
}
```

olamothe (approved):

louis-bompart (approved):

louis-bompart (changes requested):
The `print` method is really similar in the two classes, it might be good to refactor here :)

louis-bompart (approved):

PR #483, opened 2021-09-22
test: test synchronization errors **WIP**

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-567

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/cli-e2e/__tests__/orgResources.specs.ts
+++ b/packages/cli-e2e/__tests__/orgResources.specs.ts
@@ -1,5 +1,10 @@
 import {join} from 'path';
-import {CLI_EXEC_PATH, getConfig, getPathToHomedirEnvFile} from '../utils/cli';
+import {
+  answerPrompt,
+  CLI_EXEC_PATH,
+  getConfig,
+  getPathToHomedirEnvFile,
+} from '../utils/cli';
 import {ProcessManager} from '../utils/processManager';
 import {Terminal} from '../utils/terminal/terminal';
 import {config} from 'dotenv';
@@ -7,6 +12,7 @@ import {ensureDirSync} from 'fs-extra';
 import PlatformClient from '@coveord/platform-client';
 import {getPlatformClient} from '../utils/platform';
 import {readdirSync} from 'fs';
+import {EOL} from 'os';
 config({path: getPathToHomedirEnvFile()});
 
 describe('org:resources', () => {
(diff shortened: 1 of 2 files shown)


packages/cli-e2e/__tests__/orgResources.specs.ts
```diff
@@ -94,25 +126,66 @@ describe('org:resources', () => {
...
+            .when(/Previewing resource changes/)
+            .on('stdout')
+            .do()
+            .once();
+
+          await previewTerminal
+            .when(warningRegex)
+            .on('stderr')
+            .do(answerPrompt(`y${EOL}`))
+            .until(previewTerminalExitPromise);
```
louis-bompart, line 184: There's no 'expectation' in this test 🤔 

The test should fail (nicely if possible and with a nice explanation) when the previewTerminal concludes without going thru the warningRegex.

Here, it won't I reckon.

louis-bompart (changes requested):
Tests should have expectations, and fail when they are not met.

packages/cli-e2e/__tests__/orgResources.specs.ts
```diff
@@ -94,25 +121,80 @@ describe('org:resources', () => {
...
+            .when(/Previewing resource changes/)
+            .on('stdout')
+            .do(() => {
+              returnedPreview = true;
+            })
+            .until(previewTerminalExitPromise);
+
+          await previewTerminalExitPromise;
+          expect(returnedSyncronizationWarning).toBe(true);
+          expect(returnedPreview).toBe(true);
```
louis-bompart, line 193: Small suggestions:
 - Typo in var name
 - Put the exitCondition at the end, simpler to understand I reckon, and no need to 'store' the promise like this.
 
```suggestion
          let returnedSynchronizationWarning = false;
          let returnedPreview = false;

          const previewTerminal = previewChange(
            testOrgId,
            processManager,
            'org-config-preview-unsync'
          );

          previewTerminal
            .when(/Checking for automatic synchronization/)
            .on('stderr')
            .do(() => {
              returnedSyncronizationWarning = true;
            })
            .until(previewTerminalExitPromise);

          previewTerminal
            .when(/Previewing resource changes/)
            .on('stdout')
            .do(() => {
              returnedPreview = true;
            })
            .until(previewTerminalExitPromise);

          await previewTerminal
            .when('exit')
            .on('process')
            .do()
            .once();

          expect(returnedSynchronizationWarning).toBe(true);
          expect(returnedPreview).toBe(true);
```

louis-bompart (approved):
LGTM!

packages/cli-e2e/__tests__/orgResources.specs.ts
y-lakhdar: If it is at the end, how can I use it in the `.until(` clause (lines 181 & 189 before your changes)?

louis-bompart: eh, you right nvm

PR #490, opened 2021-10-04
feat: support proxies (Tests in Progress)

y-lakhdar:
## Proposed changes

* Remove the `AppAuthJS` and its patch.
* Use [opener](https://www.npmjs.com/package/opener) to open the browser for the Auth.
* Use NodeJS HTTP server to handle the OAuth redirection and ‘get the code'
* Use [axios](https://www.npmjs.com/package/axios) to communicate with the Platform Authorization server and trade the code for a token. Axios also easily supports proxies.
 
## Testing (In progress)

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


louis-bompart (commented):
Good job really 👏

packages/cli/src/lib/oauth/oauthClientServer.ts
```diff
@@ -0,0 +1,102 @@
...
+            method: 'post',
+            url: tokenEndpoint,
+            ...this.requestConfig,
+            data,
+          });
+          const accessToken = authRequest.data.access_token;
+
+          res.end('Close your browser to continue');
+          resolve({accessToken});
+          server.close();
```
olamothe, line 39: Those `close()` call should be handled in some kind of `finally`, because it does not close when we throw on line 25, or for any other random error that could happen, right ?

packages/cli/src/lib/oauth/oauthClientServer.ts
```diff
@@ -0,0 +1,102 @@
...
+        username: client_id,
+        password: client_id,
+      },
+    };
+    const proxy = process.env.HTTPS_PROXY;
+    if (proxy) {
+      const proxyUrl = new URL(proxy);
+      config.proxy = {
+        host: proxyUrl.host,
+        port: parseInt(proxyUrl.port),
```
olamothe, line 65:
```suggestion
        port: parseInt(proxyUrl.port, 10),
```

packages/cli/src/lib/oauth/oauthClientServer.ts
```diff
@@ -0,0 +1,102 @@
...
+          const data = this.getTokenQueryString(code);
+          const authRequest = await axios({
+            method: 'post',
+            url: tokenEndpoint,
+            ...this.requestConfig,
+            data,
+          });
+          const accessToken = authRequest.data.access_token;
+
+          res.end('Close your browser to continue');
```
olamothe, line 37: Is this needed ? Can we auto-close the browser for the user ?

olamothe (approved):

packages/cli/src/lib/oauth/oauthClientServer.ts
y-lakhdar: Yes, it is needed. I thought I could return an HTML page containing a script tag to automatically close the browser, but it does not work every time.

Even if the CLI opens the browser tab, it is not allowed to close it once the OAuth flow ends.
Instead, we get the following error when we try to close it `Scripts may close only the windows that were opened by them.`

However, it works as expected when the user is already logged in to the platform.

But even when the CLI is able to auto-close the browser tab, there is a small glitch in the terminal that I have not been able to figure out yet.

All that to say that the auto-closing is currently creating more issues rather than solving the existing one.

packages/cli/src/lib/oauth/oauthClientServer.ts
y-lakhdar: Yes, it "should" close the server even if an error is fired.
Any error will be intercepted in the `catch` block which will close the server and gracefully return the error message to be handled by Oclif.

image

louis-bompart: It's true that line 42 we do call server.close but @olamothe ain't wrong to say we should do the `server.close()` in a `finally`:
 - It avoid repetition (one line, I admit, that's nothing)
 - More importantly, it's semantically more accurate: "we do this, if things goes haywire we do catch, and no matter what we finish by closing the server"

PR #512, opened 2021-10-22
fix(proxy): pass proxy along to platform-client

louis-bompart:
## Proposed changes

coveo/platform-client is using `isomorphic-fetch` for querying which does not support 'natively' the HTTPS_PROXY environment variable. To fix that, we pass a https-proxy-agent along iif the HTTPS_PROXY is there.

## Testing

- [X] Functionnal Tests: Added a smoke test using mitmproxy to check that the traffic is properly routed thru.

-----
CDX-650


packages/cli-e2e/utils/terminal/do.ts
```diff
@@ -152,7 +152,7 @@ export class Do implements Promise<void> {
       onParams = [
         'data',
         (data: Buffer) => {
-          if (stripAnsi(data.toString()).replace(/\n/g, '').match(condition)) {
+          if (condition.test(stripAnsi(data.toString()).replace(/\n/g, ''))) {
```
louis-bompart, line 155: This was a loose assertion that could've been well not loose. Now it is.

y-lakhdar (approved):

PR #514, opened 2021-10-26
feat: catch precondition errors

y-lakhdar:
## Proposed changes

Previously, since the preconditions where retuning `false` when not respected, the error message was not [handled in the `catch` method](https://oclif.io/docs/error_handling#error-handling-in-the-catch-method) and, therefore, not tracked in the UA. 

The idea is to throw [PreconditionError](https://github.com/coveo/cli/blob/63cb1223c3308b7dc28ac3bf6a49e8b76a492147/packages/cli/src/lib/errors/preconditionError.ts) whenever a precondition is not respected.

This will facilitate the integration of Amplitude in the CLI when we have to track errors.

Furthermore, having a  [PreconditionError](https://github.com/coveo/cli/blob/63cb1223c3308b7dc28ac3bf6a49e8b76a492147/packages/cli/src/lib/errors/preconditionError.ts) class will make sure errors are consistent in the Amplitude dashboards.

### Commits
* Throw `PreconditionError` instead of returning false (https://github.com/coveo/cli/commit/52fc19992ec15fed529397c2da3047f95eb26a28)
* Simplify some unit tests (https://github.com/coveo/cli/commit/c3a830b85257fd44da65e3542b119594abd72f0f, https://github.com/coveo/cli/commit/aebc606bed5a4a51816cb6b42aa43c0990da971c)
* Remove the `I` from `IDetailedReportable` (https://github.com/coveo/cli/commit/869e05cee5118079beef1b6eb00ca2969a06b55c)

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/__test__/preconditionUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const thrower = (reason: string) => {
+  throw new PreconditionError(`${reason} Precondition Error`);
+};
+
+export const mockPreconditions = <Type extends Record<string, boolean>>(
+  preconditionStatus: Type
+) => {
+  type StatusPromises = Record<
+    keyof typeof preconditionStatus,
```
louis-bompart, line 12: Can't we use `keyof Type` ?

packages/cli/src/__test__/preconditionUtils.ts
```diff
@@ -0,0 +1,28 @@
+import Command from '@oclif/command';
+import {PreconditionError} from '../lib/errors/preconditionError';
+
+const thrower = (reason: string) => {
+  throw new PreconditionError(`${reason} Precondition Error`);
+};
+
+export const mockPreconditions = <Type extends Record<string, boolean>>(
```
louis-bompart, line 8: 'Type' is quite generic for a type name

louis-bompart (approved):

packages/cli/src/lib/decorators/preconditions/preconditions.ts
```diff
@@ -3,7 +3,7 @@ import type Command from '@oclif/command';
 export type PreconditionFunction = (
   target: Command,
   instance?: Command
-) => never | void | Promise<never | void>;
+) => Promise<never | void>;
```
y-lakhdar, line 7: I think it does not make sense to not return a promise if at the end we do
```
await precondition.call(this, target);
```

packages/cli/src/__test__/preconditionUtils.ts
```diff
@@ -1,24 +1,25 @@
...
+  PreconditionStatus extends Record<string, boolean>
+>(
+  preconditionStatus: PreconditionStatus
 ) => {
-  type StatusPromises = Record<
-    keyof typeof preconditionStatus,
-    (_target: Command) => Promise<void>
-  >;
+  type preconditionKeys = keyof typeof preconditionStatus;
+  type preconditionPromises = Record<preconditionKeys, PreconditionFunction>;
```
y-lakhdar, line 15: Let's use the type we already have defined in the precondition file

y-lakhdar (commented):
@louis-bompart I applied your request and took the liberty to bring some improvements.
Let me know what you think

packages/cli/src/lib/decorators/preconditions/preconditions.ts
louis-bompart: > I think it does not make sense to not return a promise if at the end we do
> 
> ```
> await precondition.call(this, target);
> ```

An async function always returns a Promise.

y-lakhdar: yes, that's what I am saying :) 
hence the `never | void` removal

louis-bompart: My bad! I misread your comment at first lgtm! 👍

PR #523, opened 2021-11-02
chore: replace `coveo.analytics` with `@amplitude/node`

y-lakhdar:
## Proposed changes

* Created a Trackable decorator that can be used on each Command. The decorator will automatically track commands args and flags (as well as catched errors).
* Replaced the `coveo.analytics` package with `@amplitude/node`
 
## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:
Inspected fired events after running successful and failing commands.

p.s. Sorry, I wanted to make the review easier to follow by rebasing similar commits... I ended up squashing the entire thing 🤦


packages/cli/src/lib/decorators/preconditions/trackable.ts
```diff
@@ -0,0 +1,67 @@
+import Command from '@oclif/command';
```
y-lakhdar, line 1: This is the main change of this PR!

packages/cli/package.json
```diff
@@ -7,6 +7,8 @@
     "node": ">=14.15.1"
   },
   "dependencies": {
+    "@amplitude/identify": "^1.9.0",
+    "@amplitude/node": "^1.9.0",
```
louis-bompart, line 11: Can we remove `coveo.analytics`?

packages/cli/src/hooks/analytics/analytics.spec.ts
```diff
@@ -97,114 +104,23 @@ describe('analytics hook', () => {
...
-          errorName: err.name,
-          errorStacktrace: err.stack,
-        }),
-      })
-    );
-  });
+  // TODO: CDX-651: should send environment from config
+  // TODO: CDX-651: should send organization type from config
+  // TODO: CDX-651: should send region from config
+  // TODO: CDX-651: should send hashed username from platform-client
```
louis-bompart, line 123: create `it.todo` for the spec already ;)

packages/cli/src/hooks/analytics/analytics.ts
```diff
@@ -1,89 +1,46 @@
...
 export interface AnalyticsHook {
-  commandID: string;
-  flags: Record<string, unknown>;
+  event: Event;
   config: IConfig;
-  err?: Error;
 }
 
-const analyticsAPIKey = 'xx01ad67bb-f837-4a3e-8669-16397994a6f2';
+// import {Identifier} from './identifier';
```
louis-bompart, line 16: What is Identifier, why is it commented out?

packages/cli/src/hooks/analytics/analytics.ts
```diff
@@ -101,61 +58,63 @@ const platformInfoIdentifier = async () => {
...
+  const schema = {
+    message: 'string',
+    errorCode: 'string',
+    requestID: 'string',
+  };
+  const isErrorFromAPI = validate(arg, schema);
+  return isErrorFromAPI
+    ? new APIError(arg as APIErrorResponse)
+    : new Error('Unknown Error');
 };
```
louis-bompart, line 118: In the end, are we gonna log only the instanceOf `BaseCLIError` (e.g. name, you get the gist)?

If so, I would keep that bit really dead simple with only the arg instanceof Error and call it a day

packages/cli/src/lib/decorators/preconditions/trackable.spec.ts
```diff
@@ -0,0 +1,112 @@
+// jest.mock('@oclif/command');
```
louis-bompart, line 1: Sooo... do we need it?

packages/cli/src/lib/decorators/preconditions/trackable.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+import {getFakeCommand} from './testsUtils/utils';
+import {Trackable} from './trackable';
+
+describe('trackable', () => {
+  const mockedAnalyticHook = jest.fn();
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('if command successfully completes', () => {
```
louis-bompart, line 12:
```suggestion
  describe('when the command successfully completes', () => {
```

packages/cli/src/lib/decorators/preconditions/trackable.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+      expect(mockedAnalyticHook).toHaveBeenNthCalledWith(1, 'analytics', {
+        event: {...event, ...{event_type: 'started foo bar'}},
+      });
+      expect(mockedAnalyticHook).toHaveBeenNthCalledWith(2, 'analytics', {
+        event: {...event, ...{event_type: 'completed foo bar'}},
+      });
+    });
+  });
+
+  describe('if the command throws an error', () => {
```
louis-bompart, line 62:
```suggestion
  describe('when the command throws an error', () => {
```

packages/cli/src/lib/decorators/preconditions/trackable.ts
louis-bompart:
```suggestion
import type Command from '@oclif/command';
```

packages/cli/package.json
```diff
@@ -7,6 +7,8 @@
     "node": ">=14.15.1"
   },
   "dependencies": {
+    "@amplitude/identify": "^1.9.0",
```
louis-bompart, line 10: Can we remove `coveo.analytics` ?

louis-bompart (approved):
LGTM overall, some changes here and there but GG 👍

packages/cli/package.json
y-lakhdar: @olamothe, do you see any issue with replacing Coveo UA? or you think we should wait until we have Amplitude data in prod

olamothe: I think it's okay if we remove coveo.analytics, just make sure that it does not crash/there's no error before doing so.

That's the most important part.

olamothe (approved):
Really like the `@TrackablePattern` !

packages/cli/src/hooks/analytics/analytics.ts
y-lakhdar: Sorry. This will be part of the next PR (CDX-651). I will add the Jira number to the comment.

packages/cli/src/hooks/analytics/analytics.ts
y-lakhdar: Yes. This will be part of the UA event sanitizing process (CDX-660).
There is no rush since we are only tracking Trial orgs.

PR #534, opened 2021-11-05
feat: add "coveo ui:create:atomic" command

ThibodeauJF:
## Proposed changes

Adds the new "coveo ui:create:atomic" command (hidden)

## Testing

No UTs for now as the command is hidden and in development

CDX-670 https://coveord.atlassian.net/browse/CDX-670


packages/atomic-template/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@coveo/create-atomic",
```
ThibodeauJF, line 2: **For the next steps, I have a few questions!**

All the other projects use a separate generator package, or a plugin for the generators offered by the community (e.g. `create-react-app`)

I guess we will have to do the same for Atomic, I decided to name that package `@coveo/create-atomic` so that it works standalone when doing `npm init @coveo/atomic`. The CLI will instantiate a new project and set the .env variables.

My question is do you have a technology to recommend for a generator, I'm seeing the use of [@schematics/angular](https://angular.io/guide/schematics) in the angular project which appears to offers everything necessary but seems to be more for well, Angular. React & Vue also have their own tools.

Exploring and playing around I found a few possible options:
- [Yeoman Generator](https://yeoman.io/authoring/)
- [Plop](https://plopjs.com/)
- Not using any technology (or smaller separate packages) seems also possible, but perhaps counter-productive

The Atomic project that will be create will basically contain:
- index.html file with an Atomic interface
- a few CSS files
- the `search-token-server` leveraging the .env variables

PS: In the future, when we want to create a more complex setup that can be used to create custom Atomic component (see [design doc](https://docs.google.com/document/d/1QzqezewjxUNdd2MhEQ4bmMT_0K1PcQbF3FBhzR2yka4/edit#heading=h.fkb38024b4a9)) we would then have something similar to what Stencil offers with their `npm init stencil` command (see [create-stencil](https://www.npmjs.com/package/create-stencil))

y-lakhdar: I think both Plop and Yeoman are good candidates for the @coveo/atomic project.

olamothe: I used Plop before, it does the job and I don't recall facing any bugs.

packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  public static args = [
+    {
+      name: 'name',
+      description: 'The name of the application to create.',
+      required: true,
+    },
+  ];
+
+  @Preconditions(
```
olamothe, line 34: // TODO: Add the `@Trackable()` decorator from this pr: https://github.com/coveo/cli/pull/523

louis-bompart:
Looking good so far!
I propose that you do the following to keep things reviewable:
 1. Mark the command as hidden (see suggestion)
 2. Revert the change to the readme
 
 And let's merge that PR, like this you'll have the basic scaffolding to build on

packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -0,0 +1,74 @@
...
+import {spawnProcess} from '../../../lib/utils/process';
+
+interface AtomicArguments {
+  name: string;
+}
+
+export default class Atomic extends Command {
+  public static description =
+    "Create a Coveo Headless-powered search page with Coveo's own Atomic framework. See <https://docs.coveo.com/atomic> and <https://docs.coveo.com/headless>.";
+
```
louis-bompart, line 25:
```suggestion
  public static hidden = true;
```

To hide the command, if you go with what I proposed

packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -0,0 +1,76 @@
...
+    // @Trackable(),
+    HasNecessaryCoveoPrivileges(createApiKeyPrivilege, impersonatePrivilege)
+  )
+  public async run() {
+    await this.createProject();
+    this.displayFeedbackAfterSuccess();
+    await this.config.runHook('analytics', buildAnalyticsSuccessHook(this, {}));
+  }
+
+  public async catch(err?: Error) {
```
y-lakhdar, line 46: `// TODO: @Trackable()` here too. since we also want to track catched errors

louis-bompart (approved):

packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -0,0 +1,76 @@
...
+    await this.config.runHook(
+      'analytics',
+      buildAnalyticsFailureHook(this, {}, err)
+    );
+    throw err;
+  }
+
+  private async createProject() {
+    spawnProcess(appendCmdIfWindows`echo`, [
+      '...creating new Atomic project:',
```
y-lakhdar, line 56: Maybe just add the name of the JIRA in a `// TODO: `.

y-lakhdar (approved):

packages/cli/src/commands/ui/create/atomic.ts
ThibodeauJF: updated in https://github.com/coveo/cli/pull/536

louis-bompart (approved):

PR #535, opened 2021-11-05
feat: improve compatibility with WSL

louis-bompart:
## Proposed changes

[open](https://github.com/sindresorhus/open) works better with WSL, so I replaced `opener` by the former. 🎉 

## Testing

- [x] Unit Tests: existing, and our UT don't care about the OS
- [x] Functional Tests: existing use cases are still tested, however WSL is not (I reckon it's not worth the trouble yet)
- [x] Manual Tests: Tested on my WSL & Windows environment
-----
CDX-664


packages/cli/src/lib/oauth/oauth.ts
```diff
@@ -54,7 +54,9 @@ export class OAuth {
 
   private openLoginPage(state: string) {
     const url = this.buildAuthorizationUrl(state);
-    opener(url);
+    open(url).then((browserProcess) => {
```
y-lakhdar, line 57: why not use await?

louis-bompart: Because we don't need for the browser to be open and all to start opening the 'listening' server, so we might as well do so to shave some time

louis-bompart: Another alternative would be to change the `openLoginPage` to be asynchronous and thus await on the `open` call but not on the call of `openLoginPage`.

What do you prefer?

olamothe (approved):

y-lakhdar (approved):

PR #536, opened 2021-11-08
feat: add "create-atomic" package

ThibodeauJF:
## Proposed changes

This adds the new package `@coveo/create-atomic` which the CLI will use. I used [Plop.js](https://plopjs.com/) as a generator.

Upcoming features are (next PRs):
- Make the CLI setup the env variables
- Connect Atomic app with token server (in index.html, refresh token method)
- Add cli command doc & unhide command

## Testing

Tests will be in an incoming PR

CDX-670
https://coveord.atlassian.net/browse/CDX-670


packages/create-atomic/templates/public/index.hbs
```diff
@@ -0,0 +1,69 @@
+<html lang='en'>
```
olamothe, line 1: If I remember correctly from your design doc, down the line, these files (index.hbs, layout.css, theme.css) would be { returned/powered/controlled/templated/etc } by the search interface service as the source of truth, right ?

olamothe (approved):

packages/create-atomic/templates/public/index.hbs
ThibodeauJF: Yes for phase 2 https://docs.google.com/document/d/1QzqezewjxUNdd2MhEQ4bmMT_0K1PcQbF3FBhzR2yka4/edit#heading=h.xc32y4ockguh

Currently there is no endpoint that returns the Atomic html configuration

ThibodeauJF:
Temporarily closing

packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -20,9 +23,11 @@ interface AtomicArguments {
 }
 
 export default class Atomic extends Command {
+  public static cliPackage = '@coveo/create-atomic';
+  public static requiredNodeVersion = '>=8.9.4';
```
ThibodeauJF, line 27: Plop requires this, does `@coveo/search-token-server` require a specific version?

packages/create-atomic/templates/.env.example
```diff
@@ -0,0 +1,33 @@
+# The Plaform URL to use.
```
ThibodeauJF, line 1: Copied from the cra-template project and adapted

packages/create-atomic/templates/scripts/clean-up.js
```diff
@@ -0,0 +1,19 @@
+const {unlinkSync, readFileSync, writeFileSync} = require('fs');
```
ThibodeauJF, line 1: like the other /scripts/, i've only adapted from other projects

packages/create-atomic/templates/scripts/start-app.js
```diff
@@ -0,0 +1,12 @@
+const {config} = require('dotenv');
```
ThibodeauJF, line 1: This one is an original script, for the Atomic app

packages/cli/src/commands/ui/create/atomic.ts
y-lakhdar: Good question.
I was able to make it work with Node v7. However, the search token server uses dotenv v10 which, based on their documentation, requires node v10.

packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -32,11 +37,15 @@ export default class Atomic extends Command {
...
     IsAuthenticated(),
+    IsNpxInstalled(),
+    IsNodeVersionInRange(Atomic.requiredNodeVersion),
     HasNecessaryCoveoPrivileges(createApiKeyPrivilege, impersonatePrivilege)
   )
   public async run() {
     await this.createProject();
     this.displayFeedbackAfterSuccess();
+    // TODO: add env variables
+    await this.config.runHook('analytics', buildAnalyticsSuccessHook(this, {}));
```
y-lakhdar, line 48: We no longer need to call the `runHook` method. Analytic tracking is being tacking care of by the `@Trackable` decorator.

packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -45,8 +54,8 @@ export default class Atomic extends Command {
   }
 
   private async createProject() {
-    spawnProcess(appendCmdIfWindows`echo`, [
-      '...creating new Atomic project:',
+    return spawnProcess(appendCmdIfWindows`npx`, [
+      `${Atomic.cliPackage}@${getPackageVersion(Atomic.cliPackage)}`,
```
y-lakhdar, line 58: I think this will fail in the E2E tests since the tests need to target a specific version (in Verdaccio).
You can solve that by adding a `--version` flag

```
public static flags = {
  version: flags.string({
    char: 'v',
    description: `The version of ${Atomic.cliPackage} to use.`,
    default: getPackageVersion(Atomic.cliPackage),
  }),
};
```

and then,

```
return spawnProcess(appendCmdIfWindows`npx`, [
      `${Atomic.cliPackage}@${flags.version}`,
```

packages/create-atomic/index.js
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+const path = require('path');
+const args = process.argv.slice(2);
+const {Plop, run} = require('plop');
+const argv = require('minimist')(args);
```
y-lakhdar, line 5: I don't see `minimist` in the package.json... Is that expected?

packages/create-atomic/templates/scripts/clean-up.js
```diff
@@ -0,0 +1,19 @@
...
+const {join} = require('path');
+const {getPackageManager} = require('./utils');
+const pkgJson = JSON.parse(readFileSync('package.json'));
+
+['postinstall', 'setup-env', 'setup-server', 'setup-cleanup'].forEach(
```
y-lakhdar, line 6: Good idea. I think we should do the same for the other frameworks.

y-lakhdar (approved):

packages/create-atomic/plopfile.js
```diff
@@ -0,0 +1,58 @@
...
+    prompts: [
+      {
+        type: 'input',
+        name: 'project',
+        message: 'Name of the project',
+        validate: (input) => {
+          const pattern =
+            '^(?:@[a-z0-9-*~][a-z0-9-*._~]*/)?[a-z0-9-~][a-z0-9-._~]*';
+          if (!new RegExp(pattern).test(input)) {
+            return `The project name should match the pattern "${pattern}" (https://docs.npmjs.com/cli/v7/configuring-npm/package-json#name)`;
```
olamothe, line 19: What would be the potential problem by letting users input "anything" for the project name ? 
Do we really need to validate here ?

Because the pattern is pretty convoluted, so I'm not sure, as an end user, that I would be able to figure out "oh yeah I should change this in my project name".

Maybe adding an example of a valid name ?

olamothe (approved):
GG !

packages/create-atomic/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+Plop.launch(
+  {
+    cwd: argv.cwd,
+    configPath: path.join(__dirname, 'plopfile.js'),
+    require: argv.require,
+    completion: argv.completion,
+  },
+  (env) => run(env, undefined, true)
+);
```
louis-bompart, line 15: If possible, can you try with Node16 and `import`.
We're keen to move fully to ECMA Script Module (ESM), so I'd say, especially given the simplicity of the script we could do an `index.mjs` ;)

packages/create-atomic/plopfile.js
```diff
@@ -0,0 +1,58 @@
...
+        templateFiles: [
+          'templates/public/style/*',
+          'templates/scripts/*',
+          'templates/.env.example',
+          'templates/.gitignore',
+        ],
+      },
+      function installPackages(answers) {
+        const {project} = answers;
+        spawn('npm', ['i'], {
```
louis-bompart, line 50: You need the OS handling here see `packages/create-atomic/templates/scripts/utils.js`.

packages/create-atomic/plopfile.js
```diff
@@ -0,0 +1,58 @@
...
+  plop.setGenerator('@coveo/atomic', {
+    description: 'A Coveo Atomic Generator',
+    prompts: [
+      {
+        type: 'input',
+        name: 'project',
+        message: 'Name of the project',
+        validate: (input) => {
+          const pattern =
+            '^(?:@[a-z0-9-*~][a-z0-9-*._~]*/)?[a-z0-9-~][a-z0-9-._~]*';
```
louis-bompart, line 17: voodoo regexp, what's that 😂

louis-bompart (changes requested):
The overall idea is there, however, I think we should either use TypeScript, or at the very least import instead of require given the simplicity of the package.

- example: `https://github.com/plopjs/plop/tree/master/example/typescript`

packages/create-atomic/.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "../../.eslintrc",
+    "rules": {
+      "node/no-unpublished-import": ["off"]
```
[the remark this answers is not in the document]
louis-bompart: (in fact, probably linked to what @y-lakhdar said about minimist

packages/create-atomic/plopfile.js
ThibodeauJF: That's NPM's official regex, I'll remove it and perhaps just kebab case the input

louis-bompart: It's OK to keep it, just 'document it', i.e. assign it to a variable and name it well, and add a link to your source.

packages/create-atomic/index.js
ThibodeauJF: Weird right, that's what Plop documents, it's once or their dependencies https://github.com/plopjs/plop/blob/master/package.json#L51

I'll add it just to be safe

packages/create-atomic/templates/scripts/clean-up.js
ThibodeauJF: Wasn't my idea 😉

packages/create-atomic/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "name": "@coveo/create-atomic",
+  "description": "Coveo Atomic Generator",
+  "version": "1.18.0",
+  "bin": {
+    "create-atomic": "./index.mjs"
+  },
+  "main": "index.mjs",
+  "scripts": {
+    "start": "plop",
+    "test": "echo \"Error: no test specified\" && exit 1",
```
louis-bompart, line 11: This will might make CI crash because the root test script invoke lerna exec test iirc

packages/create-atomic/plopfile.ts
```diff
@@ -0,0 +1,62 @@
...
+        type: 'input',
+        name: 'project',
+        message: 'Name of the project',
+        validate: (input) => {
+          // Taken from http://json.schemastore.org/package
+          const npmPackageRegex =
+            '^(?:@[a-z0-9-*~][a-z0-9-*._~]*/)?[a-z0-9-~][a-z0-9-._~]*';
+          if (!new RegExp(npmPackageRegex).test(input)) {
+            return 'Project name is invalid, please follow the guidelines: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#name)';
+          }
```
louis-bompart, line 24:
```suggestion
          const npmPackageRegex =
            /^(?:@[a-z0-9-*~][a-z0-9-*._~]*/)?[a-z0-9-~][a-z0-9-._~]*/;
          if (!npmPackageRegex.test(input)) {
            return 'Project name is invalid, please follow the guidelines: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#name)';
          }
```

packages/create-atomic/plopfile.ts
```diff
@@ -0,0 +1,62 @@
+import {NodePlopAPI} from 'plop';
+import {spawn} from 'child_process';
+import {getPackageManager} from './utils';
+
+interface PromptsAnswers {
+  project: string;
+}
+
+module.exports = function (plop: NodePlopAPI) {
```
louis-bompart, line 9: IDK if plop support ESM exports. If yes:
```suggestion
export default function (plop: NodePlopAPI) {
```

If it doesn't support it (or if if just ain't working), do this:
```suggestion
export = function (plop: NodePlopAPI) {
```
See: https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Modules.md#export--and-import--require

packages/create-atomic/plopfile.ts
```diff
@@ -0,0 +1,62 @@
...
+        base: 'templates',
+        templateFiles: [
+          'templates/public/style/*',
+          'templates/scripts/*',
+          'templates/.env.example',
+          'templates/.gitignore',
+        ],
+      },
+      function installPackages(answers) {
+        const {project} = answers as PromptsAnswers;
```
louis-bompart, line 53: Wouldn't that work too?
```suggestion
      function installPackages(answers: PromptAnswers) {
        const {project} = answers;
```

packages/create-atomic/.eslintrc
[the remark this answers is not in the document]
louis-bompart: @ThibodeauJF I pulled the branch, I'm pretty sure you do not need it.

packages/create-atomic/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "npm:bump:template": "npm --no-git-tag-version version",
```
louis-bompart, line 12: Otherwise, linting is not run at the CI level.
```suggestion
    "npm:bump:template": "npm --no-git-tag-version version",
    "lint": "prettier --check . && eslint ."
```

packages/create-atomic/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@coveo/create-atomic",
+  "description": "Coveo Atomic Generator",
+  "version": "1.18.0",
```
louis-bompart, line 4: I think we should not publish this package yet, we should add https://docs.npmjs.com/cli/v7/configuring-npm/package-json#private for now.

packages/create-atomic/plopfile.ts
ThibodeauJF: Screen Shot 2021-11-11 at 12 04 22 PM
 Nope, plop's type are 💩

ThibodeauJF:
Oups! gotta fix linting now 😄

packages/create-atomic/plopfile.ts
louis-bompart: :feelsgood: oh well ✅

packages/create-atomic/templates/scripts/start-server.js
```diff
@@ -0,0 +1,40 @@
...
+    stdio: 'inherit',
+    env: getEnvVariables(),
+    cwd: resolve(serverPath),
+  });
+  const handleKill = () => {
+    child.kill();
+  };
+  process.on('SIGINT', handleKill);
+  process.on('SIGHUP', handleKill);
+  process.on('SIGBREAK', handleKill);
```
louis-bompart, line 33: Yeah that's a Windows oddities. It's unlike me to say that about smthg windows related but fdat, it's too exotic.
Even PowerShell/WindowsTerminal behave strangely when you send a SIGHUP. Don't wanna deal with that 😂 
```suggestion
```

packages/cli/package.json
```diff
@@ -44,6 +44,7 @@
     "@coveo/angular": "^1.19.0",
     "@coveo/cra-template": "^1.19.0",
     "@coveo/vue-cli-plugin-typescript": "^1.19.0",
+    "@coveo/create-atomic": "^1.19.0",
```
ThibodeauJF, line 47: Should I remove this for now?

louis-bompart (approved):

packages/cli/package.json
louis-bompart: Oh yeah. Good catch. Indeed, remove it.

packages/create-atomic/package.json
```diff
@@ -0,0 +1,34 @@
...
+  "description": "Coveo Atomic Generator",
+  "version": "1.19.0",
+  "bin": {
+    "create-atomic": "./index.mjs"
+  },
+  "main": "index.mjs",
+  "scripts": {
+    "start": "plop",
+    "npm:bump:template": "npm --no-git-tag-version version",
+    "npm:publish:template": "npm publish --access public --registry http://localhost:4873",
```
louis-bompart: To fix the E2E for now.

```suggestion
    "notyet:npm:bump:template": "npm --no-git-tag-version version",
    "notyet:npm:publish:template": "npm publish --access public --registry http://localhost:4873",
```

PR #537, opened 2021-11-09
chore: track unique users in Amplitude

y-lakhdar:
## Proposed changes

Each amplitude event need at least a userId or a deviceId.
### User ID
Hashing the user email with a sha256 hash function.

### Device ID
[node-machine-id](https://www.npmjs.com/package/node-machine-id) is used to generate a unique ID based on the user system.

### User properties
A bunch of other user properties are sent to Amplitude. I made sure to review each one of them to preserve the user anonymity. For the full list of user properties -> [Amplitude Taxonomy](https://docs.google.com/spreadsheets/d/1lexw5Nmg0qmtwaHHtBydDlJxllS_K5xSeigE4z3pGDk/edit#gid=656872197).

[Amplitude documentation](https://help.amplitude.com/hc/en-us/articles/115003135607-Track-unique-users-in-Amplitude).

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:
Previewed the events in the Amplitude Dashboard


packages/cli/src/hooks/analytics/identifier.ts
```diff
@@ -0,0 +1,85 @@
...
+    return {
+      shell,
+      arch,
+      platform,
+      windows,
+      bin,
+      userAgent,
+      debug,
+    };
+  }
```
louis-bompart, line 80: So, we're doing fingerprinting, which is OK in itself, but in what conditions are we doing it (i.e. user consent)?

If we're doing without consent, please loop me in thru DM just to be sure we're on the up-and-up on this.

louis-bompart (changes requested):
To fix: Regexp
To discuss: 🇪🇺 laws lawl.

packages/cli/src/hooks/analytics/identifier.ts
olamothe: It should be the same condition as previous: 
* When you first run the CLI, we ask you about telemetry tracking, you can say `y/n`
* We store your choice for telemetry in config.
* Analytics are then only sent if config is set to `y` and you are a trial user.

louis-bompart: > It should be the same condition as previous:
> 
> * When you first run the CLI, we ask you about telemetry tracking, you can say `y/n`
> * We store your choice for telemetry in config.
> * Analytics are then only sent if config is set to `y` and you are a trial user.

👍 all good then!

olamothe (approved):

packages/cli/src/hooks/analytics/analytics.ts
```diff
@@ -5,53 +5,54 @@ import {
...
+  const {organization, analyticsEnabled, authenticatedClient} =
+    await platformInfoIdentifier();
 
   if (!analyticsEnabled) {
     return;
   }
-  const platformClient = await new AuthenticatedClient().getClient({
+
+  // TODO: CDX-676: remove this block to track events from all org types
+  // ↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓
```
louis-bompart, line 32: Hint, there's `#region` with VSCode that can be useful to mark 'area' of code.
Tho no big deal really here.

louis-bompart (approved):

packages/cli/src/hooks/analytics/analytics.ts
y-lakhdar: Good to know... but it's another extension to install

PR #539, opened 2021-11-10
chore: define user sessions

y-lakhdar:
## Proposed changes

Unlike other client-side SDKs, The Node.js SDK for Amplitude does not automatically generate session IDs on logged events.
I simply added a simple session manager to generate session IDs following the same logic Amplitude implements on browsers.
Meaning that all events fired within 30 minutes of each other are counted as part of the same session. Otherwise, a new session is generated ([Amplitude documentation](https://help.amplitude.com/hc/en-us/articles/115002323627-Tracking-sessions-in-Amplitude)).

If you are asking why the session id should be a Unix timestamp, read [this section](https://help.amplitude.com/hc/en-us/articles/115002323627-Tracking-sessions-in-Amplitude#h_408dc7ad-4919-4fa1-a506-c43174f68096) to understand how Amplitude calculates session length.

### Result in Amplitude
Screenshot of events within the same session
image

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:
Ran event within the same session and outside the session


packages/cli/src/hooks/analytics/session.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+import type {IConfig} from '@oclif/config';
+import {mocked} from 'ts-jest/utils';
+import {Config} from '../../lib/config/config';
+import check from './session';
+
+describe('session', () => {
+  const mockedConfig = mocked(Config, true);
+  const mockedConfigGet = jest.fn();
+  const mockedConfigSet = jest.fn();
+  const theDate = new Date('1997-08-29T12:00:00Z');
```
louis-bompart, line 12: I guess you need a non 0 epoch that's why?

packages/cli/src/hooks/analytics/session.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+  const dummySessionId = 1234567;
+
+  const doMockConfig = () => {
+    mockedConfig.prototype.get = mockedConfigGet;
+    mockedConfig.prototype.set = mockedConfigSet;
+  };
+
+  const freezeTime = () => {
+    jest.useFakeTimers().setSystemTime(theDate.getTime());
```
louis-bompart, line 21: As [Jest](https://jestjs.io/docs/timer-mocks) says: [Great Scott!](https://www.youtube.com/watch?v=QZoJ2Pt27BY&ab_channel=VideoConvers)

packages/cli/src/hooks/analytics/session.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+  };
+
+  const getTimeBefore = (minutes: number) => {
+    return theDate.getTime() - minutes * 60e3;
+  };
+
+  beforeAll(() => {
+    freezeTime();
+    global.config = {configDir: 'the_config_dir'} as IConfig;
```
louis-bompart, line 30: Mh, you should instead mock `Config` itself IMO.
This could lead to side-effects in other tests suite (current or future)

packages/cli/src/hooks/analytics/session.ts
```diff
@@ -0,0 +1,35 @@
...
+    : amplitudeSession;
+
+  cfg.set('amplitudeSession', session);
+  cfg.set('lastEventLoggedTime', now);
+
+  return session;
+};
+
+const hasBeenToolong = (lastEventFired: number, threshold: number) => {
+  console.log(lastEventFired < threshold);
```
louis-bompart, line 31: Console.log? :suspect:

packages/cli/src/hooks/analytics/session.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    mockedConfig.prototype.get = mockedConfigGet;
+    mockedConfig.prototype.set = mockedConfigSet;
+  };
+
+  const freezeTime = () => {
+    jest.useFakeTimers().setSystemTime(theDate.getTime());
+  };
+
+  const getTimeBefore = (minutes: number) => {
+    return theDate.getTime() - minutes * 60e3;
```
louis-bompart, line 25: feel free to ignore, but, you can use: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat to formulate duration/elapsed time in a more human-friendly manner (begone multiplication! Begone! 😂 )

Same apply to the `session.ts` file ;)

louis-bompart (commented):
I would avoid changing the `global` in testing, I think it'll bite us in the arse later.
Other than that, just had a few proposals, but nothing that I feel is important :)

packages/cli/src/hooks/analytics/session.spec.ts
y-lakhdar: Nice, I did not know about that... but I don't see how I can use RelativeTimeFormat to perform compare dates

louis-bompart: Hm you right, there'd be ways by playing with the rest of Intl... But that's not the goal of this library thus we may hit hiccups. Ignore this thread 😂

packages/cli/src/hooks/analytics/session.spec.ts
y-lakhdar: Newbedev says
> As every test suite run its own environment, you can mock globals by just overwriting them. All global variables can be accessed by the global namespace [...] The overwrite has only effects in your current test and will not effect others.

If it's coming from Newbedev, it's necessarily true :P

louis-bompart: regardless, the `session` class, from a unit test standpoint should not care about the intricacy of the `Config` class.
This is why it's mocked in all the other unit test suites ;)

packages/cli/src/hooks/analytics/session.spec.ts
y-lakhdar: No. Just wanted to use the date during which Skynet became self-aware.

louis-bompart (approved):

PR #540, opened 2021-11-10
chore: track invalid commands

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-663

## Proposed changes

Track commands that are invalid...

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/lib/decorators/preconditions/trackable.ts
```diff
@@ -68,7 +68,7 @@ function getEventName(target: Command) {
   return target.id!.replace(/:/g, ' ');
 }
 
-function buildEvent(
+export function buildEvent(
```
louis-bompart, line 71: I would extract this method and `errorIdentifier` in another file (e.g. `src/lib/analytics/eventUtils.ts`)

Like this we would keep a good import tree (i.e. I do not need a precondition/decorator for a Oclif hook)

packages/cli/src/hooks/commandNotFound/commandNotFound.ts
```diff
@@ -0,0 +1,10 @@
+import {Hook} from '@oclif/config';
+import {buildEvent} from '../analytics/eventUtils';
+
+const hook: Hook<'command_not_found'> = async function (options) {
+  this.config.runHook('analytics', {
+    event: buildEvent('invalid command', {command: options.id}),
```
y-lakhdar, line 6: @tcamire, here is an event name that does not really fit the proposed syntax . Do you have any recommendation to flag invalid commands?

y-lakhdar: That works with me.

louis-bompart (approved):

louis-bompart (approved):

PR #541, opened 2021-11-11
chore: track command interruption

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-661

## Proposed changes

Track whenever an interruption signal is detected

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/hooks/init/termination-signals.ts
```diff
@@ -0,0 +1,16 @@
...
+  signals.forEach((signal) => process.on(signal, () => trackEvent(signal)));
+};
+
+function trackEvent(termination_signal: string) {
+  config.runHook('analytics', {
+    event: buildEvent('interrupted operation', {termination_signal}),
+  });
+
+  // eslint-disable-next-line no-process-exit
+  process.exit(0);
```
louis-bompart, line 15: In case of termination, the exitCode should be `128+n` with n being the ' Signal numbering for standard signals'
Luckily, the second args of the signal event is `n`.

See:
 - https://nodejs.org/api/process.html#exit-codes
 - https://man7.org/linux/man-pages/man7/signal.7.html

louis-bompart (changes requested):
Exit with proper signal ;)

packages/cli/src/hooks/init/termination-signals.ts
```diff
@@ -0,0 +1,20 @@
...
+  );
+};
+
+function trackEvent(termination_signal: string, exitCode: number) {
+  config.runHook('analytics', {
+    event: buildEvent('interrupted operation', {termination_signal}),
+  });
+
+  // eslint-disable-next-line no-process-exit
+  process.exit(128 + exitCode);
```
louis-bompart, line 19: add some comment w/ doc ref to explain what's that, it's quite esoteric otherwise 😂

louis-bompart (approved):
LGTM!

PR #543, opened 2021-11-12
chore: sanitize data sent to Amplitude

y-lakhdar:
### Follow convention for Amplitude event ([Taxonomy Definition](https://coveord.atlassian.net/wiki/spaces/RD/pages/2855141440/New+Taxonomy+Definition))
* User properties should be in snake case (not camel case)
* Manually specify event names for commands that do not follow Taxonomy syntax.

### Sanitize events
* Remove flags and args from Amplitude events
* Do not log error messages. Simply the error name should be sent. That is why all errors now extend the CLIBaseError.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/hooks/analytics/eventUtils.ts
```diff
@@ -19,35 +25,22 @@ export function buildEvent(
...
 export function buildError(arg: unknown) {
-  /**
-   * TODO: CDX-660: Make sure to remove any PII from the Error object.
-   *       error.message could contain data that is not allowed to be tracked for non-Trial users
-   *       example: orgID, sourceID, ...
-   */
-  if (arg instanceof Error) {
-    return arg;
-  }
+  const isErrorFromAPI = validate(arg, APIErrorSchema).valid;
```
louis-bompart, line 28: Extract this in a separate function:
```ts
  function isErrorFromAPI(arg: unknown): arg is APIErrorResponse {
    return validate(arg, APIErrorSchema).valid;
  }
```

Then, l31 instead of `if(isErrorFromAPI) { return new APIError(arg as APIErrorResponse); }` do:
`if(isErrorFromAPI(arg)) { return new APIError(arg); }`

packages/cli/src/lib/errors/CLIBaseError.ts
```diff
@@ -0,0 +1,6 @@
+export class CLIBaseError extends Error {
+  public name = 'CLI Error';
+  public constructor(message?: string) {
+    super(message);
+  }
+}
```
louis-bompart, line 6: I think you should make this an abstract class ;)
```suggestion
export abstract class CLIBaseError extends Error {
  public name = 'CLI Error';
  public constructor(message?: string) {
    super(message);
  }
}
```

packages/cli/src/lib/utils/string.ts
```diff
@@ -0,0 +1,3 @@
+export function camelToSnakeCase(str: string) {
+  return str.replace(/[A-Z]/g, (letter) => `_${letter.toLowerCase()}`);
+}
```
louis-bompart, line 3: https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABBAhgWwKYBsAqcDKYKA1hgMIoDOGAFJVAE4Bci9DMYA5gJSIDeAKESIGGKCAZI2AOlEAHLCgi0A9AG0AggFoAWgF0VnADSIaWMVAwNeAXgB8iAAYB9ACR9zUSw2lQ4AGTgAdysKahpuAF9HbgBuAUiBAQgESjhzaSw4ThpUTFwCIlIw2gByAAUUKCqMUu5uIA

Just to prevent user-derp, I'd do:

```suggestion
export function camelToSnakeCase(str: string) {
  if(/^[A-Z]/.test(str)) {
    throw `'${str}' is not a camelCase string.`;
  }
  return str.replace(/[A-Z]/g, (letter) => `_${letter.toLowerCase()}`);
}
```

louis-bompart (approved):
LGTM overall, some syntax sugar here and there that's all ;)

packages/cli/src/lib/decorators/preconditions/trackable.ts
```diff
@@ -1,25 +1,46 @@
 import type Command from '@oclif/command';
 import {buildError, buildEvent} from '../../../hooks/analytics/eventUtils';
 
-export function Trackable(overrideEventProperties?: Record<string, unknown>) {
+export interface trackableOptions {
```
olamothe, line 4:
```suggestion
export interface TrackableOptions {
```

olamothe (approved):

packages/cli/src/lib/utils/string.ts
y-lakhdar: Instead of throwing an error, why not removing leading and trailing underscores.
This function is used before sending analytic events. So throwing errors is not a good idea.

packages/cli/src/lib/decorators/preconditions/trackable.ts
```diff
@@ -1,25 +1,46 @@
 import type Command from '@oclif/command';
 import {buildError, buildEvent} from '../../../hooks/analytics/eventUtils';
 
-export function Trackable(overrideEventProperties?: Record<string, unknown>) {
+export interface TrackableOptions {
+  /**
+   * Event name used to identify the command.
+   * The expression should go as follows: **(Subject) (Process)?**
```
louis-bompart, line 7: Why the question mark at (Process)?

packages/cli/src/lib/decorators/preconditions/trackable.ts
```diff
@@ -1,25 +1,46 @@
...
-    target: Command,
+    _target: Command,
     _propertyKey: string,
     descriptor: TypedPropertyDescriptor<() => Promise<void>>
   ) {
     const originalCommand = descriptor.value!;
     descriptor.value = async function (this: Command, ...cmdArgs: unknown[]) {
-      const name = getEventName(this);
-      const {flags, args} = this.parse(target.ctor);
+      const name = eventName || getEventName(this);
```
louis-bompart, line 36: eventName should never be an empty string then, correct?

packages/cli/src/lib/decorators/preconditions/trackable.ts
y-lakhdar: Because it is optional

packages/cli/src/lib/decorators/preconditions/trackable.ts
y-lakhdar: good catch

packages/cli/src/lib/decorators/preconditions/trackable.ts
louis-bompart: 👍 use the JSDOC standard for 'optionality' then: https://jsdoc.app/tags-param.html#optional-parameters-and-default-values
```suggestion
   * The expression should go as follows: **(Subject) [Process]?**
```

PR #544, opened 2021-11-12
feat: display config after `config:set`

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-688

## Proposed changes

Setting the config with `config:set` should return the resulting config like if a get was done right after.

 

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/commands/config/set.ts
```diff
@@ -64,4 +66,25 @@ export default class Set extends Command {
       );
     }
   }
+
+  private async displayConfig(cfg: Config) {
```
olamothe, line 70: There should be a shared function between `config/get.ts` and `config/set.ts` so that we always display the same thing for both.

y-lakhdar: Should we display the access token with the `config:get`, `config:set`, both or neither?

louis-bompart: If request explicitly with `get` yes (i.e. `coveo config:get access_token` (I don't recall exactly the syntax but well, you get the gist of it :P )

In other cases, we shall not.

louis-bompart (commented):
refacto for displaying, re-request review once done ;)

packages/cli/src/lib/config/configRenderer.ts
```diff
@@ -0,0 +1,22 @@
...
+      .filter((key) => keysToDisplay.includes(key as keyof BaseConfiguration))
+      .reduce(
+        (obj, key) => ({
+          ...obj,
+          [key]: configuration[key],
+        }),
+        {}
+      );
+
+    cli.styledJSON(allowedConfig);
```
y-lakhdar, line 20: Juste pour toi @louis-bompart  :)

packages/cli/src/lib/config/configRenderer.ts
```diff
@@ -0,0 +1,22 @@
+import {cli} from 'cli-ux';
```
y-lakhdar, line 1: Eventually we could have different method to render the config (similar to sfdx).
Meaning we could have JSON and table formats.

packages/cli/src/lib/config/configRenderer.ts
louis-bompart: nice

louis-bompart (approved):

PR #546, opened 2021-11-12
feat: make atomic command pass arguments to @coveo/create-atomic

ThibodeauJF:
## Proposed changes
Added prompts so the @coveo/create-atomic works better standalone. This removes the need for the env variable script, ass plop does all the work for us.

## Testing

Tests will be in an upcoming PR

https://coveord.atlassian.net/browse/CDX-670


packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -60,17 +62,44 @@ export default class Atomic extends Command {
...
+      cfg.organization,
+      '--api-key',
+      apiKey.value!,
+      '--platform-url',
+      platformUrl({environment: cfg.environment}),
+      '--user',
+      userInfo.providerUsername,
+    ];
+
+    return spawnProcess(appendCmdIfWindows`npx`, cliArgs);
```
ThibodeauJF, line 86: What's the best way of testing this locally, by using verdaccio? I was able to publish locally but I'm getting syntax errors i'm not usually seeing by running `node index.mjs ...args` or `npm start ...args`

y-lakhdar: Yes, Verdaccio is the best way of testing. Do you mean you get errors when running the package as a standalone or via the CLI?

y-lakhdar (approved):

louis-bompart (commented):
LGTM, I'd just want to be sure you've been able to test before approving ;)
Request review when you're done, or DM me if you need help with that

packages/cli/src/commands/ui/create/atomic.ts
ThibodeauJF: Fixed!

packages/create-atomic/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@coveo/create-atomic",
-  "private": false,
```
louis-bompart: :D

louis-bompart (approved):
Smooth 👍

PR #547, opened 2021-11-15
feat: add colors to console logs

y-lakhdar:
### Updated login message visual
before:
image

After:
image

### Created a color-utils
Made sure all commands to execute are colorized using the same style.
If that is fine with you I would like to replace all imports of `chalk` with `color-utils`. That was, we keep a level of consistency across the commands.

### Question:
Should colors-utils should have its own repo?

 

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/lib/utils/color-utils.ts
```diff
@@ -0,0 +1,23 @@
...
+  cmd: chalk.blueBright.bold,
+};
+
+type availableColors = typeof customColors | typeof chalk;
+type augmentedChalk = typeof customColors & typeof chalk;
+
+export const color: typeof customColors & typeof chalk = new Proxy(chalk, {
+  get: (chalk, name: keyof availableColors) =>
+    customColors[name] ? customColors[name] : chalk[name],
+}) as augmentedChalk;
```
louis-bompart, line 21: I'm really not fond of proxies. They tend to obfuscate logs and were often a smell of hacks in my past experiences.
I understand you want to 'augment' chalk, but I think there's a lot of more maintainable design patterns to consider before going the proxy way.

louis-bompart (changes requested):
Proxies 😬

packages/cli/src/lib/utils/color-utils.ts
y-lakhdar: I understand. So far, `Proxy` is the only pattern that worked for me.
I tried simply augmenting the object like `{...chalk, ...customColors}` but it causes some run time errors.

If you think Proxy should absolutely be avoided, I can try figuring out a cleaner approach.

louis-bompart:
After investigation, we should be able to use inheritance to extend chalk behaviour (we can see in the readme that there's a `new Chalk()` being called, so that means that there's a constructor to extend, I think 😅 ).
However, there is an ongoing issue with the typing: https://github.com/chalk/chalk/issues/519 which may hold us off.

louis-bompart:
```ts
import chalk, { Chalk } from "chalk";

interface Colors extends Chalk, CustomColors {}

interface CustomColors {
  header: Chalk;
  error: Chalk;
  cmd: Chalk;
}

const customColorFactory = (chalk: Chalk): CustomColors => ({
  header: chalk.bold.hex("#1CEBCF"),
  error: chalk.bgHex("#F64D64").hex("#272C3A"),
  cmd: chalk.blueBright.bold,
});

const colorFactory = (): Colors => {
  const chalkInstance = new chalk.Instance();
  const customColors = customColorFactory(chalkInstance);
  return Object.assign(chalkInstance, customColors);
};

const colors = colorFactory();

console.log(colors.cmd`color`);
console.log(colors.bold`color`);
```

Using `Object.assign` we can build the augmented chalk

y-lakhdar:
Closing for now

PR #549, opened 2021-11-16
fix: flush analytic events before terminating the process

y-lakhdar:
## Context

Some analytics events were missing from the Amplitude Dashboard.
Turns out, processes from CLI commands were terminated before all UA events were fired.

I saw 2 options to send all queued analytic events before the process exit:
1. Log every UA event synchronously.
2. Stack UA events and consume them asynchronously.

The first option, was discarded because every logged event was causing a ~0.5 second delay (which was noticeable).

## Proposed changes

Stack all the UA events and make sure they are flushed before the end of the process. There are 3 scenarios that will lead to the event flushing:
1. after the command completes (`postrun` hook)
2. after the command fails (`@Trackable` decorator)
3. after the a unexpected process exit (`termination-signal`)

## Other enhancement
I also reduced the number of calls to the [Identify API](https://developers.amplitude.com/docs/identify-api). There is no need to call the the Identify API more than once per command (since the user identity will not change).

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/hooks/analytics/amplitudeClient.ts
```diff
@@ -0,0 +1,11 @@
...
+
+// TODO: CDX-656: replace with Production API key on build
+const analyticsAPIKey = '2b06992f1a80d36396ba7297a8daf913';
+
+export interface AmplitudeClient extends NodeClient {
+  identified: boolean;
+}
+
+// TODO: CDX-667: support proxy
+export const amplitudeClient = init(analyticsAPIKey);
```
y-lakhdar, line 11: The `amplitudeClient` needs to be a singleton to stack and flush analytic events.

packages/cli/src/hooks/analytics/analytics.ts
```diff
@@ -1,32 +1,52 @@
...
+  if (!(await canLogEvent())) {
+    return;
+  }
+
+  const {userId, deviceId, identify} = await new Identifier().getIdentity();
+  if (options.identify) {
+    amplitudeClient.identify(userId, deviceId, identify);
+  }
+
+  amplitudeClient.logEvent({
```
y-lakhdar, line 28: `amplitudeClient.identify()` and `amplitudeClient.logEvent()` will create and add an event to the amplitudeClient queued.

Notice the absence of the `await` operator.

Later, we do a `await amplitudeClient.flush()` to ensure all queued event are sent to Amplitude.
This is the only `amplitudeClient` method to be used with an `await`.

packages/cli/src/hooks/commandNotFound/commandNotFound.ts
```diff
@@ -2,7 +2,7 @@ import {Hook} from '@oclif/config';
 import {buildEvent} from '../analytics/eventUtils';
 
 const hook: Hook<'command_not_found'> = async function (options) {
-  this.config.runHook('analytics', {
+  await this.config.runHook('analytics', {
```
y-lakhdar, line 5: Hooks should be awaitable. But not the logging of the events.

packages/cli/src/hooks/postrun/postrun.ts
```diff
@@ -0,0 +1,8 @@
+import {Hook} from '@oclif/config';
+import {flush} from '../analytics/analytics';
+
+const hook: Hook<'postrun'> = async function (_options) {
+  await flush();
```
y-lakhdar, line 5: Send any events that are currently queued for sending.

packages/cli/src/hooks/init/termination-signals.ts
```diff
@@ -1,20 +1,29 @@
...
 export const exitSignals = ['SIGINT', 'SIGTERM', 'SIGQUIT', 'SIGHUP'] as const;
 
-export const handleTerminationSignals = () => {
+export function handleTerminationSignals() {
   exitSignals.forEach((signal) =>
-    process.on(signal, (_sig: string, code: number) => {
-      trackEvent(signal, code);
+    process.on(signal, async (_sig: string, exitCode: number) => {
+      await trackEvent(signal);
+      await flush();
```
y-lakhdar, line 10: Sending queued events whenever an exit signal is intercepted.

packages/cli/src/lib/decorators/preconditions/trackable.ts
```diff
@@ -71,12 +73,13 @@ async function trackError(
...
-    this.config.runHook('analytics', {
-      event: buildEvent('received error', properties, buildError(arg)),
+  for (let i = 0; i < args.length; i++) {
+    await this.config.runHook('analytics', {
+      event: buildEvent('received error', properties, buildError(args[i])),
     });
-  });
+  }
 
+  await flush();
```
y-lakhdar, line 82: Sending queued events whenever the command fails

packages/cli/src/hooks/analytics/identifier.spec.ts
```diff
@@ -120,26 +107,24 @@ describe('identifier', () => {
...
-    it('should not identify event with (un-hashed) email', () => {
-      const userIdCheck = expect.stringMatching(/^(?!bob@.*?\.com).*/);
-      expect(mockAmplitudeIdentify).toHaveBeenCalledWith(
-        userIdCheck,
-        expect.anything(),
-        expect.anything()
-      );
+    it('should not identify event with (un-hashed) email', async () => {
+      const identity = await new Identifier().getIdentity();
+      expect(identity.userId).not.toMatch(/^bob@.*?\.com$.*/);
```
louis-bompart, line 122: Why the `.*` at the end of the regexp?

louis-bompart, line 122: Why the `.*` at the end of the regexp?

packages/cli/src/hooks/analytics/identifier.spec.ts
```diff
@@ -120,26 +107,24 @@ describe('identifier', () => {
       expect(mockSetIdentity).toHaveBeenCalledWith('region', 'us');
     });
 
-    it('should set is_internal_user to true', () => {
+    it('should set the user ID', async () => {
+      const identity = await new Identifier().getIdentity();
```
louis-bompart, line 111: I'd refactor identity in the describe scope, and do a beforeEach with the assignation to `new Identifier().getIdentity();`, given that you do it for every test afaik

packages/cli/src/hooks/init/termination-signals.ts
```diff
@@ -1,20 +1,29 @@
...
+      /**
+       * Signal Exits: If Node.js receives a fatal signal such as SIGTERM or SIGHUP, then its exit code will be 128 plus the value of the signal code.
+       * This is a standard POSIX practice, since exit codes are defined to be 7-bit integers, and signal exits set the high-order bit,
+       * and then contain the value of the signal code.
+       *
+       * For example, signal SIGABRT has value 6, so the expected exit code will be 128 + 6, or 134.
+       * For more info, visit https://nodejs.org/api/process.html#exit-codes
+       */
+      // eslint-disable-next-line no-process-exit
+      process.exit(128 + exitCode);
```
louis-bompart, line 20: We talked a bit about that, I'm wondering, does oclif has its own handler that exit the process?
In which case we should do `process.exitCode = 128+exitCode` and let it do its thing

(IMO, try the snippet ☝️ . If it works, keep the snippet, otherwise, keep the process.exit())

packages/cli/src/lib/decorators/preconditions/trackable.ts
```diff
@@ -71,12 +73,13 @@ async function trackError(
   originalCatchCommand: (...args: unknown[]) => Promise<void>,
   args: unknown[]
 ) {
-  args.forEach((arg) => {
-    this.config.runHook('analytics', {
-      event: buildEvent('received error', properties, buildError(arg)),
+  for (let i = 0; i < args.length; i++) {
+    await this.config.runHook('analytics', {
+      event: buildEvent('received error', properties, buildError(args[i])),
```
louis-bompart, line 78:
```suggestion
  for (const arg of args) {
    await this.config.runHook('analytics', {
      event: buildEvent('received error', properties, buildError(arg)),
```

louis-bompart (approved):
LGTM 👍 Small suggestions

packages/cli/src/hooks/init/termination-signals.ts
y-lakhdar: I tried 2 approaches

`process.exitCode = 128 + exitCode;`
Does nothing since nothing is asking the process to exit.

`cli.exit(128 + exitCode);`
Built-in method in [cli-ux](https://www.npmjs.com/package/cli-ux) to exit the process. However, it also throws an [ExitError](https://github.com/oclif/errors/blob/95b0ea57bb23d0c4fcb0d877260418387e940492/src/errors/exit.ts). But nothing is intercepting it... so we don't want that.

Oclif recommends throwing an error rather than exiting the process. However, in this case, I think it's too late to throw an error.

olamothe (approved):

olamothe:
Good PR with some nice explanation for context, helps understanding quite a lot !

packages/cli/src/hooks/init/termination-signals.ts
louis-bompart: 👍 all good, process.exit is then the way to go.

PR #554, opened 2021-11-18
fix: erase resources folder content before pulling resources

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-695

## Proposed changes

Simply reset the resources/ folder before doing another `org:resources:pull`.
 
## Testing

- [ ] Unit Tests:

- [x] E2E Tests:

- [x] Manual Tests:


packages/cli/src/commands/org/resources/pull.ts
```diff
@@ -59,6 +59,11 @@ export default class Pull extends Command {
       default: true,
       allowNo: true,
     }),
+    override: flags.boolean({
```
louis-bompart, line 62: I think you're misusing the word override here.
It should be overwrite I think :)

packages/cli/src/commands/org/resources/pull.ts
```diff
@@ -115,6 +122,21 @@ export default class Pull extends Command {
     await project.refresh(snapshotBlob);
   }
 
+  private async ensureProjectReset(project: Project) {
+    const {flags} = this.parse(Pull);
+    if (!flags.override && project.contains(Project.resourceFolderName)) {
+      const override =
+        await cli.confirm(dedent`There is already a Coveo project with resources in it.
+        This command will override the ${Project.resourceFolderName} folder content, do you want to proceed? (y/n)`);
```
louis-bompart, line 130: Same here, overwrite I think?

louis-bompart (approved):

louis-bompart (commented):
Not sure about the use of `override` in the wording.

olamothe (approved):

packages/cli/src/commands/org/resources/pull.ts
olamothe: Yep !

y-lakhdar: Oups. thanks

PR #555, opened 2021-11-19
fix(source:push): only parse valid JSON files

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-707

## Proposed changes

Running `source:push:add <SOURCE_ID> -d <FOLDER>`, will throw an error if the folder contains a non-JSON file.
This error can be really hard to troubleshoot. Specially if the folder contains dot files.
The returned error is simply, Uncaught SyntaxError: Unexpected number in JSON at position 0.

### Fixes:

1. Skip dot files
2. Provide a better error message for invalid files
3. Add names to File error for Amplitude tracking

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/commands/source/push/add.ts
```diff
@@ -95,13 +96,15 @@ export default class SourcePushAdd extends Command {
...
         return Promise.all(
-          readdirSync(folder).flatMap(async (file) => {
-            const fullPath = path.join(folder, file);
-            const docBuilders =
-              parseAndGetDocumentBuilderFromJSONDocument(fullPath);
-            this.successMessageOnParseFile(fullPath, docBuilders.length);
-            await send(docBuilders);
-          })
+          readdirSync(folder)
+            .filter((file) => !isDotFile(file))
```
olamothe, line 100: Instead of `.filter((file) => !isDotFile(file))`, how about `.filter((file) => isJsonFile(file))` ?

olamothe (approved):

packages/cli/src/commands/source/push/add.ts
y-lakhdar: I though it would be useful if the user had saved JSON data using a different filetype extension (e.g. JSON in a txt file)...

I don't know if it will happen and if it might create confusion... Anyhow, we can simply accept `.json` files.

olamothe: I would be tempted to say that since the whole command description is about pushing JSON file... it's fair to say that it's a no brainer that only `.json` file would be processed, and that it should not cause confusion.

louis-bompart (changes requested):
Can you add some unit tests for that?

packages/cli/src/lib/utils/file.ts
```diff
@@ -0,0 +1,5 @@
+import {PathLike} from 'fs';
```
louis-bompart, line 1:
```suggestion
import type {PathLike} from 'fs';
```

packages/cli/src/lib/push/parseFile.ts
```diff
@@ -15,16 +15,18 @@ import {
...
   documentPath: PathLike
 ) => {
-  const isAFile = isFile(documentPath);
-  if (!isAFile) {
-    throw new NotAFileError(documentPath);
-  }
+  ensureFileIntegrity(documentPath);
 
   const fileContent = JSON.parse(readFileSync(documentPath).toString()) as
     | Record<string, string>
```
louis-bompart, line 32: Alright, here, it's not great perf-wise: We're reading the file twice because of the `isValidJsonFile`.
The cost will be non-negligible when facing big chunky JSON.

So, I would remove the isValidJSONFile method altogether and trycatch the `filecontent` var assignment instead.
I would also throw a NotAJsonFileError only and only if this is actually what's causing the program to throw (i.e. checks on the `e` of the catch

louis-bompart (changes requested):
Parse only once you shall ;)

packages/cli/src/lib/push/parseFile.ts
y-lakhdar: Good point. However, instead of checking on the `e` of the catch, why not create a method (e.g. `safeParse`) that would only parse the JSON.

That way, we are sure any error throw by this method is actually a `NotAJsonFileError`

packages/cli/src/lib/push/validatorErrors.ts
```diff
@@ -2,18 +2,27 @@ import {PathLike} from 'fs';
...
 
 export class NotAFileError extends CLIBaseError {
+  public name = 'Not A File Error';
   public constructor(p: PathLike) {
     super(`${p} is not a valid file, or does not exists.`);
   }
 }
+export class NotAJsonFileError extends CLIBaseError {
+  public name = 'Not A JSON File Error';
+  public constructor(p: PathLike) {
```
louis-bompart, line 12: I was checking what pathlike is, why not just a string really?.
`type PathLike = string | URL | Buffer` iirc. I don't think a user will input a URL object (it can't to be exact)

y-lakhdar: I used it to stay consistent with other Validation Errors. I think the same logic would apply to other errors as well

louis-bompart (approved):

PR #556, opened 2021-11-19
fix(source:push): handle `ErrorFromAPI` as well as other error types

y-lakhdar:
## Proposed changes

We cannot assume the error will always be an Error from the API. The error can be coming from Axios (which is the case if you try to push a very large document)

This PR by itself does not bring a lot of value, but it is the foundation for the next [bug](https://coveord.atlassian.net/browse/CDX-709) which can be a blocker for a lot of users.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/commands/source/push/add.ts
```diff
@@ -221,7 +219,7 @@ export default class SourcePushAdd extends Command {
       });
       this.successMessageOnAdd(fileNames, batch.length, res);
     } catch (e) {
-      this.errorMessageOnAdd(e as ErrorFromAPI);
+      this.errorMessageOnAdd(e as Error);
```
y-lakhdar, line 222: This is the main reason of this PR. Anything can be intercepted here

packages/cli/src/lib/errors/APIError.ts
```diff
@@ -1,22 +1,89 @@
...
   message: string;
   errorCode: string;
   requestID: string;
 }
 
+export interface AxiosErrorFromAPI {
+  response: {
+    status: number;
+    data: APIErrorResponse;
+  };
```
y-lakhdar, line 15: I took the opportunity to link the `APIErrorResponse` (returned when using the platform client) and `AxiosErrorFromAPI`.

olamothe (approved):

louis-bompart (approved):

packages/cli/src/lib/errors/APIError.ts
```diff
@@ -1,22 +1,89 @@
...
+      this.message += dedent`
+
+      Status code: ${status}
+      `;
+    }
+    this.message += dedent`
+
+    Error code: ${red(errorCode)}
+    Message: ${red(message)}
+    `;
```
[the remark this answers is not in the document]
y-lakhdar: thanks

PR #557, opened 2021-11-19
fix(source:push): support pushing large documents 

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-710

## Proposed changes

Small change, big impact!

If a document size was greater than 5MB (max content length), an empty batch was uploaded to the API.
This was creating a significant amount of "empty batch" requests (see image).

### Example
Here, I am pushing 2 documents of 9 MB each.

#### Before change
Since both documents exceed the maxContentLength limit, the `accumulator` is always empty. Therefore, 2 initial (empty) batches are sent by the `send` method. 
Finally, the `close` method upload another batch with only one document
image

#### After change
We do not want to upload a batch if it is empty
image

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


olamothe (approved):

louis-bompart (changes requested):
Can we add tests to cover this use case?

packages/cli/src/commands/source/push/add.spec.ts
```diff
@@ -170,6 +176,26 @@ describe('source:push:add', () => {
       expect(ctx.stdout).toContain('Status code: 202 👌');
     });
 
+  test
+    .do(() => {
+      const largeFileSize = 5 * 1024 * 1024;
```
louis-bompart, line 181: So 5 mebibytes ?
If you mean 5 megabytes, it's 5e6

louis-bompart (approved):
LGTM, double-check if the size limits are in metric megabytes or in legacy megabytes aka mebibytes and adapt the tests accordingly

packages/cli/src/commands/source/push/add.spec.ts
y-lakhdar: nice

PR #559, opened 2021-11-23
fix: limit concurrent promises on `source:push`

y-lakhdar:
## Proposed changes

Using an iterator to limit the number of concurrent promises.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/commands/source/push/add.ts
```diff
@@ -41,6 +42,13 @@ export default class SourcePushAdd extends Command {
       description:
         'One or multiple folder containing json files. Can be repeated',
     }),
+    maxConcurrent: flags.integer({
+      exclusive: ['file'],
+      char: 'c',
+      default: 10,
+      description:
+        'The maximum number of requests to send concurrently. Increasing this value will increase the speed at which documents are pushed to the Coveo platform. However, if you run into memory or throttling issues, consider reducing this value.',
```
y-lakhdar, line 50: @JPLachance does that make sense?

packages/cli/src/lib/utils/iterator.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+  }
+
+  public async resolve() {
+    this._resolve();
+    return this.promise;
+  }
+}
+
+describe('iterator', () => {
```
y-lakhdar, line 19: That's the best/easiest way to test the promise concurrency...
I am not able to test the concurrency via the `source:push:add` command since I did not find any way to run any `expect` while there are still pending promises going on.

packages/cli/src/commands/source/push/add.ts
jpmarceau, line 50:
```suggestion
        'The maximum number of requests to send concurrently. Increasing this value increases the speed at which documents are pushed to the Coveo platform. However, if you run into memory or throttling issues, consider reducing this value.',
```
Looks good to me 👍 
(I suspect you were speaking to me rather than JPLachance 😛 )

jpmarceau (approved):

packages/cli/src/lib/utils/iterator.spec.ts
louis-bompart: Nice one! You isolated the behavior to test, this should offer 'good enough' coverage ;)

louis-bompart (approved):

packages/cli/src/commands/source/push/add.ts
y-lakhdar: oups 🤦  sorry

PR #562, opened 2021-11-25
feat: use search-token-lambda with generated Atomic project & add deploy command

ThibodeauJF:
## Proposed changes

Replaces the search-token-server by search-token-lambda & leverages the netlify cli

Next step will be to document and publish the new cli command and create-atomic package!

## Testing

e2e tests are my next PR

https://coveord.atlassian.net/browse/CDX-669


packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -53,7 +53,6 @@ export default class Atomic extends Command {
   )
   public async run() {
     await this.createProject();
-    this.displayFeedbackAfterSuccess();
```
ThibodeauJF, line 56: handled in plop directly, which is better since it will also display the message standalone

packages/create-atomic/src/plopfile.ts
```diff
@@ -72,15 +72,42 @@ export default function (plop: NodePlopAPI) {
...
       },
+      function installPackagesPrompt() {
+        return 'Installing packages 🚀';
+      },
       function installPackages(answers) {
-        const {project} = answers as PromptsAnswers;
-        spawn(getPackageManager(), ['install'], {
-          stdio: 'inherit',
-          cwd: `${currentPath}/${project}/`,
+        return new Promise((resolve, reject) => {
```
ThibodeauJF, line 82: swtiched to a promise in order to wait until displaying the next message

packages/create-atomic/templates/.env.example
```diff
@@ -1,33 +1,24 @@
 # The Plaform URL to use.
 # See https://docs.coveo.com/en/2976/coveo-solutions/deployment-regions-and-strategies
-ATOMIC_APP_PLATFORM_URL=https://platform.cloud.coveo.com
+PLATFORM_URL=https://platform.cloud.coveo.com
```
ThibodeauJF, line 3: Removing the ATOMIC_APP_ and copying the variables seemed like an extra, a bit useless, step?

packages/create-atomic/templates/.env.example
```diff
@@ -1,33 +1,24 @@
...
+USER_EMAIL=<YOUR_USER_EMAIL>
 
 # The endpoint to the server returning Coveo search tokens.
-# Example: ATOMIC_APP_TOKEN_ENDPOINT=https://acme.com/token
+# Example: TOKEN_ENDPOINT=https://acme.com/token
 # If not specified, the application will fetch the search token
 # from the local server within this project
-# ATOMIC_APP_TOKEN_ENDPOINT=<YOUR_ENDPOINT>
-
-# Port to the search token server
```
ThibodeauJF, line 26: Port is handled by the netlify dev server

packages/create-atomic/src/plopfile.ts
louis-bompart: Ideally, it'd be nice to have a little spinny boi or progression bar (probably the former because we don't have a good measure of progression) to indicate we didn't crash.

packages/create-atomic/templates/package.json
```diff
@@ -3,29 +3,29 @@
...
   "devDependencies": {
-    "concurrently": "6.4.0",
     "css-loader": "^6.5.1",
     "dotenv-webpack": "^7.0.3",
     "fs-extra": "10.0.0",
     "gts": "^3.1.0",
     "html-loader": "^3.0.1",
     "html-webpack-plugin": "^5.5.0",
     "mini-css-extract-plugin": "^2.4.4",
+    "netlify-cli": "^6.14.25",
```
louis-bompart, line 28: Ensure to be compatible with latest version, I reckon we're at 8 now

louis-bompart (approved):

ThibodeauJF:
Example live site https://atomic-site.netlify.app/

olamothe (approved):

PR #563, opened 2021-11-26
Oclif 2

louis-bompart:
A big one, the move to Oclif core. PR is here for showcases purposes for now.

# Minimal work left:
 - Fix `handleSnapshotError` method
 
# Would be nice:
 
## Some home-made syntax sugar:
 `cli-ux` table are a 🍞 with their `Record<string, any>` shenanigans, same things is probably the case for the catch.
 
 Investigate a 'decoratype' to make the pieces fit more nicely

 # Future works:
 
  - Compile as ESM (probably?)
  - Update dependency we locked down because of that.

### DIFF
--- a/packages/cli/bin/run
+++ b/packages/cli/bin/run
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
-require('@oclif/command').run()
-.then(require('@oclif/command/flush'))
-.catch(require('@oclif/errors/handle'))
+require('@oclif/core')
+  .run()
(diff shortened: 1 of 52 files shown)


olamothe (commented):
I've skimmed through the draft because it's a chunky one, but I'm happy to see that it's mostly "superficial" stuff (like changing some types and import), and that the core of how a command is built still stays the same 👍

y-lakhdar:
Replaced by #650

PR #573, opened 2021-12-01
chore: migrate to Plop v3

ThibodeauJF:
https://coveord.atlassian.net/browse/CDX-716


packages/create-atomic/src/plopfile.ts
```diff
@@ -1,6 +1,6 @@
 import {NodePlopAPI} from 'plop';
 import {spawn} from 'child_process';
-import {getPackageManager} from './utils';
+import {getPackageManager} from './utils.js';
```
ThibodeauJF, line 3: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm
https://github.com/microsoft/TypeScript/issues/41887#issuecomment-741504718

packages/create-atomic/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "module": "commonjs",
+    "module": "ES2020",
```
louis-bompart, line 3: Can you try with `nodenext` ? would require typescript 4.5 but it should be the best way now I reckon.

https://www.typescriptlang.org/docs/handbook/esm-node.html

packages/create-atomic/index.js
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env node
```
louis-bompart, line 1: Why do you need the shebang? 🤔

louis-bompart (commented):
iirc, sindresorhus guide to ESM was made prior to TS 4.5 and Node 16, which bring 'real' support for ESM. Some tweaks here and there and we should be good to go

packages/create-atomic/index.js
```diff
@@ -0,0 +1,29 @@
...
+import path from 'node:path';
+import minimist from 'minimist';
+import {Plop, run} from 'plop';
+
+const args = process.argv.slice(2);
+const argv = minimist(args);
+
+import {dirname} from 'node:path';
+import {fileURLToPath} from 'node:url';
```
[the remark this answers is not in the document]
ThibodeauJF: This is straight from their docs 🤷

packages/create-atomic/index.js
ThibodeauJF: just copied from their docs too

packages/create-atomic/src/plopfile.ts
ThibodeauJF: it is in TS, but we need to set it properly to the output with a .js in order for it to work with ES2020, like the links indicate. I'll try with nodenext, though

packages/create-atomic/index.js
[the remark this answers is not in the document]
ThibodeauJF: https://github.com/plopjs/plop#wrapping-plop

packages/create-atomic/tsconfig.json
ThibodeauJF: You mean 4.6? Doesn't appear to work with 4.5. What's the issue with ES2020 ?

ThibodeauJF:
> Node 16

That's a harsh requirement, I wasn't on it until recently

packages/create-atomic/index.js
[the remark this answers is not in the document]
louis-bompart: Yeah I guess that's OK
https://nodejs.org/api/esm.html#esm_node_imports

packages/create-atomic/index.js
louis-bompart: I'm curious abt it, because shebang does not work on windows soo...

louis-bompart:
> > Node 16
> 
> That's a harsh requirement, I wasn't on it until recently

True, it is more stringent, however, because overall NodeJS and ECMA as a whole is turning its back on CJS and moving toward ESM modules, I think that I would prefer to limit support on Node 16 on the command, and not have to rework it next year

Overall, my two cents:
 - For existing and released features, we should maintain support on the latest Active and Maintenance LTS version of Node.js (so, for example, I'll drop support for Node 14 in 11 months, or just about 6 months before its official EoL)
 - For new features, we should support the Active LTS version of Node.js. Support of the Maintenance LTS is encouraged but is not mandatory.
 
This position is motivated by the rising number of packages that we cannot use because we're stuck with CJS module and engine. I prefer for us to be a step in advance than having to do the work twice.

packages/create-atomic/index.js
ThibodeauJF: Seems like a good thing to keep https://stackoverflow.com/a/33510581

This way, maybe when i install `npm i -g @coveo/create-atomic` and then run `create-atomic myproject` it will invoke node correctly 🤷  I haven't tested, just thinking out loud

PR #574, opened 2021-12-01
ci: get and use the latest tag commit

louis-bompart:
## Proposed changes

Currently, our CD steps in Jenkins are fetching the HEAD of our main branch, instead of the commit of the latest tag, this could lead to discrepancies.
This changes that.

## Testing

Hard to test outside Jenkins :/ 

-----
CDX-718

### DIFF
--- a/scripts/get-release-artifact-and-commit-sha.js
+++ b/scripts/get-release-artifact-and-commit-sha.js
@@ -7,7 +7,7 @@ async function main() {
   // This folder structure needs to be respected in order for the CLI update plugin to
   // be able to do it's job properly.
   const topLevelDirectory = './artifacts';
-  const subDirectoryForTarball = `${topLevelDirectory}/coveo-${tag}`;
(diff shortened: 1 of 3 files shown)


scripts/get-release-artifact-and-commit-sha.js
```diff
@@ -29,6 +29,8 @@ async function main() {
     }
   });
 
+  fs.writeFileSync('latest-commit', tag.commit.sha);
```
olamothe, line 32: Why use a file instead of stdout to pass that info around ? 🤔

olamothe (approved):

scripts/get-release-artifact-and-commit-sha.js
louis-bompart: Cause I'm unsure how to feed stdout to a groovy var (I ain't a very good groovy dancer let's say)

So.... Yeah. readFile groovy func looked nice.

PR #575, opened 2021-12-02
fix: update `deleteOlderThan` helpValue

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-713

## Proposed changes

The deleteOlderThan needs to be a timestamp in milli-seconds (see [API docs](https://platformdev.cloud.coveo.com/docs?urls.primaryName=PushAPI#/Item/delete_organizations__organizationId__sources__sourceId__documents_olderthan)).

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/commands/source/push/delete.ts
```diff
@@ -23,8 +23,8 @@ export default class SourcePushDelete extends Command {
       char: 'd',
       exclusive: ['delete'],
       description:
-        'If this flag is set, all items that have been added or updated in the source before the specified ISO 8601 date or Unix timestamp will be deleted. The documents will be deleted using the default queueDelay, meaning they will stay in the index for about 15 minutes after being marked for deletion.',
-      helpValue: '2000-01-01T00:00:00-06:00 OR 946702800',
+        'If this flag is set, all items that have been added or updated in the source before the specified ISO 8601 date or timestamp in milliseconds will be deleted. The documents will be deleted using the default queueDelay, meaning they will stay in the index for about 15 minutes after being marked for deletion.',
```
louis-bompart, line 26: Unsure about that, what motivates this change?

louis-bompart:
To be clear, I think we should keep the "Unix" thing, because a millisecond timestamp need a reference to start with.

packages/cli/src/commands/source/push/delete.ts
y-lakhdar: The `orderingId` should be a timestamp in milliseconds (see [swagger docs](https://platformdev.cloud.coveo.com/docs?urls.primaryName=PushAPI#/Item/delete_organizations__organizationId__sources__sourceId__documents_olderthan)). Otherwise it will not work.
@francistb can you confirm?

francistb: orderingId can be customer provided, it can be any number. 

our crawlers, and pushapi if you don't specify anything, use a timestamp in milliseconds (something like 1638541545341)

francistb: image
ratlab in dev for example of custom ordering

olamothe (approved):

packages/cli/src/commands/source/push/delete.ts
louis-bompart:
```suggestion
        'If this flag is set, all items that have been added or updated in the source before the specified ISO 8601 date or Unix timestamp in milliseconds will be deleted. The documents will be deleted using the default queueDelay, meaning they will stay in the index for about 15 minutes after being marked for deletion.',
```

Specify that it's a Unix timestamp. I know and agree that it's quite obvious and that Unix/Epoch is the ubiquitous one, but there's other reference point out-there, so let's be precise when we can without much fluff (which is the case here)

louis-bompart (approved):
put the word unix back in there and GG

PR #581, opened 2021-12-06
feat(atomic): add snyk coverage for atomic

louis-bompart:
## Proposed changes

 -  Updated all package-lock.json
 -  Tweaked create-atomic so we could cover its template without
    breaking stuff in it.

## Testing

- [x] Unit Tests: no logical changes.
- [x] Functionnal Tests: Covered by existing one I think (@ThibodeauJF , confirm?)
- [x] Manual Tests: ran `npm install` in package/create-atomic. both package-lock were generated.

-----
CDX-723


packages/create-atomic/script/npmInstallTemplate.ts
```diff
@@ -0,0 +1,24 @@
+import {spawn} from 'node:child_process';
+import {join} from 'node:path';
+import {cwd} from 'node:process';
+import {getPackageManager} from '../src/utils.js';
+
+(async () =>
+  await new Promise((resolve, reject) => {
+    const process = spawn(
+      getPackageManager(),
+      ['install', '--ignore-scripts'],
```
louis-bompart, line 10: The magic is in the `ignore-scripts`: it prevent the postinstall to run in the `template`, which prevent the deletion of its bits.

packages/create-atomic/templates/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "{{project}}",
   "description": "{{titleCase project}} project",
   "version": "0.1.0",
+  "private": "true",
```
ThibodeauJF, line 5: if you were a developper using this, would you want the project to be private by default? I'm wondering... Is that what we do for other generated projects?

ThibodeauJF (approved):

packages/create-atomic/templates/package.json
louis-bompart: I don't know if it's what we're doing today, but I do not think people would want to publish their project usually?

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -31,14 +32,20 @@ describe('ui:create:atomic', () => {
...
         .on('stdout')
         .do()
         .once(),
     ]);
+
+    await buildTerminal
+      .when(/\(y\)/)
+      .on('stderr')
+      .do(answerPrompt(`y${EOL}`))
+      .until(buildTerminalExitPromise);
```
louis-bompart, line 48: This is only triggered with the local setup because `npx` behave differently in a CI (it warn and install without prompt)

packages/cli-e2e/docker/Dockerfile
```diff
@@ -48,6 +50,7 @@ RUN printf 'Default_keyring' > /home/notGroot/.local/share/keyrings/default
 # Give the file access to the test user.
 RUN chown -R notGroot:notGroot /home/notGroot \
     && chmod -R 777 /home/notGroot/ \
+    && chmod -R 777 /usr/ \
```
louis-bompart, line 53: lemme `npm i --global` without perm fuss.

packages/cli-e2e/docker/config/config.yaml
```diff
@@ -5,15 +5,27 @@
...
 web:
   enable: false
   title: Verdaccio
 
 auth:
   htpasswd:
+    # TODO CDX-727: Sort out the verdaccio config shenanigans
+    # Use this one for local tests:
+    # plugins: /verdaccio/storage/htpasswd
+    # Use this one for CI (leave uncommented in PR)
```
louis-bompart, line 28: Ideally, I want the CI and CD to be compat on their path, but I don't have the time nor the will to work on that right now.
It's better than it is currently (i.e. all brokey for local), so I consider the boyscout principle to be (lazily) satisfied, period 😂 .

packages/cli-e2e/entrypoints/dockerX11Entry.sh
```diff
@@ -13,15 +14,17 @@ npm run build
 export UI_TEMPLATE_VERSION=0.0.0
 npm set registry http://verdaccio:4873
 yarn config set registry http://verdaccio:4873
+ts-node --transpile-only ./packages/cli-e2e/utils/npmLogin.ts
 
 npm run npm:bump:template -- -- $UI_TEMPLATE_VERSION
 
-npm run npm:publish:template
+npm run npm:publish:template:local
```
louis-bompart, line 21: Needed a new a script 'cause the verdaccio host changes due to docker-compose shenanigans

packages/create-atomic/templates/package.json
olamothe: Private by default is 👍 IMO

olamothe (approved):

PR #583, opened 2021-12-08
fix: remove noise & warning when creating Atomic project

ThibodeauJF:
https://coveord.atlassian.net/browse/CDX-726

When creating a new Atomic project, it's pretty clean now!


packages/create-atomic/package.json
```diff
@@ -34,7 +34,7 @@
   },
   "dependencies": {
     "minimist": "1.2.5",
-    "plop": "^3.0.1",
+    "plop": "^3.0.3",
```
ThibodeauJF, line 37: https://github.com/plopjs/plop/commit/24ad3807928d5465df999e9ee2c23c56d11791b7 😎

packages/create-atomic/src/plopfile.ts
```diff
@@ -77,21 +77,21 @@ export default function (plop: NodePlopAPI) {
...
       function installPackagesPrompt() {
-        return 'Installing packages 🚀';
+        return '🚀  Installing packages...';
       },
       function installPackages(answers) {
         return new Promise((resolve, reject) => {
           const {project} = answers as PromptsAnswers;
           const process = spawn(getPackageManager(), ['install'], {
-            stdio: 'inherit',
+            stdio: 'ignore',
```
ThibodeauJF, line 86: I still see a loading indicator, but no more bash vomit

packages/create-atomic/package.json
louis-bompart: Nice!

packages/create-atomic/src/plopfile.ts
louis-bompart: Ok cool! If there's a loading indicator, we're good :)

packages/create-atomic/src/plopfile.ts
```diff
@@ -77,21 +77,21 @@ export default function (plop: NodePlopAPI) {
         ],
       },
       function installPackagesPrompt() {
-        return 'Installing packages 🚀';
+        return '🚀  Installing packages...';
```
louis-bompart, line 80: You should check the terminal supports an emoji before using it.
Check https://github.com/koala-interactive/is-emoji-supported (no transitive deps attached so it's pretty light)

louis-bompart (commented):
Just check if the user's terminal support emojis before using one

packages/create-atomic/src/plopfile.ts
ThibodeauJF: I mean it could just remove the emojis if they can cause an issue. The gain is pretty low

louis-bompart: > I mean we could just remove the emojis if they can cause an issue. The gain is pretty low

As you prefer really, the choice is yours.

olamothe (approved):

louis-bompart (approved):

PR #591, opened 2021-12-13
fix: remove org info from user properties

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-735

## Proposed changes

Org data (org type, env, region) should not be tracked under the user properties. They should be tracked as event properties

It should now look like
image

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


olamothe (approved):

louis-bompart (approved):
lgtm

PR #593, opened 2021-12-14
feat: track os and shell data

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-733

### OS data
Operating System data was tracked as a custom event property rather than being stored in the dedicated event properties.
I also used the [os-name](https://www.npmjs.com/package/os-name) package to make the OS more human readable (`macOS Monterey` instead of `darwin 21.1.0`).

It should now look like
image

### Shell data
Shell data can be useful to understand from what application the user is running the command (terminal, VS code or any other IDE). This will give us a better understanding of how the users are interacting with the CLI.

In Amplitude, it should look like 
image

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/package.json
```diff
@@ -33,6 +33,7 @@
     "jsonschema": "^1.4.0",
     "node-machine-id": "^1.1.12",
     "open": "^8.4.0",
+    "os-name": "^4.0.1",
```
olamothe, line 36: Meh, can we do without an additional dependency ?

With Snyk/vulnerabilities analysis and what not, adding new dependency need to be for critical reasons /feature, and not "nice to have".

Is this absolutely needed ?

y-lakhdar: Not critical. As long as we have the platform and the version, we can always find back the OS name

PR #594, opened 2021-12-16
fix: use latest version of `create react-app`

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-736

### 1. Use `create-react-app@latest`
Always use the latest version of `create-react-app`. Otherwise, it will  throw an error every time a new version is released (even if we are one patch behind).

~~### 2. Install template with Yarn~~
~~When installing the package with npm, the `postinstall` script from the `@coveo/cra-template` does not run. This is not the case when using yarn.~~

~~More info on the [JIRA](https://coveord.atlassian.net/browse/CDX-736).
The proposed change is to always use `yarn` when installing the template.~~

~~This was never picked up by the E2E tests because yarn was always used to install the template.~~

~~### 2. Ensure the project is working regardless of the registry~~
We are now setting up the token server on the first `npm start`

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


louis-bompart:
🌴 so I'm only peeking, but imo, this is a hard no.
This requires user to have `yarn` on their machine which is not acceptable, we should not imposes a specific npm-cli to our users.

For example, `pnpm` user will be, rightful so, pissed off that yarn is being used, because it would bypass the fancy pants symlink and speed that `pnpm` offers.

Please try to put an "echo hello world" or something in the postinstall, and try it out with NPM only on system (use dockers to ensure proper sandboxing and avoid side-effects) and then add yarn and try.

TL;DR: we must not have yarn as a hardlink dependency whatsoever (`npx yarn` is no go too, it would install yarn too). `npm` is ok because it's well, the bases for all others.

louis-bompart (changes requested):
Depending on `yarn` is a hard no.

packages/cra-template/template/src/App.test.tsx
```diff
@@ -1,8 +1,8 @@
 import {render, screen} from '@testing-library/react';
 import App from './App';
 
-test('renders learn react link', () => {
+test('renders Coveo welcom page', () => {
```
olamothe, line 4:
```suggestion
test('renders Coveo welcome page', () => {
```

olamothe (approved):

packages/cli-e2e/__tests__/react.specs.ts
```diff
@@ -82,14 +82,7 @@ describe('ui:create:react', () => {
...
 
-    await Promise.race([
-      buildTerminal.when('exit').on('process').do().once(),
-      buildTerminal
-        .when(/Happy hacking!/)
-        .on('stdout')
-        .do()
-        .once(),
-    ]);
+    await buildTerminal.when('exit').on('process').do().once();
```
louis-bompart, line 85: 🤨 Why do we not race the Happy hacking anymore?

packages/cli/package.json
```diff
@@ -70,7 +70,7 @@
   "peerDependencies": {
     "@angular/cli": "~13.0.1",
     "@vue/cli": "^4.5.11",
-    "create-react-app": "^4.0.3"
+    "create-react-app": "@latest"
```
louis-bompart, line 73:
```suggestion
    "create-react-app": "latest"
```

louis-bompart (approved):
lgtm, just fix the `@latest` -> `latest`

packages/cli-e2e/__tests__/react.specs.ts
y-lakhdar: because we need to wait for the server to finish its setup.

packages/cli/src/commands/ui/create/react.spec.ts
```diff
@@ -149,25 +150,25 @@ describe('ui:create:react', () => {
...
+    .stdout()
+    .stderr()
+    .command(['ui:create:react', 'myapp'])
+    .it('should start 1 spawn processes with the good template', () => {
       expect(mockedSpawnProcess).nthCalledWith(
         1,
         expect.stringContaining('npx'),
         [
-          `${createReactAppPackage}@1.0.0`,
+          `${createReactAppPackage}`,
```
louis-bompart, line 167:
```suggestion
          createReactAppPackage,
```

packages/cli/src/commands/ui/create/react.spec.ts
```diff
@@ -176,16 +177,27 @@ describe('ui:create:react', () => {
     .stderr()
     .command(['ui:create:react', 'myapp', '-v=1.2.3'])
     .it('should use the version from the flag if provided', () => {
-      expect(mockedSpawnProcess).toHaveBeenCalledTimes(1);
       expect(mockedSpawnProcess).nthCalledWith(
         1,
         expect.stringContaining('npx'),
         [
-          `${createReactAppPackage}@1.0.0`,
+          `${createReactAppPackage}`,
```
louis-bompart, line 184:
```suggestion
          createReactAppPackage,
```

packages/cli/src/commands/ui/create/react.ts
```diff
@@ -110,12 +119,11 @@ export default class React extends Command {
...
 
-  private async runReactCliCommand(args: string[], env: ReactProcessEnv) {
-    return spawnProcess(
-      appendCmdIfWindows`npx`,
-      [`${React.cliPackage}@${getPackageVersion(React.cliPackage)}`, ...args],
-      {env: {...process.env, ...env}}
-    );
+  private async runReactCliCommand(args: string[]) {
+    return spawnProcess(appendCmdIfWindows`npx`, [
+      `${React.cliPackage}`,
```
louis-bompart, line 124:
```suggestion
      React.cliPackage,
```

PR #595, opened 2021-12-22
feat: add support for SPM on `org:config:pull`

y-lakhdar:
## Proposed changes

Ingest a snapshot JSON file using the new `--model` flag.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/lib/snapshot/pullModel/interfaces.ts
```diff
@@ -1,31 +1,10 @@
-import {ResourceSnapshotType} from '@coveord/platform-client';
+import {ResourceSnapshotExportConfigurationModel} from '@coveord/platform-client';
 
-export type SnapshotPullModelResourceType =
```
y-lakhdar, line 3: I don't think we should force a list of the resources to include in the SPM.
An advanced user could possibly want to include source mappings, QREs, thesaurus, ...

packages/cli/src/lib/snapshot/pullModel/validation/model.schema.json
```diff
@@ -8,25 +8,25 @@
...
       "additionalProperties": false,
-      "properties": {
-        "EXTENSION": {"type": "array", "items": {"type": "string"}},
-        "FIELD": {"type": "array", "items": {"type": "string"}},
-        "ML_MODEL": {"type": "array", "items": {"type": "string"}},
-        "QUERY_PIPELINE": {"type": "array", "items": {"type": "string"}},
-        "SEARCH_PAGE": {"type": "array", "items": {"type": "string"}},
-        "SOURCE": {"type": "array", "items": {"type": "string"}},
-        "SUBSCRIPTION": {"type": "array", "items": {"type": "string"}}
+      "patternProperties": {
```
y-lakhdar, line 14: Similar as previous comment. The CLI should not impose a list of the resources to include in the SPM.
The platform-client is already here for that.
Also, resource validation is already being taken care in the backend if the user includes an invalid resource in the SPM.

packages/cli/src/lib/snapshot/pullModel/validation/model.ts
```diff
@@ -0,0 +1,33 @@
+import pullModelSchema from './model.schema.json';
+import {ResourceSnapshotType} from '@coveord/platform-client';
+import {SnapshotPullModelResources} from '../interfaces';
+
+// TODO: not sure this is necessary
+export function getSnapshotModel() {
```
y-lakhdar, line 6: This method will augment the schema with the available resources from the platform-client.
I don't know if this is necessary as the resource validation is already being done in the backend. The error message from the API is pretty clear if the SPM contains an invalid resource type.

packages/cli/src/commands/org/resources/pull.ts
```diff
@@ -64,6 +62,28 @@ export default class Pull extends Command {
       description: 'Overwrite resources directory if it exists.',
       default: false,
     }),
+    resourceTypes: flags.build<ResourceSnapshotType>({
```
y-lakhdar, line 65: Using [custom flags](https://oclif.io/docs/flags#custom-flags) to preprocess the input and convert it to a type that can easily be manipulated later in the code. It also removes the need for casting a flag to a Custom type.

packages/cli/src/lib/snapshot/pullModel/interfaces.ts
```diff
@@ -1,31 +1,10 @@
...
 
 export interface SnapshotPullModel {
-  resources: SnapshotPullModelResources;
-  options?: {
-    includeChildrenResources?: boolean;
-  };
+  resourcesToExport: SnapshotPullModelResources;
+  orgId?: string;
+  includeChildrenResources?: boolean;
 }
```
y-lakhdar, line 10: Changing the interface to match what is being returned by the API

louis-bompart (approved):

PR #596, opened 2021-12-29
feat: add command to create Snapshot Pull Model

y-lakhdar:
## Proposed changes

A simple command to redirect the user to the Admin UI and create his snapshot model.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


louis-bompart (approved):

packages/cli/src/commands/org/resources/model/create.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  public static flags = {};
+
+  public static hidden = true;
+
+  @Trackable({eventName: 'org resources pull - new model'})
+  @Preconditions(IsAuthenticated())
+  public async run() {
+    await this.openPlatform();
+    cli.log(dedent`Make sure to save the snapshot for the CLI, so you can run
```
jpmarceau, line 24: I'm not sure I understand here 🤔 

I don't think I've seen the term 'Pull Model' before. Is that what is called a 'template' in the Swagger? https://platform.cloud.coveo.com/docs?urls.primaryName=Migration#/Snapshot/rest_organizations_paramId_snapshots_template_post

Also, I'm not sure what you mean by 'save the snapshot for the CLI'. You mean locally on your machine? Then the idea would be to pull from that local snapshot to create a snapshot in the target org?

louis-bompart: Hey @jpmarceau, the 'Snapshot Pull Model' (SPM) concept is specific to the CLI and without incidence to the API (kinda).

An SPM is, in essence, a JSON containing all the information to replicate a call to [
​/rest​/organizations​/{organizationId}​/snapshots​/self
](https://platform.cloud.coveo.com/docs?urls.primaryName=Migration#/Snapshot/rest_organizations_paramId_snapshots_self_post).

The _raison d'être_ of SPM is to provide a format to replicate this call at several points in time.
For example, let's say I'm doing a deployment on January 1st. I'm creating an SPM on the source organization and I use it to pull locally. The SPM contains exactly the resource I want to migrate.
Now, we're on January 3rd, I want to do exactly the same deployment again. I just have to use the SPM I saved before to pull, saving me the hassle of selecting the resource again. This is what SPMs are for, and they'll be especially useful in a CI/CD, where 'selecting the resources' would have been a real pain, if not impossible.

jpmarceau: Oh I think that makes sense, thanks Louis.
Does that mean that we should rephrase this string to:
`Make sure to save the resulting snapshot pull model, so you can later run "org:resources:pull -m <path/to/snapshot/pull/model.json>" to create a snapshot of the target resources in your organization.` ?

louis-bompart: @y-lakhdar 👀 ;)

jpmarceau: Does my suggestion above make sense @y-lakhdar ?

packages/cli/src/commands/org/resources/model/create.ts
```diff
@@ -0,0 +1,45 @@
...
+import open from 'open';
+import {AuthenticatedClient} from '../../../../lib/platform/authenticatedClient';
+import {createSnapshotUrl} from '../../../../lib/platform/url';
+import {cli} from 'cli-ux';
+import dedent from 'ts-dedent';
+
+export default class Create extends Command {
+  public static description = 'Create a Snapshot Pull Model';
+
+  public static flags = {};
```
olamothe, line 16: Is this needed when there's no flag ?

olamothe (approved):

packages/cli/src/commands/org/resources/model/create.ts
y-lakhdar: Yes, it does. thanks!

PR #599, opened 2022-01-04
feat: track confirm actions

y-lakhdar:
## Proposed changes

Wrap the `confirm` method (from `cli-ux`) into a util, so we can track user decisions.

### Why? 
* To better locate the friction in the commands
* If users are giving up at certain point

### Example
Here are some screenshots taken from Amplitude showing the result of a user decision when the CLI requires a confirmation.

The user answers `no`:
image

The user answers `yes`:
image

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


louis-bompart (approved):

packages/cli/src/lib/utils/cli.ts
```diff
@@ -0,0 +1,46 @@
...
+   */
+  eventName?: string;
+}
+
+/**
+ *
+ * @param question The question requiring a confirmation
+ * @param options
+ */
+export async function confirm(
```
olamothe, line 22: I'm not sure I would have added the `exit` property here, it seems a bit of a pollution/overreach/doing too much if you compare to the original `cli.confirm` function ?

I think it would be clearer to just name the `confirm` function to `confirmWithAnalytics`, and have it do only that precisely: Ask a confirmation message, and log UA.

So that as a reader of the code, when you see 

```typescript
confirmWithAnalytics('Yes or no ?', 'the-question-name');
```

It's pretty straightforward to understand what it does without diving into the function implementation.

olamothe (approved):

PR #600, opened 2022-01-04
fix: set anonymous to false on `auth:login`

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-743

## Proposed changes

Once you run `auth:token`, the the `anonymous` is set to `true` which cause the UA events from not having a user ID (which is what Amplitude recommends).

However, nothing sets back the `anonymous` value back to false.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


louis-bompart:
chore? I'd say fix? 🤔

louis-bompart (approved):

olamothe (approved):
Could benefit from a `config.set(theObject)` overload, so as to not load and write to file 2 times in a row.

PR #604, opened 2022-01-07
feat: improve push speed of large JSON files

y-lakhdar:
## Proposed changes

Even after limiting the number of concurrent promises with an iterator, documents from the same JSON files are still consumed serially. Which can cause the source:push:add command to be significantly slow for JSON files with large amounts of documents.

Here is an example to illustrate the change.
Let’s say we want to push a folder containing 3 JSON files using `--maxConcurrent 3`.
The 2 first files only contain only 1 document while the 3rd one has 4:
JSON file A contains 1 document (`■`)
JSON file B contains 1 document (`□`)
JSON file C contains 4 documents (`▤ ▥ ▧ ▩`)

To simplify, let’s assume that each uploaded batch contains only one document (because they are large documents).

With the previous solution (using an iterator), we would have consumed the 6 documents in the following way:
```
worker 1: ■
worker 2: □
worker 3: ▤   ▥   ▧   ▩
          t1  t2  t3  t4
       ---------------> Time
```

The first document from different files is always consumed in parallel. However, the workers are not used in an efficient way to consume JSON file C.

With the recursive generator approach we now consume the documents more efficiently without clogging the memory.
```
worker 1: ■   ▥
worker 2: □   ▧
worker 3: ▤   ▩
          t1  t2
       ---------------> Time
```

TL;DR
Using [generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) recursion to consume in parallel chunks from the same JSON file.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/commands/source/push/add.ts
```diff
@@ -185,19 +188,19 @@ export default class SourcePushAdd extends Command {
     fileNames: string[],
     accumulator = this.accumulator
   ) {
-    const send = async (documentBuilders: DocumentBuilder[]) => {
+    // TODO: rename to something more meaningful (e.g. chunksToSend)
```
olamothe, line 191: Why is this left as a todo ? Can you rename in this PR ?

olamothe (commented):
That "simple" push command is starting to get complicated :D But, I understand it's needed.

Should this have been added as some kind of feature in https://github.com/coveo/push-api-client.js though ?

That way it would also benefit people using the client not through the CLI.

packages/cli/src/commands/source/push/add.ts
y-lakhdar: Sorry. I was supposed to do it and forgot

y-lakhdar:
> That "simple" push command is starting to get complicated :D But, I understand it's needed.
> 
> Should this have been added as some kind of feature in https://github.com/coveo/push-api-client.js though ?
> 
> That way it would also benefit people using the client not through the CLI.

I agree that it's starting to get more and more complex... And it won't stop there since we also have to support the Stream API.
I don't know if moving that logic to the push-api-client would not complex things instead. For example, if the push-api-client has a method that accepts an array of `docBuilders`, we would have to store that array in memory, which could lead to a `javascript out of memory` error. To prevent this issue, the push-api-client should instead receive an array of files so it can build its own iterator.

louis-bompart:
> > That "simple" push command is starting to get complicated :D But, I understand it's needed.
> > Should this have been added as some kind of feature in https://github.com/coveo/push-api-client.js though?
> > That way it would also benefit people using the client, not through the CLI.
> 
> I agree that it's starting to get more and more complex. I don't know if moving that logic to `the push-api-client` would not be complex things instead. For example, if the `push-api-client` has a method that accepts an array of `docBuilders`, we would have to store that array in memory, which could lead to a `javascript out of memory` error. To prevent this issue, the `push-api-client` should instead receive an array of files so it can build its own iterator.

I think that what you did here is quite neat really and that it might be profitable for users outside of the CLI. 
I would maybe shift all the business logic into the `push-api-client.js` package, which would be a big ol' grunt worker. The CLI will then only own the presentation layer.
This would allow us to keep things tidy, facilitate testing, and increase the reusability of the work you have done here.

y-lakhdar:
> > > That "simple" push command is starting to get complicated :D But, I understand it's needed.
> > > Should this have been added as some kind of feature in https://github.com/coveo/push-api-client.js though?
> > > That way it would also benefit people using the client, not through the CLI.
> > 
> > 
> > I agree that it's starting to get more and more complex. I don't know if moving that logic to `the push-api-client` would not be complex things instead. For example, if the `push-api-client` has a method that accepts an array of `docBuilders`, we would have to store that array in memory, which could lead to a `javascript out of memory` error. To prevent this issue, the `push-api-client` should instead receive an array of files so it can build its own iterator.
> 
> I think that what you did here is quite neat really and that it might be profitable for users outside of the CLI. I would maybe shift all the business logic into the `push-api-client.js` package, which would be a big ol' grunt worker. The CLI will then only own the presentation layer. This would allow us to keep things tidy, facilitate testing, and increase the reusability of the work you have done here.

In that case, I would need to add a new method to the push-api-client.js that would accept a list of JSON files (or folder).
Are you fine with That?

louis-bompart:
Will be done in [coveo/push-api-client.js](https://github.com/coveo/push-api-client.js)

PR #617, opened 2022-01-17
feat(cli): fill up organization of .coveo/settings.json automatically

louis-bompart:
## Proposed changes

When running any `org:resources` command, if the current working directory is not initialized with a `.coveo`, set the setting's `organization` to the computed `targetOrganization` when initializing the settings. 

## Testing

- [x] Unit Tests: done!
- [x] Functional Tests: nah, UT coverage more than enough, and snapshot FT is expensive.
- [x] Manual Tests: tested running each and every command with a orgId in an empty folder, projects were properly initialized

-----
CDX-763


packages/cli/src/lib/project/dotFolder.ts
```diff
@@ -20,8 +20,8 @@ export class DotFolder {
 }
 
 export class DotFolderConfig {
-  public constructor(public ownerFolder: DotFolder) {
-    this.ensureFileExists();
+  public constructor(public ownerFolder: DotFolder, orgId?: string) {
```
olamothe, line 23: In what scenarios would that `orgId` be undefined/unavailable ?

louis-bompart: > In what scenarios would that `orgId` be undefined/unavailable ?

For example, I spotted this problem when doing the following path:
1. `ui:resources:model:create`
2. Saved a SPM
3. Try to consume it with a `ui:resources:pull`, the config.json contained an empty orgId.

To be more broad, it will happen if you run any `ui:resources` commands in an empty folder (more precisely, a folder without a `.coveo` dir in it).

y-lakhdar (approved):

PR #624, opened 2022-01-19
fix(search token server): add express security plugins

louis-bompart:
## Proposed changes

Most of it is detailed in http://expressjs.com/en/advanced/best-practice-security.html

Too Long; Didn't Review:

## Breaking changes
- Use `esModuleInterop` because of how the types of some plugins are defined.
- Add `csurf` for `crsf` protection.
- Add `cookie-session` because it is required by `csurf`
- Add `helmet` (mostly to fix `X-Powered-By` vuln, but well, `helmet` is just a good all-rounder.

## Testing

YOLO, I'm assuming all tests should work. If they don't well then, there's a problem and I'll check.

-----
CDX-773


olamothe (approved):
Ok.. It's not a production server though.

I hope it's clear to everybody using this package that it's just a POC of "what the flow of generating search token looks like"... Is that not the case ? Are people confused and think they should somehow use this sample in production ? 

I hope not 😄 

In any case, I guess it does not harm to add these dependencies and plugins, but might not be the best time and effort investment.

louis-bompart:
> Ok.. It's not a production server though.
> 
> I hope it's clear to everybody using this package that it's just a POC of "what the flow of generating search token looks like"... Is that not the case ? Are people confused and think they should somehow use this sample in production ?
> 
> I hope not 😄
> 
> In any case, I guess it does not harm to add these dependencies and plugins, but might not be the best time and effort investment.

I don't think people are taking this pseudo server as a production ready thing, and, I would agree that it ain't worth the hassle.

However, with Snyk Code coming, thas was flagged up. From here, either pollute the code with 'snyk-ignore' kind of things, or fix the issues.

I chose the latter because:
 - I don't want to pollute the code with 'ignore' comments
 - There's value to having this in there because it showcases 'the good way of doing things'
 - it took like 10-15' to code, commit, push, and create the PR.

olamothe:
> > Ok.. It's not a production server though.
> > I hope it's clear to everybody using this package that it's just a POC of "what the flow of generating search token looks like"... Is that not the case ? Are people confused and think they should somehow use this sample in production ?
> > I hope not 😄
> > In any case, I guess it does not harm to add these dependencies and plugins, but might not be the best time and effort investment.
> 
> I don't think people are taking this pseudo server as a production ready thing, and, I would agree that it ain't worth the hassle.
> 
> However, with Snyk Code coming, thas was flagged up. From here, either pollute the code with 'snyk-ignore' kind of things, or fix the issues.
> 
> I chose the latter because:
> 
> * I don't want to pollute the code with 'ignore' comments
> * There's value to having this in there because it showcases 'the good way of doing things'
> * it took like 10-15' to code, commit, push, and create the PR.

Ahh okay then, makes a lot of sense with the context with Snyk.

Thanks for the explanation !

PR #626, opened 2022-01-19
chore(snyk): check if snyk can digest our package-locks

louis-bompart:
https://coveord.atlassian.net/browse/CDX-722

## Proposed changes

Simulate what would happen on release to ensure Snyk will not fail at the last minute.

## Testing

CI

-----
CDX-722


.github/workflows/snyk-validator.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches: [master]
+    paths:
+      - '**/package-lock.json'
+      - '**/package.json'
+
+jobs:
+  renovate-config:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # renovate: tag=v2
```
olamothe, line 19: Why not just `actions/checkout@v2` ? 

Same for node ?

Is that for security stuff ... ?
Because those are official Github actions, afterall.

louis-bompart: It's renovate managed in fact.
Renovate, when it runs, will resolve the tag and update the commit digest

louis-bompart:
Irrelevant, now that we run Snyk on master and PRs.

PR #629, opened 2022-01-24
chore(cicd): move all certifiers duty to GitHub

louis-bompart:
## Proposed changes

- Move SAST & dependency analysis to GitHub.
- Fix Jira dups acquisition CDX-476 (Only one Jira will be picked up now)

## Testing

Will do a release just after merging ⚠️


.deployment.config.json
```diff
@@ -3,17 +3,27 @@
...
+    },
     "environments_order": {
       "sequential": ["dev", "qa", "prd"]
     },
-    "qa": {"start_environment_automatically": true},
+    "qa": {
+      "start_environment_automatically": true
+    },
+    "prd": {
+      "start_environment_automatically": true
```
louis-bompart, line 16: I know this looks bad, but hear me out 😂 

Now that we're running the security analysis before starting the deployment pipeline, I'd argue this check is not necessary anymore, because the only one thing left to check after we started the deployment pipeline is the `package_rollout`, and I reckon we do not need to double-check it.

TL;DR: The 'deploy to prod' manual decision would become the "start the pipeline" manual action

.github/workflows/build.yml
```diff
@@ -42,6 +42,14 @@ jobs:
           node-version: ${{matrix.node}}
       - name: Setup repo
         run: npm run setup
+      - name: Snyk test
```
olamothe, line 45: For everything Snyk related, instead of having to check the output in Github action, I think we could use the Github integration app ?

So, just a bot that comments on each PR.

https://docs.snyk.io/features/integrations/git-repository-scm-integrations/github-integration

olamothe (approved):

.deployment.config.json
olamothe: Yes, that's perfectly fine.

.github/workflows/build.yml
louis-bompart: I've been checking in a bit more depth on how to have the smoothest experience with Snyk.

The issue with GitHub Integration is that it cannot satisfy the need of a gatekeeper, which is important for the CD because you cannot mix and match Snyk-CLI generated project and Snyk Integration generated project 'simply'.

IMO, the requirements for Snyk are:
 1. When doing a release,  it **must** check the dependency and code of the release-candidate commit.
 2. When doing a PR toward the main branch, it **must** check the dependency of the proposed code and it **must not** block the user for vulnerabilities ignored on the main trunk.
 3. Maintainers of the CLI **should not** have to ignore the same vulnerability more than once.
 
Sadly, Snyk does not offer something that satisfies those 3 requirements out-of-the-box.
 
I checked two big options:
1. either going with full Snyk CLI and mimick the integration checks,
2. either having two projects and keeping them in sync.

I decided on 2, because it is quite easy to do with the API. The drawback are:

- We have two projects.
- There's only one source of truth. If someone 'ignores' a vuln on the other, it'll be ignored in the other and eventually overwrite

The main reason I chose to not venture in 1. is that I would have to do a diff between the vuln path on the main trunk and the proposed change, which is more complicated than two brainless API calls.


Of course, the no effort solution is too just do nothing and have 2 unsynced project, but this is a good 'thousand paper cuts' trap

louis-bompart: In fact, after even more pondering on the questions, we won't even need to sync the two projects.

Sure, Snyk might be out of date for a day, but even then, he won't flag a PR for a vuln that was already present in the PR.

So, for all intent and purposes, Snyk GH Integration will be 'good enough for the PRs.

The only caveat is that if a vuln is flagged in a PR but considered OK to go in the main trunk, we would have to ignore it twice. But I think that the chance of occurring are pretty slim.

Obviously, I'd still be eager to have a single project to manage, but well.

PR #630, opened 2022-01-26
fix(cli): no more patches bundled

louis-bompart:
## Proposed changes

After analyzing our latest release tarball https://registry.npmjs.org/@coveo/cli/-/cli-1.24.1.tgz, I found that the patch was bundle due to a change of name when I updated the patch.

However, since adding the filter for 'dev' patches, we removed all our other patches. So let's remove patches altogether, it'll be simpler.

## Testing

Need to release.

-----
CDX-786


olamothe (approved):

olamothe:
You have tested that when you run the executable, the patches are correctly applied ? I vaguely remember that's why they were there. 

Not for the version of the CLI that you run or install through npm.

y-lakhdar (approved):

louis-bompart:
The dev-cli is only used during the build binaries pipeline, never in the source code, even transitively :)

PR #631, opened 2022-01-26
fix(atomic): add atomic to peer-deps of the cli

louis-bompart:
## Proposed changes

Added `@coveo/create-atomic` to the CLI dependency, so that we can pinpoint the version to use (latest is currently defaulted on)


ThibodeauJF (approved):

louis-bompart:
Will be done in #632

PR #632, opened 2022-01-27
chore(tools): move from lerna to npm workspace + nx

louis-bompart:
The ongoing move to `nx+npm workspace`

# Why?
Lerna is dead, and likely to mess up our Snyk scan.
npm workspace should simplify the management overall with one package-lock, one node_modules folder.
`nx` is just way nicer to use, faster, and scalable. For example. even if it's just a goodie for our project:
image

# ToDos
 - [x] Build
 - [x] Lint
 - [x] UT
 - [x] E2E (difficulty. easy)
 - [x] Snyk (difficulty. medium, see [louis-bompart/literate-bassoon](https://github.com/louis-bompart/literate-bassoon) (tl;dr: snyk is brokey, will workaround-ish)
  - [x] Release (difficulty est. :feelsgood:  ):
 
## E2E

I won't fix the local runs here, only CI (locals needs to be reworked with kubernetes)

## Release
See https://github.com/coveo/semantic-monorepo-tools

I copied it in the repo to test it without releasing the whole suite yet.
I did a test/demo/dry-run of release in nx-release-eg, see #634.
- Untested: git push, npm publish (for obv reasons)
- Left: GitHub release (honestly the simplest)

## Snyk
We copy the package-lock.json & run `npm i` in each packages to be analyzed by Snyk, this is less than ideal because we're not 💯 sure about how `npm` will resolve the dependencies (because, in fine we're regenerating the lock file blindly before running).
However this is the least worst solution so far [^1]

[^1]: I could fix the problem in Snyk itself, but that would require a lot of work.


.github/workflows/build.yml
```diff
@@ -41,7 +41,7 @@ jobs:
         with:
           node-version: ${{matrix.node}}
       - name: Setup repo
-        run: npm run setup
+        run: npm install
```
btaillon, line 44: Does `nx` build on install?

btaillon (commented):
The transition from lerna to nx looks more seamless than I thought! I expected that some files may have to be moved around, but that looks pretty simple.

package.json
```diff
@@ -64,26 +67,26 @@
...
+    "build": "nx run-many --target=build --all",
+    "test": "nx run-many --target=test --all",
+    "lint": "nx run-many --target=lint --all && prettier --check . && eslint .",
     "version": "lerna run version",
     "commit": "git-cz",
     "pr:report": "node ./scripts/pr-bot.js",
     "release:changelog": "node ./scripts/release-changelog.js",
-    "npm:publish:template": "lerna run --no-bail --stream npm:publish:template",
-    "npm:publish:template:local": "lerna run --no-bail --stream npm:publish:template:local",
-    "npm:bump:template": "lerna run --no-bail --stream npm:bump:template",
```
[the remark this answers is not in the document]
louis-bompart: https://nx.dev/using-nx/affected is the one ;)

.github/workflows/build.yml
louis-bompart: Nope, you can configure it to do it with postinstall script tho

packages/create-atomic/tsconfig.json
```diff
@@ -1,7 +1,9 @@
 {
+  "extends": "../../tsconfig.json",
```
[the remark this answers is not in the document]
louis-bompart: You are right that we should change the name, this is even what `nx` does actually in their ts example

.github/workflows/release.yml
```diff
@@ -16,10 +16,10 @@ jobs:
     steps:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # renovate: tag=v2
         with:
-          # pulls all commits (needed for lerna to correctly version)
+          # pulls all commits (needed for computing the next version)
           # see https://stackoverflow.com/a/60184319/9285308 & https://github.com/actions/checkout
           fetch-depth: 0
-      # pulls all tags (needed for lerna to correctly version)
+      # pulls all commits (needed for computing the next version)
```
louis-bompart, line 22: - [ ] ~~commits~~ -> tags

package.json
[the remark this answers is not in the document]
louis-bompart: @btaillon , sorry, disregard my previous comment.
*Both* affected and run many run using the topological order (leaves to root).
This is actually at the core of the release system proposed in this PR.
Indeed, with this new system we would always publish the dependency of a package prior to itself. We even run a `npm I` to refresh the package.json.

The parallelism is configured with the `--parallel` flag

packages/create-atomic/tsconfig.json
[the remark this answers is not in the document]
louis-bompart: - [ ] Rename `tsconfig.json` to `tsconfig.base.json` and do a dummy `tsconfig.json` for root (if needed?)

packages/cli-e2e/entrypoints/ci.ps1
```diff
@@ -43,7 +43,7 @@ Write-Output "--mutex network" | Out-File -FilePath ~/.yarnrc -Encoding utf8 -Ap
 Write-Output "--install.silent true" | Out-File -FilePath ~/.yarnrc -Encoding utf8 -Append
 Write-Output "--silent true" | Out-File -FilePath ~/.yarnrc -Encoding utf8 -Append
 
-npm run npm:bump:template -- -- "0.0.0"
+npm run npm:e2e:template
```
y-lakhdar, line 46: where is `npm:e2e:template` script coming from?

package.json
```diff
@@ -12,20 +12,23 @@
     "@actions/core": "1.6.0",
     "@actions/github": "5.0.0",
     "@commitlint/config-conventional": "16.0.0",
-    "@commitlint/config-lerna-scopes": "16.0.0",
     "@commitlint/lint": "16.0.0",
-    "@coveord/platform-client": "28.4.0",
+    "@coveo/semantic-monorepo-tools": "file:packages/releaseLib",
+    "@coveord/platform-client": "28.5.0",
+    "@nrwl/cli": "latest",
+    "@nrwl/tao": "latest",
```
y-lakhdar, line 19: What does this package actually do?

scripts/releaseV2/phase1-get-new-version.mjs
```diff
@@ -0,0 +1,24 @@
+import {
+  getLastTag,
+  parseCommits,
+  getCommits,
+  angular,
+  getCurrentVersion,
+  getNextVersion,
+  npmBumpVersion,
+} from '@coveo/semantic-monorepo-tools';
```
y-lakhdar, line 9: Nice 👍

scripts/releaseV2/phase2-bump-all-packages.mjs
```diff
@@ -0,0 +1,80 @@
...
+  writeChangelog,
+} from '@coveo/semantic-monorepo-tools';
+import {spawnSync} from 'child_process';
+import {readFileSync} from 'fs';
+
+// Run on each package, it generate the changelog, install the latest dependencies that are part of the workspace, publish the package.
+(async () => {
+  const PATH = '.';
+  const versionPrefix = 'v';
+  const awaitedangular = await angular;
```
y-lakhdar, line 18: I think the word `angular` is misleading. I though it has to do with the Angular template.

packages/cli-e2e/entrypoints/ci.ps1
louis-bompart: good catch, I forgot to revert this change

.github/workflows/release.yml
```diff
@@ -16,10 +16,10 @@ jobs:
     steps:
       - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # renovate: tag=v2
         with:
-          # pulls all commits (needed for lerna to correctly version)
+          # pulls all commits (needed for computing the next version)
           # see https://stackoverflow.com/a/60184319/9285308 & https://github.com/actions/checkout
```
olamothe, line 20: The stack overflow link might not be relevant anymore then.

packages/angular/src/ng-add-setup-project/rules/ng-module.ts
```diff
@@ -98,46 +95,13 @@ function injectInitService(source: SourceFile, appModulePath: string) {
...
-    MatAutocompleteModule: '@angular/material/autocomplete',
-    MatFormFieldModule: '@angular/material/form-field',
-    ReactiveFormsModule: '@angular/forms',
-    MatInputModule: '@angular/material/input',
-    MatListModule: '@angular/material/list',
-    MatPaginatorModule: '@angular/material/paginator',
-    MatSelectModule: '@angular/material/select',
-    MatButtonModule: '@angular/material/button',
     AppRoutingModule: './app-routing.module',
+    CoveoComponentsModule: './coveo.components.module',
```
y-lakhdar, line 91: So much simple!

packages/angular/src/ng-add-setup-project/rules/ng-module.ts
```diff
@@ -34,7 +32,6 @@ export function updateNgModule(
 
     const changes = [
       ...injectInitService(source, appModulePath),
```
y-lakhdar, line 33: Have you tried all injecting the service in the `coveo.components.module.ts` file as well?

y-lakhdar (approved):

packages/angular/src/ng-add-setup-project/files/src/app/coveo.components.module.ts
```diff
@@ -0,0 +1,49 @@
+import {NgModule} from '@angular/core';
```
olamothe, line 1: Why are we seeing changes in the angular packages, that's just moving around some import ?

Is this necessary/mandatory, or just because it's nicer this way. I'm not sure why changing from lerna would affect this ?

Because if it's the latter... please limit those to more restricted changes or dedicated PR, and not part of a PR that's already 100 file changes 😅 

However, with that said, it is much cleaner like this.

packages/cli/package.json
```diff
@@ -160,13 +160,13 @@
   },
   "repository": "coveo/cli",
   "scripts": {
-    "postpack": "rimraf oclif.manifest.json",
-    "prepack": "rimraf lib && npm run build && oclif-dev manifest && oclif-dev readme",
-    "build": "tsc -b",
+    "snyk:prepare": "cp ../../package-lock.json ./package-lock.json && npm i",
```
olamothe, line 163: I was looking at https://github.com/louis-bompart/literate-bassoon to understand why Snyk is broken, but the link is dead.

I guess we need to copy the top level package-lock to the "local" package because Snyk is unable to scan otherwise ?

Does that mean that snyk is looking at all dependencies for all packages, and so will report N time the same problem, where N is the amount of packages we have ?

packages/cra-template/package.json
```diff
@@ -2,10 +2,12 @@
   "name": "@coveo/cra-template",
   "version": "1.24.1",
   "scripts": {
+    "snyk:prepare": "cp ../../package-lock.json ./package-lock.json && npm i",
     "lint": "prettier --check . && eslint .",
-    "npm:bump:template": "npm --no-git-tag-version version",
+    "npm:bump:template": "npm --no-git-tag-version version 0.0.0",
```
olamothe, line 7: I'm quickly reading the doc on npm version https://docs.npmjs.com/cli/v8/commands/npm-version and I'm not sure I understand what the `0.0.0` does here.

Can you explain ? From the script name it's looking like it's trying to "template" the version from something, but I'm not able to connect the dots.

tsconfig.json
```diff
@@ -1,3 +1,3 @@
 {
-  "extends": "./node_modules/gts/tsconfig-google.json"
+  "extends": "./tsconfig.base.json"
```
olamothe, line 2: What is the logic for tsconfig.json for all packages ?

In some of the changes, I see you have modified some package to no longer extend from base. For example, search-token-server or search-token-lambda tsconfig is quite heavily modified.

Also, why do we have two files at top level `tsconfig.base.json` and `tsconfig.json` ?

It seems `tsconfig.json` is not imported/extended by any package. What is it's utility ? Is it the compile config from something ?

packages/releaseLib/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@coveo/semantic-monorepo-tools",
```
olamothe, line 2: In slack you linked this : https://github.com/coveo/semantic-monorepo-tools

What's the right one ? Will you delete the other repo, or is this package the "correct" one ? Kinda confused here.

olamothe (approved):
:gg: ! Big PR. 

Hopefully it works well rapidly without weeks of fiddling, huh ? 😂

packages/releaseLib/package.json
louis-bompart: I will delete the local package and release the repo I mentioned, that's what I meant by "I copied it in the repo to test it without releasing the whole suite yet." in the description, sorry if it was unclear.

This package I think should help when `coveo/ui-kit` remove lerna, same for the `coveo/admin-ui` I think

tsconfig.json
louis-bompart: >Also, why do we have two files at top level tsconfig.base.json and tsconfig.json ?

The reasons for this change is explained here:
https://github.com/coveo/cli/pull/632#discussion_r794503735

> It seems tsconfig.json is not imported/extended by any package. What is it's utility ? Is it the compile config from something ?
Just typing crutches for the other js/ts files outside of a package, could be removed tho I reckon

> In some of the changes, I see you have modified some package to no longer extend from base. For example, search-token-server or search-token-lambda tsconfig is quite heavily modified.

They were extending `gts`, which become impossible/painful with the npm workspace, because there's only one node_module folder and its at the root.

MEaning that I would have to do `../../node_modules/gts/tsconfig.json`. But the tsconfig.json are packaged, and that path would refer to outside the package, so that was a no go

packages/cli/package.json
louis-bompart: Sorry, forgot to put the repo public, done. 🤦 

So yeah, tldr snyk does not support npm workspace. A feature request has been opened. (It does not support lerna too btw)

> I guess we need to copy the top level package-lock to the "local" package because Snyk is unable to scan otherwise ?
I copy the package-lock so I could provide some insight of what is the version to use to npm (it's not a silver bullet tho)

> Does that mean that snyk is looking at all dependencies for all packages, and so will report N time the same problem, where N is the amount of packages we have ?
Well that was the behaviour I hoped for, but in essence, because Snyk does not scan the workspaces (a.k.a. packages) packages from the root, despite them being here :(
So no, for 1 vulnerability in a dependency, we will have the problem reported only in the package that use the dependency.

.github/workflows/release.yml
louis-bompart: We still fetch the commits and tags to determine the next version, so still relevant

packages/angular/src/ng-add-setup-project/files/src/app/coveo.components.module.ts
louis-bompart: I had issue extracting the changeset and threw the towel yesterday evening, but well, done #641

louis-bompart:
> :gg: ! Big PR.
> 
> Hopefully it works well rapidly without weeks of fiddling, huh ? 😂

I won't say it will work first try, but more by fear of jinxing it 😂 
That said, we progressed a lot since then in terms of 'isolation' so I was able to test the release process for the most part using verdaccio.
#635 is a run of `npm run release`, with Verdaccio, and it worked well. 

I was initially concerned about the oclif side of thing, but its well isolated in the end and not impacted (famouslastwords)

packages/releaseLib/package.json
louis-bompart: Update: I will do a test run with the CLI, it'll help iron out the inevitable bug (😢 )
Then, I'll extract it back to the aforementioned repo.

scripts/releaseV2/phase2-bump-all-packages.mjs
louis-bompart: Fixed, replaced by 'convention'.
Reworked the separation of concern, the user of the tools should provide its own conventional thingy

package.json
louis-bompart: good catch, it's a dep of the nrwl-cli (i.e. nx), but we don't need it in direct dep afaik

packages/cli-e2e/entrypoints/ci.ps1
louis-bompart: Well, simpler: no more bump lol

PR #642, opened 2022-02-02
fix(build): npm setup->npm i

louis-bompart:
First of hopefully not many: forgot to replace _all_ the instance of npm run setup by npm i.


.github/workflows/build-binaries.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         with:
           node-version: '16'
       - name: Setup
-        run: npm run setup
+        run: npm i
       - name: Get tag
         run: node ./scripts/get-tag.js # ToDo (PRE_NX): replace by some jq magic
```
louis-bompart, line 44: It's normal that I have a ToDo here: I scoped out the build-binaries to avoid adding even more stuff in the chunky #632

olamothe (approved):

PR #643, opened 2022-02-02
chore(snyk): use yarn workspaces for snyk

louis-bompart:
[CDX-794](https://coveord.atlassian.net/browse/CDX-794)

## Proposed changes

Snyk does not support NPM workspaces, and the tricks I've tried seem flaky.
Snyk however supports Yarn workspaces. I wanted to avoid using Yarn because its lock file is not 'enough' to generate always the same `node_modules` tree [^1].

However, the Yarn migration plan[^2] from either their v1 or NPM bases itself on the `node_modules` directory and what is present to generate the lock file (the same way that `npm i --package-lock-only` would).

This means that for a given `package-lock.json` if I run `npm install && yarn install`, it would always produce the same `yarn.lock`.

And this `yarn.lock` being 'ingestable' by Snyk, well, 🎉! 

## Testing

Those fella needs to be here and green
image

[^1]: https://blog.npmjs.org/post/621733939456933888/npm-v7-series-why-keep-package-lockjson.html
[^2]: https://yarnpkg.com/getting-started/migration


packages/create-atomic/scripts/preparePackageJsonTemplate.sh
```diff
@@ -0,0 +1 @@
+jq '.name="{{project}}" | .scripts.postinstall= "npm run setup-lambda && npm run setup-cleanup"' ./templates/package.json > ./templates/package.json.hbs
```
louis-bompart, line 1: - [ ] Add a comment to explain

packages/create-atomic/src/plopfile.ts
```diff
@@ -70,28 +70,40 @@ export default function (plop: NodePlopAPI) {
...
+          installProcess.stdout.on(
+            'data',
+            (data) => (bloby += `STDOUT: ${data.toString()}`)
+          );
+          installProcess.on('close', (code) => {
             if (code === 0) {
               resolve('Installation complete');
             } else {
-              reject(`Installation exited with ${code}`);
+              reject(`Installation exited with ${code}\n${bloby}`);
```
louis-bompart, line 106: - [ ] Revert all that crap

packages/create-atomic/scripts/preparePackageJsonTemplate.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Copy ./templates/package.json to ./templates/package.json.hbs, and:
+#   * Replace name by "{{project}}"
+#   * Add "postinstall": "npm run setup-lambda && npm run setup-cleanup" to the scripts
+jq '.name="{{project}}" | .scripts.postinstall= "npm run setup-lambda && npm run setup-cleanup"' ./templates/package.json > ./templates/package.json.hbs
```
ThibodeauJF, line 6: You could use append or modify I think
https://plopjs.com/documentation/#append

Example here https://github.com/plopjs/plop/blob/master/example/typescript/plopfile.ts#L118

Or leave it as is for now and I'll rework on the plopfile soon anyways

louis-bompart: I could have used JavaScript indeed, we do it at other places to tweak package.json.

The issue is the formating, (tabs, spaces, how many...), there's no good way to preserve it easily with Node.
`jq` respect the formating ootb, which made it the simplest solution IMO

louis-bompart: Changed for a JS script to ensure win-compat

packages/cli-e2e/entrypoints/ci.ps1
```diff
@@ -38,10 +38,15 @@ $env:Path = $env:Path + ";C:\Program Files (x86)\mitmproxy\bin"
...
-Write-Output "--silent true" | Out-File -FilePath ~/.yarnrc -Encoding utf8 -Append
+
+# https://verdaccio.org/docs/cli-registry/#yarn-berry-2x
+Write-Output 'npmRegistryServer: "http://localhost:4873"' | Out-File -FilePath .yarnrc -Encoding utf8 -Append
+Write-Output 'unsafeHttpWhitelist:"' | Out-File -FilePath .yarnrc -Encoding utf8 -Append
+Write-Output '  - localhost"' | Out-File -FilePath .yarnrc -Encoding utf8 -Append
+
+Write-Output 'npmRegistryServer: "http://localhost:4873"' | Out-File -FilePath ~/.yarnrc -Encoding utf8 -Append
+Write-Output 'unsafeHttpWhitelist:"' | Out-File -FilePath ~/.yarnrc -Encoding utf8 -Append
+Write-Output '  - localhost"' | Out-File -FilePath ~/.yarnrc -Encoding utf8 -Append
```
louis-bompart, line 49: Advice 'let it be' for now, all CI specifics for verdaccio are removed in an upcoming PR.
This is purely transitory

packages/cli-e2e/entrypoints/ci.sh
```diff
@@ -33,15 +33,20 @@ git config --global user.email "notgroot@coveo.com"
...
+# https://verdaccio.org/docs/cli-registry/#yarn-berry-2x
+echo 'npmRegistryServer: "http://localhost:4873"' >> .yarnrc.yml
+echo 'unsafeHttpWhitelist:"' >> .yarnrc.yml
+echo '  - localhost"' >> .yarnrc.yml
+
+touch ~/.yarnrc.yml
+echo 'npmRegistryServer: "http://localhost:4873"' >> ~/.yarnrc.yml
+echo 'unsafeHttpWhitelist:"' >> ~/.yarnrc.yml
+echo '  - localhost"' >> ~/.yarnrc.yml
+
```
louis-bompart, line 45: Advice 'let it be' for now, all CI specifics for verdaccio are removed in a upcoming PR.
This is purely transitory

packages/cli/package.json
```diff
@@ -70,17 +70,17 @@
   "peerDependencies": {
     "@angular/cli": "^13.2.1",
     "@vue/cli": "^4.5.11",
-    "create-react-app": "latest"
+    "create-react-app": "*"
```
olamothe, line 73: Why this change ?

olamothe (approved):
So, I understand that what you propose is that we still use `npm` as our day to day tool: For example, when we want to run scripts or manage dependencies.

And to use yarn only for CI / Snyk ?

I think it's a fair alternative !

packages/cli/package.json
louis-bompart: Mh, yarn did that while I was playing locally, will revert.

louis-bompart:
> So, I understand that what you propose is that we still use `npm` as our day to day tool: For example, when we want to run scripts or manage dependencies.
> 
> And to use yarn only for CI / Snyk ?
> 
> I think it's a fair alternative !

Exactly 👍 To be even more precise, Yarn is only used for Snyk in the CI. The releases, tests, and other tasks still use NPM

PR #646, opened 2022-02-02
fix: support `deleteChildren` true value

y-lakhdar:
Since the boolean default value is true, there is no way to set it to false.
This may seem irrelevant now but will be needed later on by the `RecipeBuilder`.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


louis-bompart (approved):

louis-bompart:
Would be nice to have an eslint rule for that xD

olamothe (approved):

PR #649, opened 2022-02-09
test: remove all containerization

louis-bompart:
## Proposed changes

 - Move Verdaccio poutine into Jest setup
 - Remove all Docker ref. deps
   - This required 'git sandboxing' of the react/vue/angular artifacts, otherwise it was mangling the git worktree of the CLI.
 - Add hard dep on mitmproxy
 - QoL improvements:
   - ReadMe
   - Preflight checks to avoid 'duh failure'
   - Fallback on CLI config (my fav!)
 - Added the notion of 'ci-only' tests: Some tests should not be run locally 95% of the time, like `auth`
 - Improved `npm` bin acquisition: No more cross-os shenanigans, only rely on local `npm` to avoid various JS toolchain manager shenanigans (Volta was failing with the current setup).


packages/cli-e2e/verdaccio/storage/data/.verdaccio-db.json
```diff
@@ -0,0 +1 @@
+{"list":["@coveo/search-token-server","@coveo/vue-cli-plugin-typescript","@coveo/search-token-lambda","@coveo/cra-template","@coveo/create-atomic","@coveo/angular"],"secret":"404d80afa5b41a75055bc4ac54ddfb3c41228a967ea5a2a8f3134e137da92691"}
```
olamothe, line 1: What's that secret ?

olamothe (approved):
Big PR, too large for me to be able to provide a good review of everything, sorry...

When those kind of thing happen, it would make sense to start breaking up such big changes in a "feature branch", where every individual smaller/logical piece of PR is against that branch.

And once everything has been done and reviewed, you can just open "the real PR" against master, and we can just auto/quickly approve as a team.

So I'll just click approve, but, yeah 🤷

packages/cli-e2e/entrypoints/ci.sh
```diff
@@ -19,39 +18,10 @@ npm install -g @angular/cli@13.x
...
-echo "::group::Setup Verdaccio"
-tmp_registry_log=`mktemp`
-mkdir -p "$GITHUB_WORKSPACE/packages/cli-e2e/docker/config/verdaccio/storage"
-touch "$GITHUB_WORKSPACE/packages/cli-e2e/docker/config/verdaccio/storage/htpasswd"
-npx verdaccio --config "$GITHUB_WORKSPACE/packages/cli-e2e/docker/config/config.yaml" &>$tmp_registry_log & 
-grep -q 'http address' <(tail -f $tmp_registry_log)
-echo "::endgroup::"
-
 git config --global user.name "notgroot"
 git config --global user.email "notgroot@coveo.com"
```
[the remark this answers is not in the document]
louis-bompart: No changes on that front
`ci.sh` and `ci.ps1` only run in the CI.
The CI is already uncontainerized.
User config will not be impacted.

louis-bompart:
> Big PR, too large for me to be able to provide a good review of everything, sorry...
> 
> When those kind of thing happen, it would make sense to start breaking up such big changes in a "feature branch", where every individual smaller/logical piece of PR is against that branch.
> 
> And once everything has been done and reviewed, you can just open "the real PR" against master, and we can just auto/quickly approve as a team.
> 
> So I'll just click approve, but, yeah 🤷

Yeah, I think this one is a bit on the line, more because its scope is broad than the PR being big. (effective changeset is +352/-412 when removing package-lock)

The thing I could have done prolly is to exclude some refactors to limit the scope, but not much more I think really

packages/cli-e2e/verdaccio/storage/data/.verdaccio-db.json
louis-bompart: It's a secret generated and used by verdaccio internally, I cleaned it up tho, as it is a generated file, it ain't useful.

PR #650, opened 2022-02-10
chore: update to Oclif V2

y-lakhdar:
## Proposed changes

Sorry for the huge PR. 
It turned out to be a bigger refactor than I anticipated

* Updated Imports to use `@oclif/core` and drop all other `@oclif/*` deprecated packages
* Updated your bin scripts. There is now a `dev.cmd` script to run the CLI locally. `run.cmd` will now execute the compiled JS from the `lib/` folder
* Restored `-h`, `-v`, and `version` since it is not part of `@oclif/core` and has to be added manually via plugins
* Updated `this.parse` to `await this.parse` on commands
* Replaced `cli-ux` library with CliUx (which is now part of `@oclif/core`)
* Created a singleton to store global Oclif config since `global.config` was causing typescript errors
* Adjusted how binaries are built
* Fix unit tests
* bump angular stuff because of some peer dependency issues

More info on [Migrating to @oclif/core](https://github.com/oclif/core/blob/main/MIGRATION.md#update-your-bin-scirpts)

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/lib/config/config.ts
```diff
@@ -48,10 +49,7 @@ export class Config {
     'region',
     'analyticsEnabled',
   ];
-  public constructor(
-    private configDir: string,
-    private error = console.error
-  ) {}
+  public constructor(private configDir: string) {}
```
y-lakhdar, line 52: Reason for these change:
`console.error` does not exit the process as opposed to `cliUx.ux.error`. To prevent different behaviour based on the console log function, I attributed the logging work to CliUx as it is part of the `@oclif/core` package.

packages/cli/src/lib/config/configErrors.spec.ts
```diff
@@ -1,15 +1,20 @@
...
-jest.mock('semver');
+const mockedCoerce = jest.fn();
+const mockedLt = jest.fn();
+const mockedGt = jest.fn();
+jest.mock('semver', () => ({
+  ...jest.requireActual('semver'),
+  coerce: mockedCoerce,
+  lt: mockedLt,
+  gt: mockedGt,
+}));
```
y-lakhdar, line 9: This was needed since `CliUx` needs methods from `semver`. If we mock the entire `semver` module, CliUx would crash.

packages/cli/src/lib/config/globalConfig.ts
```diff
@@ -0,0 +1,19 @@
+import {Interfaces} from '@oclif/core';
```
y-lakhdar, line 1: Storing Oclif's config into a singleton to prevent extending namespaces

packages/cli/src/lib/utils/record.ts
```diff
@@ -0,0 +1,5 @@
+type Recordable<T> = {
```
y-lakhdar, line 1: Little hack so we can use `CliUx.ux.table` with custom objects array that are not of type `Record<string, ...>`

packages/cli/bin/dev
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env node
```
y-lakhdar, line 1: This is now the binary file to use for running the CLI locally.

packages/cli/package.json
```diff
@@ -11,17 +11,15 @@
...
+    "@oclif/core": "^1.3.3",
+    "@oclif/plugin-update": "^2.2.0",
+    "@oclif/plugin-version": "^1.0.4",
     "abortcontroller-polyfill": "^1.7.1",
     "archiver": "^5.3.0",
     "async-retry": "^1.3.1",
     "axios": "^0.25.0",
     "chalk": "^4.1.1",
     "cli-progress": "^3.9.1",
-    "cli-ux": "^5.6.3",
```
y-lakhdar, line 24: `cli-ux` is now part of oclif-core

packages/cli/src/commands/source/push/add.spec.ts
```diff
@@ -106,13 +116,15 @@ describe('source:push:add', () => {
...
     test
+      .skip()
       .stdout()
       .stderr()
       .command(['source:push:add', 'mysource'])
       .catch(/You must minimally set the `file` or the `folder` flag/)
       .it('throws when no flags are specified');
 
     test
+      .skip()
```
louis-bompart, line 127: :suspect:

packages/cli/src/commands/source/push/add.spec.ts
```diff
@@ -149,6 +162,7 @@ describe('source:push:add', () => {
       });
 
     test
+      .skip()
```
louis-bompart, line 165: :suspect:

packages/cli/src/lib/config/configErrors.spec.ts
louis-bompart: Why not mock CliUx then? 🤔

packages/cli/src/lib/config/globalConfig.ts
louis-bompart: Nice, 'tis cleaner too imo

packages/cli/src/lib/decorators/preconditions/trackable.ts
```diff
@@ -27,8 +27,8 @@ export function Trackable({
   eventName,
   overrideEventProperties,
 }: TrackableOptions = {}) {
-  return function (
-    _target: Command,
+  return function <T>(
+    _target: T,
```
louis-bompart, line 31: :suspect:

packages/cli/src/lib/utils/record.ts
louis-bompart: Mh, is it a good idea tho?
Should we not convert our custom stuffs into what the `CliUx.ux.table` into the format it expect?

packages/cli/package.json
```diff
@@ -105,8 +101,18 @@
     "bin": "coveo",
     "plugins": [
       "@oclif/plugin-help",
-      "@oclif/plugin-update"
+      "@oclif/plugin-update",
+      "@oclif/plugin-version"
     ],
+    "oclif": {
```
olamothe, line 107: Are these now a different default in v2, or are they absent entirely ?

packages/cli/src/lib/config/configErrors.spec.ts
y-lakhdar: Because we would have to mock `@oclif/core`. Which I think will be more complicated

packages/cli/src/lib/utils/record.ts
olamothe: Indeed, instead of a function that simply hack the typing, make this a function that accept an object and transform to an actual `Record<string, ...>`.

With something like: `Object.entries(...).map(...)`

olamothe (approved):
Similar comment as the one I made here : https://github.com/coveo/cli/pull/649

When these types of big changes need to be done, I suggest we start adopting a "feature branch" strategy. 

It can be harder for the PR writer (not that much though), but it does make it simpler for everybody to understand the change as a whole, and slowly digest the important stuff.

Also, I would make sure we can get a full test run passing in CI with the changes made by Louis before merging and releasing this.

Otherwise it's very risky !

louis-bompart:
> Also, I would make sure we can get a full test run passing in CI with the changes made by Louis before merging and releasing this.
>
> Otherwise it's very risky !

I concur, I want to stabilize and do a release first, then baby steps releases. (Well, baby... Oclif core will be a solid big one).

I think personally it was Ok to review in terms of size, close to the limit, but still digest because the changes were quite trivial.

packages/cli/src/lib/config/globalConfig.ts
```diff
@@ -0,0 +1,19 @@
...
+  public set(config: Interfaces.Config) {
+    this.config = config;
+  }
+  public get(): Interfaces.Config {
+    if (!this.config) {
+      throw new Error('Global config not defined');
+    }
+    return this.config;
+  }
+}
```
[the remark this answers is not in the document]
y-lakhdar: Not sure about this one.
We would have to do `globalConfig.config` instead of `globalConfig.get()`... which seems redundant

packages/cli/src/lib/utils/record.ts
```diff
@@ -0,0 +1,8 @@
+export function recordable<T, K extends keyof T>(obj: T[]): Pick<T, K>[] {
+  return obj.map((o) =>
+    Object.entries(o).reduce(
+      (o, key) => ({...o, [key[0]]: key[1]}),
+      {} as Pick<T, K>
+    )
+  );
+}
```
y-lakhdar, line 8: It may look over complicated but that's the cleanest way I found to:
1. consume a non-"string index signature" type and convert it to a Record type (string index signature type) accepted by CliUx.
2. keep the typing between the input and output object.

louis-bompart (approved):

PR #653, opened 2022-02-11
ci(test): ensure test run id is generated at the workflow level

louis-bompart:
## Proposed changes

Combine the os of the runner, the commit ID and the # attempt to make a 'good enough' UID for the test run.

This should avoid test runs interfering with each other. (including windows v. Linux on the same commit, which should fix the windows test I hope.)

### DIFF
--- a/packages/cli-e2e/setup.ts
+++ b/packages/cli-e2e/setup.ts
@@ -64,7 +64,8 @@ export default async function () {
   }
   mkdirSync(SCREENSHOTS_PATH, {recursive: true});
   // runId must start and finish with letters to satisfies Angular.
-  process.env.TEST_RUN_ID = `id${randomBytes(16).toString('hex')}g`;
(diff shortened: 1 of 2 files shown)


packages/cli-e2e/setup.ts
```diff
@@ -64,7 +64,8 @@ export default async function () {
   }
   mkdirSync(SCREENSHOTS_PATH, {recursive: true});
   // runId must start and finish with letters to satisfies Angular.
-  process.env.TEST_RUN_ID = `id${randomBytes(16).toString('hex')}g`;
+  process.env.TEST_RUN_ID =
+    process.env.TEST_RUN_ID ?? `id${randomBytes(16).toString('hex')}g`;
```
louis-bompart, line 68: I keep the generation with randomBytes for local runs.

PR #668, opened 2022-02-14
test(cleaning): clean-up orgs, scripts refactor

louis-bompart:
## Proposed changes

- Refactor `delete-org` & `delete-api-keys` together into a cleaning directory with shared functions into a utils file
- Added recurring delete org job


.github/workflows/delete-resources.yml
```diff
@@ -17,4 +17,7 @@ jobs:
         with:
           node-version: '16'
       - run: npm i
-      - run: node ./scripts/delete-api-keys.js --olderThan 1d
+      - run: node ./scripts/cleaning/delete-api-keys.js --olderThan 1d
+        if: always()
```
louis-bompart, line 21: the 'always' is to try our best ™️ to delete most junk

scripts/cleaning/delete-orgs.js
```diff
@@ -0,0 +1,49 @@
...
+require('abortcontroller-polyfill');
+
+const {homedir} = require('os');
+const {join} = require('path');
+const {config} = require('dotenv');
+const {getClient, yargGenerator, wasCreatedBefore} = require('./utils');
+config({path: join(homedir(), '.env')});
+
+function wasCreatedByTheCli(testRunId = '') {
+  return (key) => key.displayName?.startsWith(`cli-${testRunId}`);
```
y-lakhdar, line 11: ish. that is risky. I created some orgs starting with `cli-`.
I would also check for the description:
`description === 'Generated by the Coveo CLI'`

scripts/cleaning/delete-orgs.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+async function deleteTestOrgs(platform, cliOrgs) {
+  for (const cliOrg of cliOrgs) {
+    const toNotDelete = ['qaregression2', 'connectorsteamtestsmf76kcam'];
+    if (toNotDelete.includes(cliOrg)) {
+      throw new Error('Au bûcher! Au bûcher!');
```
y-lakhdar, line 18: Are we still keeping this XD

scripts/cleaning/delete-orgs.js
louis-bompart: Organization doesn't have descriptions afaik, but because our ID start with id and ends up with g, I can do something like:
```suggestion
  return (key) => testRunId ? key.displayName?.startsWith(`cli-${testRunId}`) : key.displayName?.match(/cli-id.*g/);
```

scripts/cleaning/delete-orgs.js
louis-bompart: We could remove it. Should we? :P

olamothe: Keep it ;)

olamothe (approved):

PR #671, opened 2022-02-14
test: do not mock 'os'

y-lakhdar:
Unit tests are failing in the CI but working locally.
Trying to figure out what is going wrong

image


packages/cli/src/hooks/analytics/identifier.spec.ts
```diff
@@ -29,7 +28,7 @@ describe('identifier', () => {
   const mockUserGet = jest.fn();
   const mockSetIdentity = jest.fn();
   const mockedLogEvent = jest.fn();
-  const mockedOsVersion = jest.mocked(release);
+  const mockedOsVersion = jest.spyOn(os, 'release');
```
y-lakhdar, line 31: For some reasons, when mocking the `os` module, the mock leaks into other packages which causes the test to only fail when ran with Github actions.
Still working to find the root cause, but it definitively started to happen after the Oclif v2 migration.

PR #676, opened 2022-02-16
chore(gh): fix missing import + npm publish token

louis-bompart:
## Proposed changes

- Remove a horrible mistake.
- Prevent said horrible mistake to ever, ever, ever happen again. God damn.
- Fix missing import


y-lakhdar (approved):
haha

louis-bompart:
I added a gitignore just in case, but my test let me believe that the thing does it job properly and move the npmrc in a tempdir
image

PR #677, opened 2022-02-16
fix(oclif): patch oclif to work with npm workspaces

louis-bompart:
https://coveord.atlassian.net/browse/CDX-828

## Proposed changes

1. Fix oclif dev path acquisition to target root node_modules
2. Oclif has some logic to handle yarn workspace (which are very similar to npm workspace). So I patched it so it uses workspace root in anycase. This fix the oclif-manifest generation, which is critical because it's the 'command registry'


louis-bompart:
I ran it against the last release, to be sure:
https://github.com/coveo/cli/actions/runs/1855306112

olamothe (approved):
I 🏩  `patch-package` 😂

y-lakhdar (approved):

PR #679, opened 2022-02-17
chore(pipeline): fix release commit message

louis-bompart:
## Proposed changes

Format commit so that it satisifies the 'version bump' condition for the package rollout certifier ([see Confluence](https://coveord.atlassian.net/wiki/spaces/CM/pages/1786708612/Package+Rollout+Certifier))


louis-bompart:
e.g of commit. Note that the last line with the Jira link is added by the 🐶 husky hooks, so that won't be the case in the CI
image

PR #685, opened 2022-02-22
feat(create-atomic): kill Stencil process, update dependencie & gitignore

ThibodeauJF:
https://coveord.atlassian.net/browse/CDX-833

## Proposed changes
- With Netlify + Stencil, since the port has to always be 3333 for the proxy & Netlify which doens't kill it properly every time
- .gitignore wasn't copied over properly and /www was missing from it
- updated dependencies while at it

## Breaking changes

none

## Testing

Current test already start/stop stencil projects, it should make them less flaky.


olamothe (approved):

louis-bompart (changes requested):
> With Netlify + Stencil, since the port has to always be 3333 for the proxy & Netlify which doens't kill it properly every time

How hard is this 'has to'?

IMO, we should not impose a port on the user if we can.

Also, the `prestart` is a hard no: If the user has a legit program running on either port, we should not kill those processes.
We must never, never, never, kill processes that we did not start.

ThibodeauJF:
I was about 80% sure this solution wouldn't pass 😄 But it's good because I thought of new ones since then

The current situation is, Netlify will detect the framework and run the port according to their config.
https://github.com/netlify/cli/blob/main/docs/netlify-dev.md#project-detection
https://github.com/netlify/framework-info/blob/main/src/frameworks/stencil.json

Another solutions would be to
- play a bit more with the config in order to see why the stencil dev server isn't killed properly, previously played with it but to no good results
- stencil will automatically run to the next port eg 3334 but netlify will still try to access 3333, there might be a way to check if the port is free first and increment if it's not. Then use this port in the config.
- kill the stencil port properly after netlify has been killed

louis-bompart:
> * stencil will automatically run to the next port eg 3334 but netlify will still try to access 3333, there might be a way to check if the port is free first and increment if it's not. Then use this port in the config.

There is, we do that in the other solutions (Angular, Vue, React), that's why I was 😦 pretty hard 😂 
That's probably the best way to go IMO: yes, we 'replicate' the logic made by stencil and whatnot for the port attribution, but if we feed 'em good ports right out the gate, they will just work, and we know what to kill in the end.

packages/create-atomic/template/package.json
```diff
@@ -4,11 +4,14 @@
   "version": "1.25.4",
   "private": true,
   "scripts": {
+    "postinstall": "npm run setup-lambda && npm run setup-cleanup",
```
ThibodeauJF, line 7: Not sure why this disappeared...

packages/create-atomic/src/plopfile.ts
```diff
@@ -66,17 +66,18 @@ export default function (plop: NodePlopAPI) {
           '../template/src/**',
           '../template/scripts/*',
           '../template/.env.example',
-          '../template/.gitignore',
           '../template/tsconfig.json',
+          '../template/package.json',
```
louis-bompart, line 70: no

packages/create-atomic/src/plopfile.ts
```diff
@@ -66,17 +66,18 @@ export default function (plop: NodePlopAPI) {
...
           '../template/stencil.config.ts',
           '../template/netlify.toml',
           '../template/README.md',
+          '../template/ensure-port.js',
         ],
       },
       {
         type: 'add',
-        path: currentPath + '/{{project}}/package.json',
-        templateFile: '../template/package.json.hbs',
```
louis-bompart, line 79: no.

.gitignore
```diff
@@ -27,7 +27,6 @@ yarn.lock
 # Snyk artifacts
 .dccache
 sarifs
-packages/create-atomic/template/package.json.hbs
```
louis-bompart, line 30: no

packages/create-atomic/.prettierignore
```diff
@@ -4,4 +4,4 @@ CHANGELOG.md
 package-lock.json
 lib
 
-template/package.json.hbs
```
louis-bompart, line 7: no

packages/create-atomic/template/package.json
louis-bompart: The removal is normal and should be kept.
The create-atomic-template is considered a package of the monorepo (it helps greatly with dependencies check).
See the https://github.com/coveo/cli/blob/master/packages/create-atomic/scripts/preparePackageJsonTemplate.mjs,
The postInstall is programmatically added to the `hbs` template.

I marked with 'no' every removal that must be reverted.

louis-bompart (changes requested):
addendum: The package.json logic must be kept as it is in `master`:
It is needed for topological release (i.e. we need to release `create-atomic` after `create-atomic-template` has been updated, which can be only done after `search-token-lambda` has been released.
It is also needed for proper dependency analysis by Snyk.

packages/create-atomic/template/package.json
ThibodeauJF: Ok makes more sense now, I had to run this script before testing locally

ThibodeauJF:
New solution checks that the current process running on port 3333 is stencil & is from the current package.

I didn't implement any port finder logic to go to 3334 instead, because of the main issue you would end up with a lot of open ports open.

Also a reminder it's in the generated project, not the cli, so users have full control over this.

louis-bompart:
Ok. Let's put that on hold/reject.
First, we should support custom ports: All the other generated projects are capable of finding and taking other ports automatically. It would be not logical for our customer, nor for us as maintainers to have an exception on that front.

Given the current state, I think it's fair to say that another PR is required prior to this one to support custom ports.

PR #689, opened 2022-02-25
feat: add Stencil + Netlify port management

ThibodeauJF:
https://coveord.atlassian.net/browse/CDX-833

## Proposed changes

Netlify always links the Stencil app to the hardcoded 3333 port which sometimes is not available. We also had the issue of having the Stencil app port remain open. This PR accesses both by discovering a free port & making sure the process is killed.

## Breaking changes
none

## Testing
E2E tests are already running the app a few times in a row.
I'll add a test with an already busy :3333, see if it still works.


packages/create-atomic/template/start-netlify.mjs
```diff
@@ -0,0 +1,33 @@
...
+
+const require = createRequire(import.meta.url);
+const args = process.argv.slice(2);
+
+const main = async () => {
+  // Dynamic Stencil port discovery feature, unsupported by Netlify.
+  const stencilPort = await getPort({port: portNumbers(3333, 3399)});
+
+  process.on('SIGINT', () => {
+    // Intermittent issue with Netlify not shutting down Stencil port properly.
```
ThibodeauJF, line 14: I would open an issue to Netlify about this but I'm not sure what isolates the issue for them to find fix it. Is it the OS, terminal, node version, Netlify or stencil version, something about Atomic itself? Lots of variables 🤷‍♂️

louis-bompart (approved):
LGTM, probably that for the Netlify issue you should compile a minimum reproductible example imo

PR #690, opened 2022-02-25
chore(lockfile): ensure there's only the root lockfile

louis-bompart:
## Proposed changes

With npm workspace, we should have only one package-lock at the root of the mono repo.
This PR removes the other one that has been created by mistake, updates the 'good one' and add a workflow to ensure that this mistake does not happen again


louis-bompart:
It is expected that this PR fail the new check, given that I removed the package-lock.json from create-atomic

y-lakhdar (approved):

PR #699, opened 2022-03-01
add `--createMissingFields` flag

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-854

## Proposed changes

Leverage the `FieldAnalyser` feature (https://github.com/coveo/push-api-client.js/pull/102) in the CLI.
The Field Analyser will be triggered by default, unless specified otherwise.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/cli/src/commands/source/push/add.spec.ts
+++ b/packages/cli/src/commands/source/push/add.spec.ts
@@ -151,6 +151,43 @@ describe('source:push:add', () => {
         'source:push:add',
         'mysource',
         '-f',
+        join(pathToStub, 'jsondocuments', 'batman.json'),
+      ])
+      .it('should create missing fields by default', () => {
+        expect(mockBatchUpdate).toHaveBeenCalledWith(
+          expect.anything(),
+          expect.anything(),
+          expect.anything(),
+          expect.objectContaining({createFields: true})
+        );
+      });
+
+    test
+      .stdout()
+      .stderr()
+      .command([
+        'source:push:add',
+        'mysource',
+        '--no-createMissingFields',
+        '-f',
+        join(pathToStub, 'jsondocuments', 'batman.json'),
+      ])
+      .it('should skip field creation if specified', () => {
+        expect(mockBatchUpdate).toHaveBeenCalledWith(
+          expect.anything(),
+          expect.anything(),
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/source/push/add.spec.ts
```diff
@@ -144,6 +144,43 @@ describe('source:push:add', () => {
...
+        );
+      });
+
+    test
+      .stdout()
+      .stderr()
+      .command([
+        'source:push:add',
+        'mysource',
+        '--no-createMissingFields',
```
y-lakhdar, line 171: I find this very ugly but I guess we have to live with the choice of camel case flags.

olamothe: Why do we have to live with that choice ? 🤔 Because all other flags for other commands are like this ?

olamothe (approved):

packages/cli/src/commands/source/push/add.spec.ts
y-lakhdar: exact. If we want to support kebab-case flags, we need to do the switch manually.
https://github.com/oclif/oclif/issues/107

louis-bompart: I found their arguments toward UX a bit weak, especially with the inconsistency that it causes `--no-`.

We could create an utils to convert the keys of the flags static props from camelCase to kebab-case and comply with the format described in here.
https://github.com/oclif/oclif/issues/738#issuecomment-987157908

I'd consider that a nice-to-have that could be done in another PR.
The converter would have to also add the 'source' keys to the aliases tho, otherwise, it'll be a breaking change.
(And that aliasing could be removed in the future in a major, but I'd avoid doing it 'just' for that.)

louis-bompart (approved):

packages/cli/src/commands/source/push/add.spec.ts
y-lakhdar: I also consider this as a nice to have but I'm not sure we will be able to do it without some kind of patch.
But like you said, we could do it in a future PR.

PR #702, opened 2022-03-02
feat: add privileges preconditions on push command

y-lakhdar:
## Proposed changes

Since unauthorized operations against the Push API simply return an `INVALID_TOKEN` error message, users might not  have enough to figure this issue out.

The proposed solution is to ensure the user has the required privileges to run the Field Analyser and push documents.
 

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


y-lakhdar:
> LGTM 👍
> 
> Do you think we should start splitting packages/cli/src/lib/decorators/preconditions/platformPrivilege.ts or we're still good to go?
> 
> We could do a `packages/cli/src/lib/decorators/preconditions/platformPrivilege/index.ts` That `export * from` all the other files where the privilege are defined.

I think we're good for now. We could always split them later if we need to.

PR #703, opened 2022-03-02
feat: support different regions and environments for push

y-lakhdar:
## Proposed changes

Leverage multi-region and environment support of [push-api-client.js](https://github.com/coveo/push-api-client.js) 

### PR
https://github.com/coveo/push-api-client.js/pull/88

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/commands/source/push/add.ts
```diff
@@ -88,7 +89,11 @@ export default class SourcePushAdd extends Command {
       );
     }
     const cfg = await new AuthenticatedClient().cfg.get();
-    const source = new Source(cfg.accessToken!, cfg.organization);
+    const {environment, region} = this.configuration.get();
```
olamothe, line 92: 🤔  Isn't this information already available on the `cfg` retrieved from `AuthenticatedClient` ?

So this + line 181-183 could be removed.

olamothe (approved):

packages/cli/src/commands/source/push/add.ts
y-lakhdar: yes you're right. my bad

packages/cli/src/commands/source/push/add.spec.ts
```diff
@@ -42,15 +38,15 @@ describe('source:push:add', () => {
...
+  //   mockedConfigGet.mockReturnValue({
+  //     region: 'au',
+  //     organization: 'foo',
+  //     environment: 'prod',
+  //   });
 
-    mockedConfig.prototype.get = mockedConfigGet;
-  };
+  //   mockedConfig.prototype.get = mockedConfigGet;
+  // };
```
louis-bompart, line 49: ?

packages/cli/src/commands/source/push/add.spec.ts
```diff
@@ -84,7 +80,7 @@ describe('source:push:add', () => {
   };
 
   beforeAll(() => {
-    mockConfig();
+    // mockConfig();
```
louis-bompart, line 83: ?

packages/cli/src/commands/source/push/add.spec.ts
y-lakhdar: oups. sorry to remove that.
We no longer need to mock config, since  we are fetching the org/region and env from the authenticated client (which is already moked)

PR #708, opened 2022-03-07
fix(pull): use SPM orgId for privilege preconditions

louis-bompart:
## Proposed changes

- Allows commands to define a `flags` getter to define some custom shenanigans.
- Make sure the preconditions that used `flags` were properly resolving it.
- Made the `pull` command adopt this new behavior and tweak the `flags` getter behavior like so:
 "If the model flags is used, then use its orgId property for the `target` flag"

## Testing

- [x] Unit Tests:
- E2E in #700


packages/cli/src/commands/org/resources/pull.ts
```diff
@@ -217,10 +217,18 @@ export default class Pull extends Command {
   }
 
   private async getTargetOrg() {
-    const {flags} = this.parse(Pull);
+    const flags = this.flags;
     return getTargetOrg(this.configuration, flags.model?.orgId || flags.target);
   }
 
+  public get flags() {
```
olamothe, line 224: It is kind of an obscure parameter that you have to know will be used in the `apiKeyPrivilege` precondition check.

To alleviate that a bit, I would propose to have an interface for this (ie: `export default class Pull extends Command implements FlagsProperty` or whatever name you think might fit better), and then have this interface also be referenced in the `apiKeyPrivilege` module.

The end goal is that we would be able to inspect reference on the interface and discover where it's used.

And that if someone removes or rename the `flags` getter, the code won't compile anymore, and you would get some hint about why the property is important.

This is also a pattern that we already use in many commands I think (ie: `this.flags` getter), so it might be worth it to go and add that interface to all those commands, since it means that the getter is starting to have an impact outside of the commands.

olamothe (approved):

packages/cli/src/commands/org/resources/pull.ts
louis-bompart: Very good idea 👍 I'll do that in another PR to keep the changeset scoped tho.
CDX-869

y-lakhdar (approved):

PR #720, opened 2022-03-21
chore(e2e): stabilize `mitmproxy`-dependent tests

louis-bompart:
## Proposed changes

 - Use Python to start `mitmdump`, this bypasses some shenanigans that the .exe does with its embedded Python 3.10. (Or at least this is what I empirically guessed)
 - Install `mitmdump` with pip: This is nice, more caching, hence, speed.
 
## Testing

CI/FT


packages/cli-e2e/setup/utils.ts
```diff
@@ -153,5 +143,14 @@ export function shimNpm() {
   );
 }
 
+export const resolveBinary = (programName: string) => {
```
y-lakhdar, line 146: Do we still need that?

y-lakhdar (approved):

PR #726, opened 2022-03-28
chore(ci): changes for oclif-v2

louis-bompart:
## Proposed changes

 - Fixed deployment config (not having a Snyk Code section is making it angry, and we ditched Veracode)
 - Shifted from an exclusive to an inclusive listing for packages, `macos` builds were being painful by copying a whole npm arborescence somewhere in the project during its build process, creating duplicates among the packages.
 - Tweaked the release scripts for 'normal operation with Oclif v2'. That doesn't replace the funkeries that need to be manually done for the smooth transition from Oclif v1 to Oclif v2.

## Testing

https://github.com/coveo/cli-v2-temp.
I did 2 releases on the dev environment:
 - 1.25.5 which is the 'transition' one, so Oclif v2 but compat Oclif v1.
 - 1.25.6, a vanilla Oclif v2
 
 I ensured that 1.25.4 and 1.25.5 were able to reach 1.25.6 when running `coveo update` once or twice.


louis-bompart:
Snyk will fail because we don't run a scan on every branches, but is ok, will run on the CDX-746

scripts/github-client.js
```diff
@@ -94,7 +94,6 @@ const downloadReleaseAssets = async (tag, determineAssetLocation) => {
     );
   });
 };
-
 const getSnykCodeAlerts = () => {
```
louis-bompart, line 97:
```suggestion

const getSnykCodeAlerts = () => {
```

y-lakhdar (approved):

olamothe (approved):

PR #727, opened 2022-03-28
test(verdaccio): listen to ipv6

louis-bompart:
## Proposed changes

`minipass-fetch` used by `angular-cli` is now preferring the 'localhost' ipv6, but Verdaccio is not listening to it by default.
So, let's add it 🎉 

Ref: https://verdaccio.org/docs/configuration/#listen-port

## Testing

CI


olamothe (approved):
🙏  Thanks !

PR #728, opened 2022-03-28
feat(create-atomic): allow Coveo atomic to start from an existing Platform Search page

ThibodeauJF:
https://coveord.atlassian.net/browse/CDX-988

SIS implementation https://github.com/coveo/search-interface-service/pull/389
Platform client implementation https://github.com/coveo/platform-client/pull/440

## Proposed changes

Adds the possibility to specify a page id in the prompts, which will call the /manifest endpoint and pull the existing Atomic interface into the new project.

## Breaking changes

None

## Testing

Yes!


packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +65,126 @@ export default function (plop: NodePlopAPI) {
...
-          process.on('close', (code) => {
-            if (code === 0) {
-              resolve('Installation complete');
-            } else {
-              reject(`Installation exited with ${code}`);
-            }
+          try {
+            plopData.page = await fetchPageManifest(
+              // TODO: replace
+              'http://localhost:8222', // plopData['platform-url'],
```
ThibodeauJF, line 96:
```suggestion
              plopData['platform-url'],
```

packages/atomic/create-atomic/src/fetch-page.ts
```diff
@@ -0,0 +1,39 @@
...
+  const response = await fetch(url, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      Authorization: `Bearer ${apiKey}`,
+    },
+    body: JSON.stringify(body),
+  });
+  return replaceResultsPlaceholder((await response.json()) as PageManifest);
+}
```
louis-bompart, line 29: - [ ] Make the route available in coveo/platform-client

packages/atomic/create-atomic/src/default/default-search-interface.ts
```diff
@@ -0,0 +1,39 @@
+export default `<atomic-search-interface>
```
louis-bompart, line 1: Pretty sure you could get nicer linting etc if you just change to .tsx

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -1,9 +1,23 @@
 import {NodePlopAPI} from 'plop';
+import Handlebars from 'handlebars';
 import {spawn} from 'child_process';
 import {getPackageManager} from './utils.js';
+import {fetchPageManifest} from './fetch-page.js';
+import {defaultPageManifest} from './default/default-page.js';
+import fs from 'fs';
```
louis-bompart, line 7:
```suggestion
import fs from 'node:fs';
```

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +65,126 @@ export default function (plop: NodePlopAPI) {
...
+        },
+        function installPackagesPrompt() {
+          return 'Installing packages...';
+        },
+        function installPackages(data) {
+          return new Promise((resolve, reject) => {
+            const {project} = data as PlopData;
+            const process = spawn(getPackageManager(), ['install'], {
+              stdio: 'ignore',
+              cwd: `${currentPath}/${project}/`,
```
louis-bompart, line 161: path.join

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +65,126 @@ export default function (plop: NodePlopAPI) {
...
+            '../template/tsconfig.json',
+            '../template/stencil.config.ts',
+            '../template/netlify.toml',
+            '../template/README.md',
+            '../template/start-netlify.mjs',
+          ],
+        },
+        function generateTemplates(data) {
+          const {page, project} = data as PlopData;
+          page.results.templates.forEach((resultTemplate, index) => {
```
btaillon, line 141: I'm not very familiar with `plop`. Is this function only executed when the `page-id` is provided? If not, where is `page` set when there's no `page-id`?

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +65,126 @@ export default function (plop: NodePlopAPI) {
...
+              'http://localhost:8222', // plopData['platform-url'],
+              plopData['org-id'],
+              plopData['page-id'],
+              plopData['api-key']
+            );
+
+            return `Hosted search page "${plopData.page.config.title}" downloaded`;
+          } catch (error) {
+            throw new Error(
+              `There was an error downloading search page with id "${plopData['page-id']}" from the organization "${plopData['org-id']}"`
```
btaillon, line 105: If the error is that we don't have the privileges to download search pages, I think we should throw a more specific error.

btaillon (approved):

packages/atomic/create-atomic/src/fetch-page.ts
```diff
@@ -0,0 +1,39 @@
...
+  apiKey: string
+) {
+  const url = `${platformUrl}/rest/organizations/${orgId}/searchinterfaces/${pageId}/manifest/v1`;
+  const body: PageManifestBody = {
+    pagePlaceholders: {
+      results: ResultsPlaceholder,
+    },
+  };
+  const response = await fetch(url, {
+    method: 'POST',
```
btaillon, line 21: I'm a bit confused why this is a POST

packages/atomic/create-atomic/scripts/packageTemplate.json
```diff
@@ -12,6 +12,6 @@
     "site:init": "netlify init",
     "site:link": "netlify link",
     "site:deploy": "npm run site:link && npm run build && netlify env:import .env && netlify deploy --prod --open",
-    "postinstall": "npm run setup-lambda && npm run setup-cleanup"
+    "postinstall": "npm run setup-lambda && npm run setup-cleanup && prettier --write . --loglevel warn"
```
louis-bompart, line 15: Just curious, what motivate this change 🤔 
Also, are you sure prettier will be there?

packages/atomic/create-atomic/src/fetch-page.ts
```diff
@@ -0,0 +1,39 @@
...
+    },
+    body: JSON.stringify(body),
+  });
+  return replaceResultsPlaceholder((await response.json()) as PageManifest);
+}
+
+function replaceResultsPlaceholder(pageManifest: PageManifest): PageManifest {
+  const resultManagerComponent = '<results-manager></results-manager>';
+  pageManifest.markup = pageManifest.markup.replace(
+    pageManifest.results.placeholder!,
```
louis-bompart, line 34: 😕 why is there a non-null assertion operator?
If there's always a placeholder, then the interface should reflect that.

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +65,126 @@ export default function (plop: NodePlopAPI) {
...
-        path: currentPath + '/{{project}}/package.json',
-        templateFile: '../template/package.json.hbs',
-      },
-      function installPackagesPrompt() {
-        return 'Installing packages...';
+        type: 'input',
+        name: 'page-id',
+        default: '',
+        message:
+          '(Optional) The hosted search page ID, e.g. "7944ff4a-9943-4999-a3f6-3e81a7f6fb0a".',
```
louis-bompart, line 73: In terms of UX, how does one pick up its page ID?
(assuming the user go thru the Coveo CLI)

packages/atomic/create-atomic/src/fetch-page.ts
```diff
@@ -0,0 +1,39 @@
+import fetch from 'node-fetch';
```
louis-bompart, line 1: Use Axios, or you'll have to quell the wrath of the HTTPS Proxy (it's not fun: https://github.com/coveo/cli/pull/512)

packages/cli/src/commands/ui/create/atomic.ts
```diff
@@ -87,6 +93,8 @@ export default class Atomic extends Command {
       username,
     ];
 
+    this.args.pageId && cliArgs.push('--page-id', this.args.pageId);
```
louis-bompart, line 96: Yeah I know, I know, this is common in React, but
image
```suggestion
    if (this.args.pageId) {
      cliArgs.push('--page-id', this.args.pageId);
    }
```

louis-bompart (commented):
looks nice overall, still some changes regarding the network call.
 - [x] Move the call to the `platform-client`, afaik/imho, there are no reasons for not doing so as soon those conditions are fulfilled:
   - The call is available outside the intranet
   - The Admin UI or the CLI (or another of their JS components) need it.

This is important to prevent code duplication across the company.

- [x] Support proxy.

packages/atomic/create-atomic/scripts/packageTemplate.json
ThibodeauJF: Yes prettier is installed!

Especially when downloading the markup from the SIS, it's a way better experience to have everything formatted properly!

packages/atomic/create-atomic/src/fetch-page.ts
ThibodeauJF: There's a body with the placeholders

packages/atomic/create-atomic/scripts/packageTemplate.json
louis-bompart: Oh yeah true, ["prettier": "2.6.0",](https://github.com/coveo/cli/blob/d48e2c9d09430a8feb6baff4ad8c867ef0d97969/packages/atomic/template/package.json#L20) ma bad.

packages/atomic/create-atomic/src/plopfile.ts
ThibodeauJF: Not very good at the moment, the page id is contained in the url. It would be easy to add a new table action in the admin to copy the id though!

packages/atomic/create-atomic/src/fetch-page.ts
louis-bompart: Hmm then, if I'm not mistaken you cannot be sure that `placeholder` is actually defined when this statement is reached, correct?

packages/atomic/create-atomic/src/plopfile.ts
louis-bompart: 👍 imma adding that to the list of features that would benefit from a button to open the CLI.

packages/atomic/create-atomic/src/default/default-search-interface.ts
ThibodeauJF: It's a string literal, not JSX. It doesn't give me better lint/format

ThibodeauJF: I could change these files to html and import it as a string perhaps

louis-bompart: Worth a try :D

ThibodeauJF: Actually don't think I can do this with only the Typescript compiler, going to need esbuild or rollup, webpack, etc.

louis-bompart: Yeah, so let's keep it as is for now 😅

packages/atomic/create-atomic/src/plopfile.ts
ThibodeauJF: Will handle in platform client

packages/atomic/create-atomic/src/fetch-page.ts
```diff
@@ -0,0 +1,44 @@
...
+  if (proxyServer) {
+    const httpsProxyAgent = HttpsProxyAgent(proxyServer);
+    globalRequestSettings.agent = httpsProxyAgent;
+  }
+
+  const client = new PlatformClient({
+    globalRequestSettings,
+    organizationId,
+    accessToken,
+    host,
```
ThibodeauJF, line 23: @louis-bompart Is using the host ok, or we need to map the platform url in some way? 

Also notice the use of PlatformClient + proxy management. It will take a bit till everything is ready but locally I manage to have the full loop working:
Screen Shot 2022-04-04 at 4 03 39 PM
Screen Shot 2022-04-04 at 4 04 11 PM
Screen Shot 2022-04-04 at 4 04 38 PM

packages/atomic/create-atomic/src/plopfile.ts
ThibodeauJF: - [ ] Add prompt list to "pick your page"

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +82,151 @@ export default function (plop: NodePlopAPI) {
...
 
-        To share your site with the world:
-        > npm run site:deploy
-    
-        Happy hacking!
-        `;
-      },
+          return `The search page with the id "${input}" does not exist.`;
+        },
+      } as PromptQuestion,
```
ThibodeauJF, line 114: Needed to make the custom ListPrompt work with plop

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -1,13 +1,44 @@
...
+    platformClient = createPlatformClient(
+      answers['platform-url'],
+      answers['org-id'],
+      answers['api-key']
+    );
+  }
+
+  plop.setHelper('inc', (value) => parseInt(value) + 1);
+
+  plop.setPrompt('customList', ListPrompt);
```
ThibodeauJF, line 40: https://github.com/coveo/cli/pull/728/files#r877449481

packages/atomic/create-atomic/src/fetch-page.ts
```diff
@@ -0,0 +1,36 @@
...
+      results: '--results--',
+    },
+  });
+  return replaceResultsPlaceholder(manifest);
+}
+
+function replaceResultsPlaceholder(manifestResponse: IManifest) {
+  const resultManagerComponent = '<results-manager></results-manager>';
+  if (manifestResponse.results.placeholder) {
+    manifestResponse.markup = manifestResponse.markup.replace(
```
louis-bompart, line 29: Can you have multiple occurrences of the placeholder in one markup?
If so, should we not replace'em all?

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +82,151 @@ export default function (plop: NodePlopAPI) {
...
-        return 'Installing packages...';
-      },
-      function installPackages(answers) {
-        return new Promise((resolve, reject) => {
-          const {project} = answers as PromptsAnswers;
-          const process = spawn(getPackageManager(), ['install'], {
-            stdio: 'ignore',
-            cwd: `${currentPath}/${project}/`,
-          });
+        // Custom type necessary to allow bypassing async choices
```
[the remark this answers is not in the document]
louis-bompart: From what I saw, plop is really taking and reexposing inquirer typings.
Are you sure the promises are properly awaited?

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +82,130 @@ export default function (plop: NodePlopAPI) {
...
+              currentPath,
+              project,
+              'src',
+              'components',
+              'results-manager',
+              `template-${index + 1}.html`
+            );
+            writeFileSync(
+              filePath,
+              plop.renderString('{{{markup}}}', resultTemplate)
```
louis-bompart, line 168: Use prettier here ;)
1. Move prettier from `devDependencies` to `dependencies`
2. `import {format} from 'prettier';` somewhere uptop this file.
3.
```suggestion
              format(plop.renderString('{{{markup}}}', resultTemplate), {parser:'html'}
```

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -51,72 +82,130 @@ export default function (plop: NodePlopAPI) {
...
+            const process = spawn(getPackageManager(), ['install'], {
+              stdio: 'ignore',
+              cwd: join(currentPath, project),
+            });
+
+            process.on('close', (code) => {
+              if (code === 0) {
+                resolve('Installation complete');
+              } else {
+                reject(`Installation exited with ${code}`);
```
louis-bompart, line 189: While at it, if you can bubble up the error or stderr, it'll help with maintenance and bugz hammering

louis-bompart (commented):
I'm a bit :suspect: about the custom prompt, please double-check that your promises are properly awaited.

Other than that, great job, it's pretty clean 👏

packages/atomic/create-atomic/src/plopfile.ts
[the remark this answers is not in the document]
ThibodeauJF: Yes it's properly awaited since I can see the list of pages that are fetched asynchronously. 

In the first comment you can see the (bad) validation  plop does to prevent me from using this, even if it works, since plop uses inquirer.

packages/atomic/create-atomic/src/plopfile.ts
ThibodeauJF: Already the case 😄  You can see this in stderr, plop does that for us

packages/atomic/create-atomic/src/fetch-page.ts
ThibodeauJF: Only one

louis-bompart (dismissed):
Add a check for permission of the API Key..

packages/atomic/create-atomic/src/list-pages.ts
```diff
@@ -0,0 +1,8 @@
+import {PlatformClient} from '@coveord/platform-client';
+
+export async function listSearchPagesOptions(client: PlatformClient) {
+  return (await client.searchInterfaces.list({})).items.map((page) => ({
```
olamothe, line 4: What are the default `list` params in relation to pagination ?

Could it cause issues if you have too much search pages configured in a single org, that we would not be able to pick it up ?

packages/atomic/template/src/pages/index.html.hbs
```diff
@@ -0,0 +1,24 @@
...
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>{{titleCase project}}</title>
+    <link href="/atomic/themes/coveo.css" rel="stylesheet" />
+    <link href="/build/{{project}}.css" rel="stylesheet" />
+    <!-- 
+      When using the Coveo CDN - Make sure you're using the same minor version as the packaged Atomic.
+      E.g., if you have "@coveo/atomic@1.44.3" installed, use the "/atomic/v1.44/" path.
+    
+      <script type="module" src="https://static.cloud.coveo.com/atomic/v1.MINOR/atomic.esm.js"></script>
```
olamothe, line 13:
```suggestion
      <script type="module" src="https://static.cloud.coveo.com/atomic/vMAJOR.MINOR/atomic.esm.js"></script> 
```

olamothe (approved):
GG

y-lakhdar (approved):
Nice!

packages/atomic/create-atomic/src/list-pages.ts
ThibodeauJF: It's 150 per page with page=0. We could put a pagination system to try to fetch more than 150, but it might be unrealistic?

packages/atomic/create-atomic/scripts/packageTemplate.json
```diff
@@ -17,6 +17,6 @@
...
```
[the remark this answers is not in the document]
ThibodeauJF: We'd have to do this for all files that contain markup , not just the one we writefile with. It would have to be on each plop method with a `transform` perhaps.

I'd rather do it on the project as a whole once it's installed. Then the user still has prettier installed & configured in his to format his project as needed, which is a plus. End result is the same too.

What do you think?

ThibodeauJF:
@louis-bompart  Getting this error since I added the check 
```
[18:46:34]Precondition Error - Missing Platform Privilege: Your access token is missing the View search pages privilege. Make sure to grant this privilege to your API key before running the command again.
[18:46:34]See https://docs.coveo.com/en/1707/manage-an-organization/privilege-reference#search-pages-domain.
```

I thought the e2e key already had those privileges otherwise it wouldn't previously work?

packages/atomic/create-atomic/scripts/packageTemplate.json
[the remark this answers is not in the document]
louis-bompart: Yeah sure then, than make sense

packages/cli/src/lib/platform/authenticatedClient.spec.ts
```diff
@@ -156,7 +156,13 @@ describe('AuthenticatedClient', () => {
           description: 'Generated by the Coveo CLI',
           enabled: true,
           privileges: [
-            {targetDomain: 'IMPERSONATE', targetId: '*', owner: 'SEARCH_API'},
+            {owner: 'SEARCH_API', targetDomain: 'IMPERSONATE', targetId: '*'},
```
louis-bompart, line 159: I think you should add a second optional parameter `additionnalPrivileges` for `createImpersonateApiKey` that takes an array of privileges.

Because currently, the changes you're doing to the method is affecting every `ui:create` commands in the end, despite none of them other than the atomic one needing this right.

louis-bompart (approved):

PR #735, opened 2022-03-30
chore(ci): do not run pip on CI

louis-bompart:
https://coveord.atlassian.net/browse/CDX-903

## Proposed changes

Forgot a `ignore-script` for the CD stage w/ yarn.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install dependencies
         run: |
           npm ci
-          yarn install --no-immutable
+          yarn install --no-immutable --mode=skip-build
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Install dependencies
         run: |
           npm ci
-          yarn install --no-immutable
+          yarn install --no-immutable --mode=skip-build
```
louis-bompart, line 25: same as https://github.com/coveo/cli/blob/master/.github/workflows/snyk-master.yml#L19

olamothe (approved):

PR #736, opened 2022-03-30
chore(release): ignore `npm-shim` for snyk

louis-bompart:
## Proposed changes

Snyk should not scan this package.json.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,13 +25,13 @@ jobs:
           yarn install --no-immutable --mode=skip-build
       - name: Snyk Test
         run: |
-          npx snyk monitor      --dev --yarn-workspaces --exclude="project.json"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -25,13 +25,13 @@ jobs:
           yarn install --no-immutable --mode=skip-build
       - name: Snyk Test
         run: |
-          npx snyk monitor      --dev --yarn-workspaces --exclude="project.json"
-          npx snyk test         --dev --yarn-workspaces --exclude="project.json"
+          npx snyk monitor      --dev --yarn-workspaces --exclude="project.json,npm-shim"
```
louis-bompart, line 28: Same as https://github.com/coveo/cli/blob/master/.github/workflows/snyk-master.yml#L32

PR #740, opened 2022-03-31
chore(cd): fix binaries matcher

louis-bompart:
## Proposed changes

macos & linux doesn't follow exactly the same pattern as windows, so this fix that.
THis has been used in the current release with the branch release-temp-locked

### DIFF
--- a/scripts/get-release-artifact-and-commit-sha.js
+++ b/scripts/get-release-artifact-and-commit-sha.js
@@ -19,7 +19,7 @@ async function main() {
     'stable',
   ].join('/');
   const binariesMatcher =
-    /^coveo[_-]{1}(?<_version>v?\d+\.\d+\.\d+(-\d+)?)[_-]{1}(?<_commitSHA>\w{7})[_-]{1}(?<longExt>.*\.(exe|deb|pkg))$/;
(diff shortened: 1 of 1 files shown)


louis-bompart:
> > macos & linux doesn't follow exactly the same pattern as windows, so this fix that.
> 
> You mean windows does not follow the same pattern as every one else :P

Well, if I was accurate, I'd say they all do their own stuff, 'cause linux and macos doesn't behave the same too.

scripts/get-release-artifact-and-commit-sha.js
```diff
@@ -19,7 +19,7 @@ async function main() {
     'stable',
   ].join('/');
   const binariesMatcher =
-    /^coveo[_-]{1}(?<_version>v?\d+\.\d+\.\d+(-\d+)?)[_-]{1}(?<_commitSHA>\w{7})[_-]{1}(?<longExt>.*\.(exe|deb|pkg))$/;
+    /^coveo[_-]{1}(?<_version>v?\d+\.\d+\.\d+(-\d+)?)[_.-]{1}(?<_commitSHA>\w{7})[_-]{0,1}(\d+_)?(?<longExt>.*\.(exe|deb|pkg))$/;
```
y-lakhdar, line 22: could this happen **`(\d+[_-]{1})`** ?

y-lakhdar (approved):

scripts/get-release-artifact-and-commit-sha.js
louis-bompart: I did not see it yet. If it happens, well, we'll modify it again gee

PR #741, opened 2022-03-31
chore(cd): ignore-scripts + update node

louis-bompart:
## Proposed changes

I _think_ that will fix the issue that occured with the JenkinsfileCloudfrontInvalidation, cannot really be sure tho.


y-lakhdar (approved):
🤞

PR #744, opened 2022-04-01
chore(react): lock other react deps

louis-bompart:
## Proposed changes

Other dependencies are marked as a major update because they drop support for React 17. So we locked 'em up.


louis-bompart:
Fun fact: #742, my OG PR has been rejected... by the god darn bot.

y-lakhdar (approved):

y-lakhdar (changes requested):
Oups, I approved too fast.
There is still a peer dependency with React to fix

y-lakhdar (approved):

PR #745, opened 2022-04-01
chore(cd): await when updating versions in workspaces

louis-bompart:
https://coveord.atlassian.net/browse/CDX-910

## Proposed changes

Node was queuing up the microtask 'update workspace dependencies' but was not awaiting it, so the `npm publish` was occurring before the update was done.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,6 +55,8 @@ jobs:
         with:
           node-version: '16'
           registry-url: 'https://registry.npmjs.org'
+      # TODO: Remove later, when `npm 8.5.4` is built in.
+      - run: npm i -g npm@^8.5.4
       - name: Install dependencies
         run: npm ci
       - name: Setup credentials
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -55,6 +55,8 @@ jobs:
         with:
           node-version: '16'
           registry-url: 'https://registry.npmjs.org'
+      # TODO: Remove later, when `npm 8.5.4` is built in.
+      - run: npm i -g npm@^8.5.4
```
louis-bompart, line 59: This allow me to remove the --no-package-lock, which actually generate a more accurate package-lock. (still some stuff to fix, but all minor, can wait)

scripts/releaseV2/phase2-bump-all-packages.mjs
```diff
@@ -46,19 +46,7 @@ const rootFolder = resolve(dirname(fileURLToPath(import.meta.url)), '..', '..');
...
-    cwd: undefined,
-  });
-  publish.stdout
-    .toString()
-    .split('\n')
-    .forEach((line) => console.log(line));
-  publish.stderr
-    .toString()
-    .split('\n')
-    .forEach((line) => console.error(line));
```
louis-bompart, line 61: We were publishing twice. Luckily, the second one ain't throwing the main thread, so the release was working, despite it always failing.

y-lakhdar (approved):

PR #749, opened 2022-04-05
feat: support stream API

y-lakhdar:
## Proposed changes

* Command to stream data to a catalog source (`source:catalog:add`)
* Command to create a catalog (`source:catalog:new`)
* Refactoring since the @coveo/push-api-client bump introduces breaking changes.

## Dependencies
Merge these PR first
- [x] #783 

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/commands/source/catalog/add.ts
```diff
@@ -0,0 +1,160 @@
+import {UploadBatchCallbackData, CatalogSource} from '@coveo/push-api-client';
```
y-lakhdar, line 1: @jpmarceau, I will need your stamp of approval for this new file since this command will print a lot of messages

packages/cli/src/commands/source/catalog/add.ts
```diff
@@ -0,0 +1,160 @@
...
+export default class SourceCatalogAdd extends Command {
+  public static description =
+    'Index a JSON document into a Coveo Catalog source. See https://docs.coveo.com/en/2956/coveo-for-commerce/index-commerce-catalog-content-with-the-stream-api for more information.';
+
+  public static flags = {
+    ...withFile(),
+    ...withFolder(),
+    ...withMaxConcurrent(),
+    ...withCreateMissingFields(),
+    fullUpload: Flags.boolean({
```
y-lakhdar, line 36: By default, adding documents to a catalog source will trigger a [full document update](https://docs.coveo.com/en/l62e0540/coveo-for-commerce/full-document-update). This will act the same way as pushing stuff to a push source.

However, we recommend to perform an initial catalog update first. However, since this operation overwrites the entire source content, I was not confortable making this the default operation.

That's why I later check if the source contains items (line 87) to better guide the user.

packages/cli/src/commands/source/catalog/add.ts
```diff
@@ -0,0 +1,160 @@
...
+      default: false,
+      description: `Controls the way your items are added to your catalog source.
+
+      Setting this option to ${bold(
+        'false'
+      )} will trigger a document update (Default operation). Useful for incremental updates for smaller adjustments to your catalog without the need of pushing the entire catalog. A document update must only be performed after a full catalog upload.
```
jpmarceau, line 42:
```suggestion
      )} will trigger a document update (Default operation). Useful to perform incremental updates for smaller adjustments to your catalog that do not require pushing the entire catalog. A document update must only be performed after a full catalog upload.
```

packages/cli/src/commands/source/catalog/add.ts
```diff
@@ -0,0 +1,160 @@
...
+
+      Setting this option to ${bold(
+        'true'
+      )} will trigger a full catalog upload will start. This process acts as a full rebuild of your catalog source, therefore if the payload doesn't contain all the items, previous items will be deleted.
+      See https://docs.coveo.com/en/lb4a0344
+        `,
+    }),
+    skipFullUploadCheck: Flags.boolean({
+      default: false,
+      description: `Do not check if a catalog full load was triggered on the target source.`,
```
jpmarceau, line 53:
```suggestion
      description: `Do not check whether a full catalog upload was triggered on the target source.`,
```

packages/cli/src/commands/source/catalog/add.ts
```diff
@@ -0,0 +1,160 @@
...
+} from '../../../lib/flags/sourceCommonFlags';
+import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
+import {errorMessage, successMessage} from '../../../lib/push/userFeedback';
+import {getFileNames} from '../../../lib/utils/file';
+import {bold} from 'chalk';
+import dedent from 'ts-dedent';
+
+export default class SourceCatalogAdd extends Command {
+  public static description =
+    'Index a JSON document into a Coveo Catalog source. See https://docs.coveo.com/en/2956/coveo-for-commerce/index-commerce-catalog-content-with-the-stream-api for more information.';
```
jpmarceau, line 29:
```suggestion
    'Index a JSON document into a Coveo Catalog source. See https://docs.coveo.com/en/2956 for more information.';
```
(everything after the ID is ignored by the docs site, so we can shorten this. It means losing some information, but I think it's worth it for brevity)

jpmarceau (commented):
👍

jpmarceau (approved):
👍

louis-bompart (commented):
This is getting big, you should split the update/refactor of the push from the addition of the stream API.

louis-bompart (changes requested):
split refactor/update of push-api-client into another PR

packages/cli/src/commands/source/catalog/new.ts
```diff
@@ -0,0 +1,57 @@
...
+  public static args = [
+    {
+      name: 'name',
+      description: 'The name of the source to create.',
+      required: true,
+    },
+  ];
+
+  @Trackable()
+  @Preconditions(IsAuthenticated())
```
olamothe, line 30: Should there be a privilege check for "can create push source?"

olamothe (approved):

packages/cli/src/commands/source/catalog/new.ts
y-lakhdar: Good catch. I will address that in CDX-932

packages/cli/src/lib/utils/file.ts
```diff
@@ -0,0 +1,27 @@
...
+        .flatMap((folder) =>
+          readdirSync(folder, {withFileTypes: true}).map((dirent) =>
+            dirent.isFile() ? join(folder, dirent.name) : null
+          )
+        )
+        .filter(isString)
+    );
+  }
+  return fileNames;
+}
```
louis-bompart, line 27:
```suggestion
export async function getFileNames(entries: FilesOrFolders) {
  return [...(entries.file ?? []), ...(entries.folder ?? [])].flatMap(
    getFilenamesRecursively
  );
}

function getFilenamesRecursively(entry: string): string[] {
  const entryStat = statSync(entry, {throwIfNoEntry: false});
  if (!entryStat) {
    return [];
  }
  if (entryStat.isDirectory()) {
    return readdirSync(entry, {withFileTypes: true}).flatMap((subEntry) => {
      const subEntryPath = join(entry, subEntry.name);
      return subEntry.isDirectory()
        ? getFilenamesRecursively(subEntryPath)
        : [subEntryPath];
    });
  }
  if (entryStat.isFile()) {
    return [entry];
  }
  return [];
}
```

louis-bompart (changes requested):
Ideally, we would not introduce the directory flag in the stream API command, given that we're gonna kill it off

https://coveo.slack.com/archives/G01L9SYPNS0/p1650478552359219?thread_ts=1650468613.515809&cid=G01L9SYPNS0

PR #750, opened 2022-04-05
feat: list catalog and sources

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-912

## Proposed changes

The source list command is no longer under the push topic.
image

And when running the list command, we should now see the source type
image

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/commands/source/push/list.ts
```diff
@@ -7,10 +7,13 @@ import {
 import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
 import dedent from 'ts-dedent';
 import {Trackable} from '../../../lib/decorators/preconditions/trackable';
+import {magenta} from 'chalk';
+import SourceList from '../list';
 
 export default class SourcePushList extends Command {
```
louis-bompart, line 13:
```suggestion
/**
 * @todo CDX-917: Remove file.
 * @deprecated
**/ 
export default class SourcePushList extends Command {
```

louis-bompart (commented):
lgtm, add a todo/deprecate comment in the deprecated class.

louis-bompart (approved):

olamothe (approved):

PR #751, opened 2022-04-06
test(create-atomic): validate that files are created & deleted properly

ThibodeauJF:
https://coveord.atlassian.net/browse/CDX-913

**Why?**
- This test will be even more useful for the CLI phase 2.
- We had problems like not copying the .gitignore file properly in the past.


packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -88,6 +90,59 @@ describe('ui:create:atomic', () => {
...
+      '.gitignore',
+      '.env',
+      '.env.example',
+      'README.md',
+      'start-netlify.mjs',
+      'netlify.toml',
+      'tsconfig.json',
+      'stencil.config.ts',
+      'lambda',
+      'src/index.ts',
```
ThibodeauJF, line 118: Could be an array, but the use of `/` is easier to read IMO and keep in sync with the paths.

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -10,6 +10,8 @@ import {jwtTokenPattern} from '../utils/matcher';
 import {EOL} from 'os';
 import {DummyServer} from '../utils/server';
 import {loginWithApiKey} from '../utils/login';
+import {existsSync, ensureFileSync} from 'fs-extra';
+import {join} from 'path';
```
ThibodeauJF, line 14: should it be node:path ?

louis-bompart: could be, but the test aren't ESM so it's not that important

louis-bompart (approved):

PR #753, opened 2022-04-12
fix(create-atomic): prevent Atomic/Headless version mismatch & use installed Atomic version

ThibodeauJF:
CDX-XXX 😢  want me to create a JIRA in this project instead?

https://coveord.atlassian.net/browse/KIT-1515

## Proposed changes

some general context here https://github.com/coveo/ui-kit/pull/1885

## Breaking changes

None, only for new CLI generated projects

## Testing

Current tests should still pass


packages/atomic/create-atomic/scripts/packageTemplate.json
```diff
@@ -3,6 +3,11 @@
   "description": "{{titleCase project}} project",
   "version": "0.1.0",
   "private": true,
+  "main": "dist/index.cjs.js",
+  "module": "dist/index.js",
+  "types": "dist/types/index.d.ts",
+  "collection": "dist/collection/collection-manifest.json",
+  "collection:main": "dist/index.js",
```
ThibodeauJF, line 10: Once you have dist/dist-custom-element output targets, Stencil will instruct the user to update their package.json file

packages/atomic/template/stencil.config.ts
```diff
@@ -5,13 +5,27 @@ import html from 'rollup-plugin-html';
 // https://stenciljs.com/docs/config
 
 export const config: Config = {
+  namespace: '{{project}}',
```
ThibodeauJF, line 8: necessary once you have dist & dist-custom-elements output targets

packages/atomic/template/stencil.config.ts
```diff
@@ -5,13 +5,27 @@ import html from 'rollup-plugin-html';
...
+          dest: 'atomic',
+          keepDirStructure: false,
+        },
+      ],
+    },
+    {
+      type: 'dist',
+    },
+    {
+      type: 'dist-custom-elements',
```
ThibodeauJF, line 28: Adding these because I had some feedback that some users don't know how to bundle their custom components. Adding these targets (https://stenciljs.com/docs/distribution) will help them bundle their components and deploy them. Some additional docs will have to be written for that.

packages/atomic/template/tsconfig.json
```diff
@@ -12,7 +12,9 @@
     "noUnusedParameters": true,
     "jsx": "react",
     "jsxFactory": "h",
-    "types": ["@coveo/atomic"]
+    "paths": {
+      "@coveo/headless": ["./node_modules/@coveo/atomic/headless"]
+    }
```
ThibodeauJF, line 17: Prevents any mismatch, and getting "double suggestions" for both `coveo/atomic/headless` & `coveo/headless`

louis-bompart:
> CDX-XXX 😢 want me to create a JIRA in this project instead?

Nah should be fine, or just add a subtask on the KIT Jira.

packages/atomic/template/tsconfig.json
louis-bompart: If I'm understanding this properly, this won't be necessary on the next major release of atomic, because the typing will go away, correct?

packages/atomic/create-atomic/scripts/packageTemplate.json
```diff
@@ -3,6 +3,11 @@
   "description": "{{titleCase project}} project",
   "version": "0.1.0",
   "private": true,
+  "main": "dist/index.cjs.js",
```
louis-bompart, line 6: `cjs.js` wat?
do you mean `.cjs`?

packages/atomic/template/src/pages/index.html
```diff
@@ -4,12 +4,17 @@
...
+    <!-- 
+      When using the Coveo CDN - Make sure you're using the same minor version as the packaged Atomic.
+      E.g., if you have "@coveo/atomic@1.44.3" installed, use the "/atomic/v1.44/" path.
+    
+      <script type="module" src="https://static.cloud.coveo.com/atomic/v1.MINOR/atomic.esm.js"></script> 
+    -->
+    <script type="module" src="/atomic/atomic.esm.js"></script>
+    <!-- Custom Elements Lazy Loading -->
+    <script type="module" src="/build/{{project}}.esm.js"></script>
+    <!-- Global initialization script -->
```
louis-bompart, line 17: I think we could do both:
index.html can stay like this
index-cdn.html for the CDN-ed version?

packages/atomic/create-atomic/scripts/packageTemplate.json
ThibodeauJF: No! That's what Stencil outputs and recommends when you don't have the main set
Screen Shot 2022-04-13 at 9 57 24 AM
Screen Shot 2022-04-13 at 9 56 54 AM

packages/atomic/template/src/pages/index.html
ThibodeauJF: I don't want the users to maintain 2 html pages, I'd rather just have this comment and have them make the decision early and stick to it. The CDN is more risky when building custom components, I'd rather push for a better dev setup OOTB.

The index.html is also the one that's loaded when starting the dev server, users might be confused?

packages/atomic/template/tsconfig.json
ThibodeauJF: In the next major, we plan not to have `@coveo/atomic/headless`. 

It will still be bundled in the CDN for this use case (which was why I bundled Headless with Atomic initially) https://docs.coveo.com/en/atomic/latest/usage/headless-through-atomic/ 
Looking back, having it has an NPM subpackage & a dependency at the same time was a mistake.

I can remove the paths config for now, the typing issue is not as bad now that we have the same Atomic/Headless & Headless NPM versions.

packages/atomic/create-atomic/scripts/packageTemplate.json
louis-bompart: oh ok that's funny.

packages/atomic/template/src/pages/index.html
louis-bompart: mh okok, looks good enough for now.
One thing that we might wanna check later on is how to help user go from their development environment to prod.
I reckon that a user that deploy its website should use the CDN in its deployment for optimum perfs, correct?

ThibodeauJF: > One thing that we might wanna check later on is how to help user go from their development environment to prod.

Yes it's the phase 3!

> I reckon that a user that deploy its website should use the CDN in its deployment for optimum perfs, correct?

Yes for prod not having to serve these files is better from the customer's perspective

olamothe (approved):

PR #756, opened 2022-04-14
chore(sec-71): dangerouselySetInnerHTML surewhynot react, do your thing

louis-bompart:
Self-explanatory. GG if CI green.


packages/atomic/template/src/components/results-manager/results-manager.tsx
```diff
@@ -14,12 +14,12 @@ export class ResultsManager {
     return (
       <atomic-result-list fields-to-include="">
         <atomic-result-template>
-          <template innerHTML={defaultTemplate}></template>
+          <template dangerouslySetInnerHTML={defaultTemplate}></template>
```
olamothe, line 17: I don't understand what's this change for... 🤔 The title of the PR talks about React, but this is an Atomic related package/template.

louis-bompart: You're right. I've been confused by the `.tsx`, because so far in my mind JSX/TSX was really React specific.
So, I looked that up and yeah, definitely, that's a false positive.

I'll ignore the report from SnykCode in that regard.

At least this PR will have allowed to properly address it :D

PR #766, opened 2022-04-22
feat(snapshot): add vault error handling

louis-bompart:
- Moved all parsing of the report to the `snapshotReporter` (notably for errors).
From this parsing, the reporter now also triggers the calls relative to the status of a report:
Before, each classes implementing the reporter were doing their own 'if-this-then-that'. Now, the 'if-this-then' onus is shifted to the reporter, and the implementer just has to focus on the 'that'.

- Detected missing vault entries among resourceInErrors.
- Ignored missing vault entries for the 'generic' error and handled them separately.

Most of the tests were green from the get-go but were susceptible to breaking because of some scope being too wide (i.e. lack of mocking), I took the chance to refactor some of them and more importantly add a lot of coverage.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,6 +5,13 @@
   "version": "0.2.0",
   "configurations": [
     {
+      "name": "Attach by Process ID",
+      "processId": "${command:PickProcess}",
+      "request": "attach",
+      "skipFiles": ["<node_internals>/**"],
+      "type": "node"
+    },
+    {
       "type": "pwa-node",
       "request": "attach",
       "cwd": "${workspaceFolder}",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,13 +5,6 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Attach by Process ID",
-      "processId": "${command:PickProcess}",
-      "request": "attach",
-      "skipFiles": ["<node_internals>/**"],
-      "type": "node"
-    },
-    {
       "type": "pwa-node",
       "request": "attach",
       "cwd": "${workspaceFolder}",
--- a/packages/cli/src/commands/org/resources/__snapshots__/monitor.spec.ts.snap
+++ b/packages/cli/src/commands/org/resources/__snapshots__/monitor.spec.ts.snap
@@ -1,0 +1,126 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`org:resources:monitor when the operation conclude with no changes should output the same thing 1`] = `
+"  console.error
+     »   Warning: The org:resources commands are currently in public beta, please 
+     »   report any issue to github.com/coveo/cli/issues
+
+      43 |   @Preconditions(IsAuthenticated())
+      44 |   public async run() {
+    > 45 |     this.warn(
+         |          ^
+      46 |       'The org:resources commands are currently in public beta, please report any issue to github.com/coveo/cli/issues'
+      47 |     );
+      48 |     await this.printHeader();
+
+      at Object.warn (../../../node_modules/@oclif/core/lib/errors/index.js:61:13)
+      at Monitor.warn (../../../node_modules/@oclif/core/lib/command.js:86:20)
+      at Monitor.apply (commands/org/resources/monitor.ts:45:10)
+      at Monitor.descriptor.value (lib/decorators/preconditions/preconditions.ts:19:32)
+      at Monitor.trackCommand (lib/decorators/preconditions/trackable.ts:63:3)
+      at Monitor.descriptor.value (lib/decorators/preconditions/trackable.ts:46:9)
+      at Monitor._run (../../../node_modules/@oclif/core/lib/command.js:67:22)
+
+
+"
+`;
+
+exports[`org:resources:monitor when the operation conclude with no changes should output the same thing 2`] = `
+"
+Monitoring snapshot my-snapshot...
+Operation undefined..."
+`;
+
+exports[`org:resources:monitor when the operation fails should output the same thing 1`] = `
+"  console.error
+     »   Warning: The org:resources commands are currently in public beta, please 
+     »   report any issue to github.com/coveo/cli/issues
+
+      43 |   @Preconditions(IsAuthenticated())
+      44 |   public async run() {
+    > 45 |     this.warn(
+         |          ^
+      46 |       'The org:resources commands are currently in public beta, please report any issue to github.com/coveo/cli/issues'
+      47 |     );
+      48 |     await this.printHeader();
+
+      at Object.warn (../../../node_modules/@oclif/core/lib/errors/index.js:61:13)
+      at Monitor.warn (../../../node_modules/@oclif/core/lib/command.js:86:20)
+      at Monitor.apply (commands/org/resources/monitor.ts:45:10)
+      at Monitor.descriptor.value (lib/decorators/preconditions/preconditions.ts:19:32)
+      at Monitor.trackCommand (lib/decorators/preconditions/trackable.ts:63:3)
+      at Monitor.descriptor.value (lib/decorators/preconditions/trackable.ts:46:9)
+      at Monitor._run (../../../node_modules/@oclif/core/lib/command.js:67:22)
+
+
+undefined
+"
+`;
+
+exports[`org:resources:monitor when the operation fails should output the same thing 2`] = `
+"
+Monitoring snapshot my-snapshot...
+Operation undefined..."
+`;
+
+exports[`org:resources:monitor when the operation succeed should output the same thing 1`] = `
+"  console.error
+     »   Warning: The org:resources commands are currently in public beta, please 
+     »   report any issue to github.com/coveo/cli/issues
+
+      43 |   @Preconditions(IsAuthenticated())
+      44 |   public async run() {
+    > 45 |     this.warn(
+         |          ^
+      46 |       'The org:resources commands are currently in public beta, please report any issue to github.com/coveo/cli/issues'
+      47 |     );
+      48 |     await this.printHeader();
+
+      at Object.warn (../../../node_modules/@oclif/core/lib/errors/index.js:61:13)
+      at Monitor.warn (../../../node_modules/@oclif/core/lib/command.js:86:20)
+      at Monitor.apply (commands/org/resources/monitor.ts:45:10)
+      at Monitor.descriptor.value (lib/decorators/preconditions/preconditions.ts:19:32)
+      at Monitor.trackCommand (lib/decorators/preconditions/trackable.ts:63:3)
+      at Monitor.descriptor.value (lib/decorators/preconditions/trackable.ts:46:9)
+      at Monitor._run (../../../node_modules/@oclif/core/lib/command.js:67:22)
+
+
+"
+`;
+
+exports[`org:resources:monitor when the operation succeed should output the same thing 2`] = `
+"
+Monitoring snapshot my-snapshot...
+Operation undefined..."
+`;
+
+exports[`org:resources:monitor when the operations conclude with missing vault entries should output the same thing 1`] = `
+"  console.error
+     »   Warning: The org:resources commands are currently in public beta, please 
+     »   report any issue to github.com/coveo/cli/issues
+
+      43 |   @Preconditions(IsAuthenticated())
+      44 |   public async run() {
+    > 45 |     this.warn(
+         |          ^
+      46 |       'The org:resources commands are currently in public beta, please report any issue to github.com/coveo/cli/issues'
+      47 |     );
+      48 |     await this.printHeader();
+
+      at Object.warn (../../../node_modules/@oclif/core/lib/errors/index.js:61:13)
+      at Monitor.warn (../../../node_modules/@oclif/core/lib/command.js:86:20)
+      at Monitor.apply (commands/org/resources/monitor.ts:45:10)
+      at Monitor.descriptor.value (lib/decorators/preconditions/preconditions.ts:19:32)
+      at Monitor.trackCommand (lib/decorators/preconditions/trackable.ts:63:3)
+      at Monitor.descriptor.value (lib/decorators/preconditions/trackable.ts:46:9)
+      at Monitor._run (../../../node_modules/@oclif/core/lib/command.js:67:22)
+
+
+"
+`;
+
+exports[`org:resources:monitor when the operations conclude with missing vault entries should output the same thing 2`] = `
+"
+Monitoring snapshot my-snapshot...
+Operation undefined..."
+`;
--- a/packages/cli/src/commands/org/resources/__snapshots__/monitor.spec.ts.snap
+++ b/packages/cli/src/commands/org/resources/__snapshots__/monitor.spec.ts.snap
@@ -2,8 +2,8 @@
 
 exports[`org:resources:monitor when the operation conclude with no changes should output the same thing 1`] = `
 "  console.error
-     »   Warning: The org:resources commands are currently in public beta, please 
-     »   report any issue to github.com/coveo/cli/issues
+     ›   Warning: The org:resources commands are currently in public beta, please 
+     ›   report any issue to github.com/coveo/cli/issues
 
       43 |   @Preconditions(IsAuthenticated())
       44 |   public async run() {
@@ -33,8 +33,8 @@ Operation undefined..."
 
 exports[`org:resources:monitor when the operation fails should output the same thing 1`] = `
 "  console.error
-     »   Warning: The org:resources commands are currently in public beta, please 
-     »   report any issue to github.com/coveo/cli/issues
+     ›   Warning: The org:resources commands are currently in public beta, please 
+     ›   report any issue to github.com/coveo/cli/issues
 
       43 |   @Preconditions(IsAuthenticated())
       44 |   public async run() {
@@ -65,8 +65,8 @@ Operation undefined..."
 
 exports[`org:resources:monitor when the operation succeed should output the same thing 1`] = `
 "  console.error
-     »   Warning: The org:resources commands are currently in public beta, please 
-     »   report any issue to github.com/coveo/cli/issues
+     ›   Warning: The org:resources commands are currently in public beta, please 
+     ›   report any issue to github.com/coveo/cli/issues
 
       43 |   @Preconditions(IsAuthenticated())
       44 |   public async run() {
@@ -96,8 +96,8 @@ Operation undefined..."
 
 exports[`org:resources:monitor when the operations conclude with missing vault entries should output the same thing 1`] = `
 "  console.error
-     »   Warning: The org:resources commands are currently in public beta, please 
-     »   report any issue to github.com/coveo/cli/issues
+     ›   Warning: The org:resources commands are currently in public beta, please 
+     ›   report any issue to github.com/coveo/cli/issues
 
       43 |   @Preconditions(IsAuthenticated())
       44 |   public async run() {
--- a/packages/cli/src/commands/org/resources/push.ts
+++ b/packages/cli/src/commands/org/resources/push.ts
@@ -123,8 +123,8 @@ export default class Push extends Command {
   }
 
   private getSuccessReportHandler(snapshot: Snapshot) {
-    const successReportWithChangesHandler =
-      this.successReportWithChangesHandler.bind(this, snapshot);
+    const successReportWithChangesHandler = () =>
+      this.successReportWithChangesHandler(snapshot);
     return async function (this: SnapshotReporter) {
       return successReportWithChangesHandler();
     };
--- a/packages/cli/src/commands/org/resources/push.ts
+++ b/packages/cli/src/commands/org/resources/push.ts
@@ -123,20 +123,24 @@ export default class Push extends Command {
   }
 
   private getSuccessReportHandler(snapshot: Snapshot) {
-    return ((pushCommand: Push) =>
-      async function (this: SnapshotReporter) {
-        if (!this.hasChangedResources()) {
-          return;
-        }
+    const successReportWithChangesHandler =
+      this.successReportWithChangesHandler.bind(this, snapshot);
+    return async function (this: SnapshotReporter) {
+      if (!this.hasChangedResources()) {
+        return;
+      } else {
+        return successReportWithChangesHandler();
+      }
+    };
+  }
 
-        const {flags} = await pushCommand.parse(Push);
-        const canBeApplied =
-          flags.skipPreview || (await pushCommand.askForConfirmation());
+  private async successReportWithChangesHandler(snapshot: Snapshot) {
+    const {flags} = await this.parse(Push);
+    const canBeApplied = flags.skipPreview || (await this.askForConfirmation());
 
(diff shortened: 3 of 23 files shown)


packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +58,39 @@ export class SnapshotReporter {
...
+      for (const errors of Object.values(resourceTypeEntry)) {
+        for (const err of errors) {
+          if (!SnapshotReporter.isVaultEntryMessage(err)) {
+            return false;
+          }
+        }
+      }
+    }
+    return true;
+  }
```
louis-bompart, line 84: Ain't a fan, but I think it's the fastest.
The fastest way to know if all errors are missing vault entries errors is to check them all and return false as soon as we find an error that ain't a missing vault entry.

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +58,39 @@ export class SnapshotReporter {
...
 
+  public getReportStatus(): SnapshotReportStatus {
+    if (this.isSuccessReport()) {
+      return SnapshotReportStatus.SUCCESS;
+    }
+    if (this.isVaultEntriesMissingReport()) {
+      return SnapshotReportStatus.MISSING_VAULT_ENTRIES;
+    }
+    return SnapshotReportStatus.ERROR;
+  }
```
louis-bompart, line 69: Allow simpler handling for the caller.
However, maybe we should allow the caller to register callbacks for 'success'/'missing'/'error' and compute from here (given the refacto tagged on preview and push).

To be considered.

.vscode/launch.json
```diff
@@ -4,6 +4,13 @@
   // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
   "version": "0.2.0",
   "configurations": [
+    {
+      "name": "Attach by Process ID",
+      "processId": "${command:PickProcess}",
+      "request": "attach",
+      "skipFiles": ["<node_internals>/**"],
+      "type": "node"
+    },
```
louis-bompart, line 13: - [ ] revert dat

packages/cli/src/lib/errors/snapshotErrors.ts
```diff
@@ -139,3 +139,24 @@ export class SnapshotGenericError
     trySavingDetailedReport(this);
   }
 }
+export class SnapshotMissingVaultEntriesError
+  extends PrintableError
+  implements DetailedReportable
+{
+  public name = 'Snapshot Error';
```
louis-bompart, line 146:
```suggestion
  public name = 'Snapshot Missing Vault Entries';
```
```suggestion
  public name = 'Snapshot Error';
```

louis-bompart, line 146:
```suggestion
  public name = 'Snapshot Missing Vault Entries';
```
```suggestion
  public name = 'Snapshot Error';
```

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +58,60 @@ export class SnapshotReporter {
...
+
+  private static isVaultEntryMessage(err: string): boolean {
+    // TODO CDX-939: Define contract with backend for report and upcoming contract.
+    // Current 'contract' 😅:
+    return /^The vault entry referenced by.*could not be found in the vault\.$/.test(
+      err
+    );
+  }
+
```
y-lakhdar, line 114: Why do you need to validate that via a regex?
There is already a call for that
https://github.com/coveo/platform-client/blob/master/src/resources/Vaults/Vaults.ts#L16

packages/cli/src/lib/snapshot/reportPreviewer/reportPreviewerDataModels.ts
```diff
@@ -7,3 +7,9 @@ export interface ReportViewerResourceReportModel {
   name: string;
   operations: ResourceSnapshotsReportOperationModel;
 }
+
+export enum SnapshotReportStatus {
```
y-lakhdar, line 11: Could we have that in the Platform-client?

louis-bompart: Currently, it would not make sense because the REST API does not return anything that relates to that. When the contract for the new reports will be defined, yes prolly.
Although, we would certainly have some kind of extension (I'm adding a new ReportStatus for the "no changes")

packages/cli/src/lib/snapshot/snapshotReporter.ts
louis-bompart: We discussed it on meet, so I'll summarize:
We saw that on the [Migration Service]( https://github.com/coveo/migrationservice/blob/1d3ac8207282942e1828f942d3858ea1f36baa81/coveo-cloud-services-migration/src/main/java/com/coveo/cloud/migration/service/interpolation/SnapshotPlaceholderInterpolator.java#L95-L128)
The errors are aggregated.

That means that if there's more than one error on a resource we'll get them all in one go, allowing us to offer a comprehensive report to the user.

This will avoid the number of 'attempts' a user would do before successfully running its push (assuming he digests the report properly, but if he doesn't that'll be on our end to fix that).

I remember my first days with the Coveo For Salesforce package, when the MDAPI was the only way to push (and therefore test) stuff on an SDO, it was a hell circus circle: Launch a deploy, wait 10', get 1 error, fix it, repeat. :feelsgood: 

I would really not want to impose that on our users.

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -7,11 +7,43 @@ import {
...
+  SuccessfulReportHandler;
+
+type FixableErrorHandlers = {
+  [SnapshotReportStatus.MISSING_VAULT_ENTRIES]: /**
+   * TODO CDX-936: Define return type of the cb, & remove the SnapshotReporterHandler.
+   * Return type should be either a boolean if we do a 'all or nothing'
+   * handling (i.e. either we manage to do them all, or we do nothing)
+   * Return type should be a `Set<string>` if we do a 'best effort' strategy
+   * (i.e. if some entries are messed up, we ignore them and push the others)
+   */
```
louis-bompart, line 29: Too lazy to change the comment right now, will do tmrw:
 - [ ] Change comment @louis-bompart 
 
 But, @y-lakhdar , see `reportPreviewer` for an example of implementation (I think it makes more sense for the snapshot reporter to just be the orchestrator, and to leave the UI interactions to the reportPreviewer.

packages/cli/src/lib/snapshot/reportPreviewer/reportPreviewer.ts
```diff
@@ -102,8 +157,6 @@ export class ReportViewer {
   }
 
   private logResourceErrors(ResourceSnapshotType: string) {
```
louis-bompart, line 159: `logResourceErrors` (as of de55db1) is only refactors, fret not.

packages/cli/src/lib/snapshot/reportPreviewer/reportPreviewer.ts
```diff
@@ -32,19 +41,65 @@ export class ReportViewer {
...
+        yellow(
+          `Missing vault ${pluralizeIfNeeded(
+            ReportViewer.entryPlurable,
+            entries.length
+          )}:`
+        )
+      );
+      missingVaultEntryPrinter(entries);
+    };
+  }
```
louis-bompart, line 83: TL;DR: Leverage the 'printAbridgedMessage' extracted from the error printer, and feed him the missing entries crunched by the reporter.

packages/cli/src/lib/snapshot/reportPreviewer/reportPreviewer.ts
```diff
@@ -1,16 +1,25 @@
...
 
+type Plurable = [singular: string, plural: string];
+function pluralizeIfNeeded(plurable: Plurable, unprintedMessages: number) {
+  return plurable[unprintedMessages > 1 ? 1 : 0];
+}
+
 export class ReportViewer {
+  private static errorPlurable: Plurable = ['error', 'errors'];
+  private static entryPlurable: Plurable = ['entry', 'entries'];
+
```
louis-bompart, line 22: Chmancy Code to handle the plural, can't dry it mo'.

packages/cli/src/lib/snapshot/reportPreviewer/reportPreviewer.spec.ts
```diff
@@ -23,34 +23,11 @@ describe('ReportViewer', () => {
...
-          Fields
-            • RESOURCE_ALREADY_EXISTS: Field foo already exists.
-            • RESOURCE_ALREADY_EXISTS: Field bar already exists.
-            • RESOURCE_ALREADY_EXISTS: Field dsads already exists.
-            • RESOURCE_ALREADY_EXISTS: Field fdww already exists.
-            • RESOURCE_ALREADY_EXISTS: Field csad already exists.
-            (2 more errors)`
-        );
+      .it('should print the same report', (ctx) => {
+        expect(ctx.stdout).toMatchSnapshot();
```
louis-bompart, line 30: I don't remember if we knew about snapshots back then, I reckon not.
This is the perfect use case for them, so here they go.

packages/cli/src/lib/errors/snapshotErrors.ts
```diff
@@ -139,3 +139,20 @@ export class SnapshotGenericError
     trySavingDetailedReport(this);
   }
 }
+export class SnapshotMissingVaultEntriesError
```
louis-bompart, line 142: - [ ] UT

packages/cli/src/commands/org/resources/push.ts
```diff
@@ -86,8 +89,20 @@ export default class Push extends Command {
...
+      )
+      .setReportHandler(
+        SnapshotReportStatus.MISSING_VAULT_ENTRIES,
+        getMissingVaultEntriesReportHandler(snapshot, cfg, this.projectPath)
+      )
+      .setReportHandler(
+        SnapshotReportStatus.ERROR,
+        getErrorReportHandler(snapshot, cfg, this.projectPath)
+      )
+      .handleReport();
```
louis-bompart, line 105: - [ ] UT the handler

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +89,110 @@ export class SnapshotReporter {
...
+
+  public setReportHandler<T extends SnapshotReportStatus>(
+    status: T,
+    handler: SnapshotReporterHandlers[T]
+  ): SnapshotReporter {
+    this.reportHandlers[status] = handler;
+    return this;
+  }
+
+  public async handleReport(): Promise<void> {
```
louis-bompart, line 107: - [ ] UT. Use Istanbul to check and double-check all paths are covered

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +89,110 @@ export class SnapshotReporter {
...
+    const reportStatuses = this.getReportStatuses();
+    for (const handler of reportStatuses.fixables) {
+      await this.reportHandlers[handler].apply(this);
+      this.reportHandlers[handler] = () => {};
```
louis-bompart, line 111: - [ ] UT. Ensure all handlers are consumed and reset to noop.

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +89,110 @@ export class SnapshotReporter {
...
+    }
+    if (reportStatuses.errors.length > 0) {
+      for (const handler of reportStatuses.errors) {
+        await this.reportHandlers[handler].apply(this);
+        this.reportHandlers[handler] = () => {};
+      }
+    } else if (reportStatuses.successes.length > 0) {
```
louis-bompart, line 118: - [ ] UT ensures success is **not** called when there are errors.
- [ ] UT ensures success is called when there are no errors.

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +89,110 @@ export class SnapshotReporter {
...
+  public setReportHandler<T extends SnapshotReportStatus>(
+    status: T,
+    handler: SnapshotReporterHandlers[T]
+  ): SnapshotReporter {
+    this.reportHandlers[status] = handler;
+    return this;
+  }
+
+  public async handleReport(): Promise<void> {
+    const reportStatuses = this.getReportStatuses();
```
louis-bompart, line 108: - [ ] UT/?: what if no fixables, no errors, no successes?

packages/cli/src/lib/snapshot/reportPreviewer/reportPreviewer.ts
louis-bompart: well, 494180c changed that:
We were parsing twice the report, once to get the error, once to get the missing vault entries.
So instead, during the parsing/creation of the snapshotReporter, we create the data structure for the reportPreviewer.

Note that the `snapshot` was taken prior to this big change. So, the fact that they work means that the refactoring didn't break everything (famouslastwords, ofc, the E2E hasn't been thru... yet)

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -98,6 +101,31 @@ describe('Snapshot', () => {
     );
   };
 
+  const doMockSnapshotReporter = (
+    snapshotReportStatus: SnapshotReportStatus
+  ) => {
+    let handlerToCall: (
+      this: SnapshotReporter
+    ) => void | Promise<void> = () => {};
+    mockedSnapshotReporter.prototype.setReportHandler.mockImplementation(
```
louis-bompart, line 110: It's better to mock method individually, otherwise, strangely, `jest.mock.instances[0]` will not be `jest.mock.returns[0].value`, which mean that if you try to check if X is called with an instance of a mock, well, you're outta luck

packages/cli/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -131,7 +159,7 @@ describe('Snapshot', () => {
   });
 
   afterEach(() => {
-    jest.resetAllMocks();
+    jest.clearAllMocks();
```
louis-bompart, line 162: - [ ] revert

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -3,15 +3,50 @@ import {
...
+  ((this: SnapshotReporter) => boolean) | SnapshotReporterHandler | NoopHandler;
+};
+
+type UnfixableErrorHandlers = {
+  [K in SnapshotReportStatus]: SnapshotReporterHandler | NoopHandler;
+};
+
+type SuccessfulReportHandler = {
+  [K in SnapshotReportStatus]: SnapshotReporterHandler | NoopHandler;
+};
```
louis-bompart, line 40: May look daunting, but it helps when you call `setReportHandler`: You'll know what your callback can use.

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +92,112 @@ export class SnapshotReporter {
...
+
+    if (!this.hasChangedResources()) {
+      statuses.successes.push(SnapshotReportStatus.NO_CHANGES);
+    }
+
+    if (this.isSuccessReport()) {
+      statuses.successes.push(SnapshotReportStatus.SUCCESS);
+    }
+
+    if (this.resourceInErrorCount > 0) {
```
louis-bompart, line 156: Note, may need to change in the future if we have another kind of error. (sofar so good?)

packages/cli/src/lib/snapshot/reportPreviewer/reportPreviewer.ts
louis-bompart: (note, sha has changed, 'cause rebase)

packages/cli/src/lib/decorators/preconditions/apiKeyPrivilege.ts
```diff
@@ -64,7 +64,7 @@ async function getConfiguration() {
 }
 
 function hasGetFlagMethod(
-  candidate: any
+  candidate: Command & {getFlags?: () => Promise<unknown>}
```
louis-bompart, line 67:
```suggestion
  candidate: any
```

y-lakhdar (approved):

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +90,115 @@ export class SnapshotReporter {
...
+    status: T,
+    handler: SnapshotReporterHandlers[T]
+  ): SnapshotReporter {
+    this.reportHandlers[status] = handler;
+    return this;
+  }
+
+  public async handleReport(): Promise<void> {
+    const reportStatuses = this.getReportStatuses();
+    for (const handler of reportStatuses.fixables) {
```
olamothe, line 110: nit: Could remove some repetition and improve readability

```typescript
public async handleReport(): Promise<void> {
  const reportStatuses = this.getReportStatuses();
   // or extract a type for what's returned by `getReportStatuses` top level in this file
   type reportStatusProp = keyof typeof reportStatuses;
  
  const executeHandlers = async (k: reportStatusProp) => {
     for (const handler of reportStatuses[k]) {
          await this.callAndReset(handler);
        }
  }

  const has = (k: reportStatusProp) => {
       return reportStatuses[k].length > 0;
  }
  
  await executeHandlers('fixables');
  if (has('errors')) {
      await executeHandlers('errors');      
      return;
  }

  if (has('successes')) {
      await executeHandlers('successes');      
  }
}
```

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +90,115 @@ export class SnapshotReporter {
...
+    }
+
+    return statuses;
+  }
+
+  private computeMissingVaultEntries(): void {
+    for (const [resourceType, resource] of Object.entries(
+      this.report.resourceOperationResults
+    )) {
+      for (const errors of Object.values(resource)) {
```
olamothe, line 165: There might be a possibility to first extract all errors with a `flatMap`, and then iterate on that.

Would remove the multiple nesting level and improve readability.

olamothe (approved):

packages/cli/src/commands/org/resources/__snapshots__/monitor.spec.ts.snap
```diff
@@ -0,0 +1,126 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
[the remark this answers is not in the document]
louis-bompart: https://jestjs.io/docs/snapshot-testing

packages/cli/src/lib/errors/snapshotErrors.ts
```diff
@@ -139,3 +139,20 @@ export class SnapshotGenericError
...
+{
+  public name = 'Snapshot Missing Vault Entries';
+  public constructor(
+    public snapshot: Snapshot,
+    public cfg: Configuration,
+    public projectPath?: string
+  ) {
+    super(SeverityLevel.Error);
+    this.message = dedent`Your snapshot is missing some vault entries.
+    Ensure that all entries are present on ${snapshot.targetId} and try again.`;
```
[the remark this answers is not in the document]
louis-bompart: It will be thrown if the user rejects all forms of resolution

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -57,6 +90,115 @@ export class SnapshotReporter {
     );
   }
 
+  private reportHandlers: SnapshotReporterHandlers = {
+    [SnapshotReportStatus.SUCCESS]: () => {},
+    [SnapshotReportStatus.NO_CHANGES]: () => {},
+    [SnapshotReportStatus.MISSING_VAULT_ENTRIES]: () => {},
+    [SnapshotReportStatus.ERROR]: () => {},
```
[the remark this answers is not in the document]
louis-bompart: Error for now? We can create special cases later in but this was not present currently and is a whole other feature.

packages/cli/src/lib/errors/snapshotErrors.ts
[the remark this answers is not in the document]
y-lakhdar: then what happens once the vault entries are created?
* terminate the command and the user to repush his snapshot ? 
* go through the push operation a second time

packages/cli/src/lib/snapshot/snapshotReporter.ts
louis-bompart: I thought about it, but I would not do this for two reasons:
1. We currently use two data from the whole object: the resourceType & the message of the error. One is at the very first depth level, the other at the deepest one. Just extracting the error with a `flatMap` would not be enough, and adding the resources to the 'array' generated by the `flatMap` would create more complexity than it'd resolve.
2. If we go in two steps, it means we also go through the array twice. If possible, this is something I'd like to avoid to keep the processing time as low as possible.

packages/cli/src/lib/errors/snapshotErrors.ts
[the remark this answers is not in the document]
louis-bompart: If the vault entries are created outside of the command (i.e. the user rejected all forms of resolutions), then yes, he'd go through the command again.
If the vault entries are created through the resolution step, the missingVaultEntryHandler should not reach the throw statement.
You'll note that the throw statement is after two TODOs, one for CDX-935 and CDX-936. It's the job of those to ensure that when then handling has been done properly to not reach the throw statement.

So, what does that mean:
 If the report has other issues than the secrets, it should fail when the error is handled.
 If the report has no other issues, the success handler should then be called. For the push, that means applying the snapshot.

packages/cli/src/lib/snapshot/snapshotReporter.ts
louis-bompart: To alleviates the issue tho, I extracted the scope of the deepest for in another func

packages/cli/src/lib/snapshot/snapshotReporter.ts
louis-bompart: I took that and tweaked it a bit.
I'm not found of using string as keys, it's a pain to refactor, likewise, I think has is a bit overkill. but the 'spirit' of the comment has been followed.

packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -3,15 +3,48 @@ import {
...
+    | NoopHandler;
+};
+
+type UnfixableErrorHandlers = {
+  [K in SnapshotReportStatus]: SnapshotReporterHandler | NoopHandler;
+};
+
+type SuccessfulReportHandler = {
+  [K in SnapshotReportStatus]: SnapshotReporterHandler | NoopHandler;
+};
```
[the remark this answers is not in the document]
louis-bompart: refactored, for now the callback are all getting the same stuff.

PR #777, opened 2022-05-02
feat(snapshot): add manifest file on pull

louis-bompart:
## Proposed changes

When pulling, it can be interesting to know where we just pulled from.
To do this, this PR adds a manifest file meant to contain data relative to the resources.

## Testing

- [x] Unit Tests: yep
- [ ] Functional Tests: existing cover the most important cases, maybe the e2e with snapshot vault will include that, but otherwise, as it stands, too much overlap with the ut to be worth it.


packages/cli/src/lib/project/project.ts
```diff
@@ -88,12 +93,24 @@ export class Project {
         archive.directory(this.resourcePath, false);
         archive.finalize();
       });
+      if (cachedManifest) {
```
olamothe, line 96: Could this block and lines 101-103 be moved to a `finally` block to remove repetition ?

packages/cli/src/lib/project/project.ts
```diff
@@ -88,12 +93,24 @@ export class Project {
...
+        writeJsonSync(this.resourceManifestPath, cachedManifest);
+      }
       CliUx.ux.error(error as string | Error);
     }
   }
 
+  public writeResourcesManifest(orgId: string) {
+    const manifestJson =
+      readJsonSync(this.resourceManifestPath, {throws: false}) ?? {};
+    writeJsonSync(this.resourceManifestPath, {...manifestJson, orgId});
```
olamothe, line 111: Could this line throw ? 
If so, shouldn't it be handled ?

olamothe (approved):

packages/cli/src/lib/project/project.ts
louis-bompart: That was my first attempt, but because there's a `return` statement 3 lines down below, the finally clause doesn't work.
If you have another suggestion, be my guest tho, I ain't liking the repetition either.

y-lakhdar (approved):

packages/cli/src/lib/project/project.ts
louis-bompart: It's quite unlikely, but yeah, let's handle it by noop its error (i.e. the manifest, it's nice, but its not the most important)

packages/cli/src/lib/project/project.ts
```diff
@@ -106,9 +106,13 @@ export class Project {
   }
 
   public writeResourcesManifest(orgId: string) {
-    const manifestJson =
-      readJsonSync(this.resourceManifestPath, {throws: false}) ?? {};
-    writeJsonSync(this.resourceManifestPath, {...manifestJson, orgId});
+    try {
+      const manifestJson =
+        readJsonSync(this.resourceManifestPath, {throws: false}) ?? {};
+      writeJsonSync(this.resourceManifestPath, {...manifestJson, orgId});
```
y-lakhdar, line 112: When would this fail?

louis-bompart: It could fail if for some odd reason the file is read-only or if the file does not exist and the directory is read-only.
TL;DR: odd chmod could cause EACCESS issue, theoretically

PR #778, opened 2022-05-02
feat: allow user to fill in their missing vault entries

y-lakhdar:
## Proposed changes

Open an editor so the user can enter the required vault entries.

## Prerequisites
- [x] Merge PR on platform client (https://github.com/coveo/platform-client/pull/451)

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/commands/org/resources/preview.spec.ts
```diff
@@ -402,8 +402,9 @@ describe('org:resources:preview', () => {
       test
         .stdout()
         .stderr()
+        .stub(CliUx.ux, 'confirm', () => async () => false)
```
louis-bompart, line 405: https://github.com/oclif/fancy-test#stdin-mocking ?

packages/cli/src/commands/org/resources/preview.ts
```diff
@@ -39,6 +39,7 @@ export default class Preview extends Command {
   public static flags = {
     ...wait(),
     ...sync(),
+    // TODO: CDX-935 add flag to automatically transfer vault entries when transferable
```
louis-bompart, line 42: For push, sure, it'll be the `-y` flag.
Outta rethink abt it for preview.
I'm not sure it's really a good idea to have flags to skip stuff for the preview yet.

Too much options is as bad as not enough so let's find the good middle

packages/cli/src/commands/org/resources/push.ts
```diff
@@ -41,6 +41,7 @@ export default class Push extends Command {
   public static flags = {
     ...wait(),
     ...sync(),
+    // TODO: CDX-935 add flag to automatically transfer vault entries when transferable
```
louis-bompart, line 44: `-y` should do it, let's not add a myriad of flags.

Adding a flag, even just a boolean, add quite a lot of  combination of flags, and thus corner case.

We should be mindful of that when adding new flags.

packages/cli/src/lib/snapshot/snapshotCommon.ts
```diff
@@ -87,9 +88,30 @@ export function getMissingVaultEntriesReportHandler(
...
+    // **** Pseudo-code END ****
+
+    const shouldCreate = await CliUx.ux.confirm(
+      `\nWould you like to create the missing vault entries in the destination organization (${snapshot.targetId})? (y/n)`
+    );
+    if (shouldCreate) {
+      const vault = new Vault(snapshot.targetId);
+      await vault.createEntries(Array.from(this.missingVaultEntries));
+      return;
+    }
```
louis-bompart, line 113: - [ ] Refactor in a function.

we could do something fancy
```ts
const missingVaultEntriesSubHandlers = [transferVaultEntries, manuallyCreateVaultEntries, throwError]
for(const subhandler of missingVaultEntriesSubHandlers) {
  if (await subHandler.apply(this, snapshot, cfg, projectPath)) {
    break
  }
}

const throwError = function (this: void, snapshot: Snapshot, cfg:Configuration, projectPath: string) {
 throw new SnapshotMissingVaultEntriesError(snapshot, cfg, projectPath);
}
```

packages/cli/src/lib/snapshot/vault.ts
```diff
@@ -0,0 +1,154 @@
...
+import {Snapshot} from './snapshot';
+import {VaultEntryAttributes} from './snapshotReporter';
+
+export class Vault {
+  private static defaultEntryValue = '';
+  public constructor(private organizationId: string) {}
+
+  public async createEntries(entries: VaultEntryAttributes[]) {
+    const client = await this.client();
+    const vaultEntryFilePath = join(cwd(), `${this.organizationId}-vault.json`);
```
louis-bompart, line 24: cwd or project root?

packages/cli/src/lib/snapshot/vault.ts
```diff
@@ -0,0 +1,154 @@
...
+    while (!valid) {
+      await this.openFile(vaultEntryFilePath);
+      const key = await CliUx.ux.prompt('', {
+        type: 'single',
+        required: false,
+        prompt: 'Press any key once finished. Press q to abort\n',
+      });
+      if (key === 'q') {
+        return null;
+      }
```
louis-bompart, line 38: Using `fs.watch`, you should be able to see/know when the file is saved and thus should be able to run `this.ensureEntriesValidity` etc without having to ask the user to press a key

packages/cli/src/lib/snapshot/vault.ts
```diff
@@ -0,0 +1,154 @@
...
+      entries
+    );
+    rmSync(vaultEntryFilePath);
+
+    if (vaultEntryModels.length === 0) {
+      return;
+    }
+
+    return Promise.all(
+      vaultEntryModels.map((entry) => client.vault.create(entry))
```
louis-bompart, line 57: - [ ] Todo, future check with the team if they prefer to have a batch call.

packages/cli/src/lib/snapshot/vault.ts
```diff
@@ -0,0 +1,154 @@
...
+import open from 'open';
+import {join} from 'path';
+import {cwd} from 'process';
+import dedent from 'ts-dedent';
+import {InvalidVaultEntry} from '../errors/vaultErrors';
+import {AuthenticatedClient} from '../platform/authenticatedClient';
+import {Snapshot} from './snapshot';
+import {VaultEntryAttributes} from './snapshotReporter';
+
+export class Vault {
```
louis-bompart, line 18: VaultHandler

packages/cli/src/lib/snapshot/vault.ts
```diff
@@ -0,0 +1,154 @@
...
+  private static defaultEntryValue = '';
```
louis-bompart, line 19:
```suggestion
  private static readonly defaultEntryValue = '';
```

packages/cli/src/commands/org/resources/preview.ts
y-lakhdar: We could have one flag (e.g. `--yes`) to automatically approve all questions (similar to [Pulumi](https://www.pulumi.com/docs/reference/cli/pulumi_up/#options)).

packages/cli/src/lib/snapshot/vault.ts
y-lakhdar: cwd will always be equal to project root since the push and preview commands should be executed in the project root.

packages/cli/src/lib/utils/string.ts
```diff
@@ -1,3 +1,11 @@
+export type Plurable = [singular: string, plural: string];
+export function pluralizeIfNeeded(
+  plurable: Plurable,
+  unprintedMessages: number
+) {
+  return plurable[unprintedMessages > 1 ? 1 : 0];
+}
+
```
y-lakhdar, line 8: This is useful, let's make it reusable

packages/cli/src/lib/snapshot/vault.ts
y-lakhdar: mmh... I think that would be bad UX.
The user might not understand what is happening after seeing the terminal logging stuff.
Specially if the terminal and the editor are 2 separate applications. If that is the case, the user's focus will no longer be in the terminal and could miss the warning popping up. 

Anyway, we should wait for the user consent before running the next operation. It's more verbose but less confusing.

louis-bompart: You raise a good point about the application that would be in focus.

I think that focus hands-off should be done without additional user interactions (i.e. the user is not in charge of saying "k now the cli take back the focus", because it's just another 'useless' prompt (i.e. the user does not provide valuable info for the ongoing operation).

I think we should follow the example of Git on that regard:

image

To ensures that the user is seeing the instructions, they put it in the doc that they open.
We should do the same, so, our file would be like this:
```json5
// Fill in your secrets, then save and close the file.
{
    "mysecret1": null,
    //...
}
```

Note that does rely on the JSON5 format `npm i JSON5`

packages/cli/src/commands/org/resources/preview.ts
louis-bompart: I'm still unsure for the preview due to its nature that is fundamentally oriented towards a human, and thus where interactions are not bad if well spaced.
The `--yes` flag is really there for our cold metal overlord
image

packages/cli/src/lib/snapshot/vault.ts
louis-bompart: (Or if dotenv, just use do as git, as line starting with `#` are considered comments)

packages/cli/src/lib/snapshot/vaultHandler.ts
```diff
@@ -0,0 +1,176 @@
...
+import {Plurable, pluralizeIfNeeded} from '../utils/string';
+import {ProcessAbort} from '../../lib/errors/processError';
+import dedent from 'ts-dedent';
+
+export class VaultHandler {
+  private static readonly defaultEntryValue = '';
+
+  public constructor(private organizationId: string) {}
+
+  public async createEntries(entries: VaultEntryAttributes[]) {
```
olamothe, line 29: It's a beefy method that (obviously) execute a lot of stuff... 

Can we try to split it with private functions just for readability sake and see how it looks like ?

pseudo: 
```
public async createEntries(entries: VaultEntryAttributes[]) {
   this.prepareAndOpenFile();
   this.ensureValidityOrAbort();
   this.doCreateVaultEntries();
}

olamothe (approved):

louis-bompart (approved):

PR #779, opened 2022-05-03
fix(snapshot): reporter considers fixable reports for onSuccess callback

louis-bompart:
https://coveord.atlassian.net/browse/CDX-946

## Proposed changes

- Removed `isSuccessReport` method so that all usage fit the pattern of the snapshotReporter.
- Shift the last usage of the defunct method to the new pattern
- Fixed UT by removing unforeseen side-effect: because we were only mocking the return/resolved value of some function, they were sharing instances across tests. Mocking the implementation fixed that, by making sure the returned value of the mocked implementation is always 'fresh new'

## Testing

- [X] Unit Tests: refactored; I added a TODO for post GA to rewrite some big chunk of tests in preview and push, that's needed. I spent too much time on this PR because of the debt we accrued here.
- [x] Functionnal Tests: existing


packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -132,12 +122,10 @@ export class SnapshotReporter {
...
     }
 
-    if (this.isSuccessReport()) {
-      statuses.successes.push(SnapshotReportStatus.SUCCESS);
-    }
-
     if (this.resourceInErrorCount > 0) {
       statuses.errors.push(SnapshotReportStatus.ERROR);
+    } else {
+      statuses.successes.push(SnapshotReportStatus.SUCCESS);
```
louis-bompart, line 128: This is bound to really evolve as we improve the error handling with better warnings, auto-fixes etc.

y-lakhdar (approved):
Nice.
discussed with Louis in Slack

y-lakhdar (changes requested):
the e-fuse does not work

olamothe (approved):

PR #780, opened 2022-05-03
fix(precond): use new flag name

louis-bompart:
https://coveord.atlassian.net/browse/CDX-950

## Proposed changes

#773 renamed target to the organization, this was not reflected here.
Still checking for why this was not flagged up on CI, another PR will be open to addressing the false ✔️.

## Testing

- [x] Unit Tests: adapted

- [x] Functional Tests: failing on master currently, this pr fix that.


y-lakhdar (approved):

y-lakhdar:
good catch

PR #783, opened 2022-05-04
feat: introduce the files flag

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-952

## Proposed changes

Only use one flag for both JSON files and directories. 
This is not a breaking change since I kept the support for legacy flags even tho they are hidden.

- [ ] Merge before #749

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/cli/src/commands/source/push/add.ts
+++ b/packages/cli/src/commands/source/push/add.ts
@@ -23,19 +23,25 @@ export default class SourcePushAdd extends Command {
 
   public static flags = {
     file: Flags.string({
+      // For retro compatibility
       multiple: true,
-      exclusive: ['folder'],
-      char: 'f',
-      helpValue: 'myfile.json',
-      description: 'One or multiple file to push. Can be repeated.',
+      exclusive: ['folder', 'files'],
+      hidden: true,
     }),
     folder: Flags.string({
+      // For retro compatibility
       multiple: true,
-      exclusive: ['file'],
+      exclusive: ['file', 'files'],
       char: 'd',
-      helpValue: './my_folder_with_multiple_json_files',
+      hidden: true,
+    }),
+    files: Flags.string({
+      multiple: true,
+      exclusive: ['folder', 'file'],
+      char: 'f',
+      helpValue: 'myfile.json',
       description:
-        'One or multiple folder containing json files. Can be repeated',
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/source/push/add.ts
```diff
@@ -23,19 +21,22 @@ export default class SourcePushAdd extends Command {
...
+      hidden: true,
+    }),
+    files: Flags.string({
+      multiple: true,
+      char: 'f',
+      helpValue: 'myfile.json',
       description:
-        'One or multiple folder containing json files. Can be repeated',
+        'Combinaison of JSON files and folders (containing JSON files) to push. Can be repeated.',
     }),
```
louis-bompart, line 40: 1. Ensure that `file` and `folder` are aliases to `files` 
2. Create an Epic 'CLI-v2.0' and a User Story for this specific deprecation (i.e. removing the file and folder flags
3. Add a //TODO with the Identifier of the User Story

louis-bompart (commented):
The multiple aspect of the flags made me think, we might want to do things differently.

Instead of specifying the source as a args, and the files as flags, we could do the opposite.

So `coveo source:push:new mySource -f myFirstFile -f mySecondFile -f myThirdFile` would become
`coveo source:push:new -s mySource myFirstFile mySecondFile myThirdFile`

packages/cli/src/commands/source/push/add.ts
y-lakhdar: > Instead of specifying the source as a args, and the files as flags, we could do the opposite.

You don't need to specify the flag multiple time. It is
`coveo source:push:new mySource -f myFirstFile mySecondFile myThirdFile`

1. Can we create aliases for flags with oclif?

louis-bompart: No, but what you do is that you ensure all flags are mutually exclusive, and then you treat 'em all the same (i.e. dir now support file and file support dir, but that's fine it's not breaking :P )

y-lakhdar: At the end, it really doesn't matter what flag you use because the SDK checks whether the entry is a file or directory.
So in CDX-856, we can remove all flags.

PR #786, opened 2022-05-17
feat: handle invalid field names

y-lakhdar:
## Proposed changes

Provide detailed errors and next steps whenever invalid fields are detected by the SDK.

There are essentially 2 possible outcome:

### Outcome 1: invalid fields **can** be normalized
The user can rerun the the same command along with a flag to automatically fix fixable fields.
image

### Outcome 2: invalid fields **cannot** be normalized
The user has to fix his data before running the command
image

## Before merging
- [x] Merge https://github.com/coveo/push-api-client.js/pull/139
- [x] Bump push-api-client version
- [x] Unit Tests:
- [x] Manual Tests:


y-lakhdar:
Not sure about the verbiage. Is that clear enough?

packages/cli/src/lib/push/addCommon.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const allowedCharRegExp = new RegExp('^[a-z]+[a-z0-9_]*$');
+
+export const handleAddError = (err: unknown) => {
+  const {UnsupportedFieldError} = errors;
+  if (err instanceof UnsupportedFieldError) {
+    CliUx.ux.error('Invalid field name detected while parsing your data.', {
+      exit: false,
+    });
+    const normalizations = err.unsupportedFields.map((f) => {
```
louis-bompart, line 16: avoid using single letter for your callback.
Yes, it might be less verbose, but is also less clear at first glance.

packages/cli/src/lib/push/addCommon.ts
```diff
@@ -0,0 +1,85 @@
...
+      return {
+        original,
+        normalized,
+        valid: isFieldNameValid(normalized),
+      };
+    });
+
+    const fixable = normalizations
+      .filter(({valid}) => valid)
+      .map((f) => {
```
louis-bompart, line 28: You can use a flatMap to avoid iterating twice
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map

packages/cli/src/lib/push/addCommon.ts
```diff
@@ -0,0 +1,85 @@
...
+  const count = fields.length;
+  const pluralized = pluralizeIfNeeded(['field', 'fields'], count);
+  if (count === 0) {
+    return;
+  }
+
+  const fixable = fields[0].valid;
+  logNewLine();
+  CliUx.ux.log(
+    ` ${count} ${pluralized} detected in your data that can${
```
louis-bompart, line 67:
```suggestion
    ` ${count} ${pluralized} detected in your data can${
```

packages/cli/src/lib/push/addCommon.ts
```diff
@@ -0,0 +1,85 @@
...
+      fixable ? '' : 'not'
+    } be normalized`
+  );
+  CliUx.ux.table(fields, {
+    original: {header: 'Original'},
+    normalized: {header: 'Normalized'},
+  });
+};
+
+const logNewLine = (lines = 1) => {
```
louis-bompart, line 77: Unsure, this v.s. `Cli.ux.log(new Array(lines).fill('\n').join(''))` ?

y-lakhdar: turns out doing that creates the double amount of new lines

packages/cli/src/commands/source/push/add.ts
```diff
@@ -42,6 +49,7 @@ export default class SourcePushAdd extends Command {
     ...withFiles(),
     ...withMaxConcurrent(),
     ...withCreateMissingFields(),
+    ...normalizeInvalidFields(),
```
louis-bompart, line 52: respect the standard, `withXYZ`

packages/cli/src/commands/source/push/add.ts
```diff
@@ -65,38 +73,51 @@ export default class SourcePushAdd extends Command {
...
   public async catch(err?: Error & {exitCode?: number}) {
+    formatErrorMessage(err);
+    CliUx.ux.action.stop(red.bold('!'));
     throw err;
   }
 
+  public async finally() {
+    const {args} = await this.parse(SourcePushAdd);
+    const source = await this.getSource();
+    await source.setSourceStatus(args.sourceId, 'IDLE');
```
louis-bompart, line 109: You should run the finally if and only if we did successfully affected the status of the source IMO.

packages/cli/src/lib/push/addCommon.ts
```diff
@@ -0,0 +1,94 @@
+import {CliUx} from '@oclif/core';
+import {red, dim, green} from 'chalk';
+import {BuiltInTransformers, errors} from '@coveo/push-api-client';
+import {normalizeInvalidFields} from '../flags/sourceCommonFlags';
+import {pluralizeIfNeeded} from '../utils/string';
+import dedent from 'ts-dedent';
+
+const allowedCharRegExp = new RegExp('^[a-z]+[a-z0-9_]*$');
```
louis-bompart, line 8: Can a fieldname ends with a `_` ?

packages/cli/src/lib/push/addCommon.ts
```diff
@@ -0,0 +1,94 @@
...
+  normalized: string;
+  valid: boolean;
+}
+
+export const formatErrorMessage = (err: unknown) => {
+  const {UnsupportedFieldError} = errors;
+  if (err instanceof UnsupportedFieldError) {
+    const normalizations = getNormalizations(err.unsupportedFields);
+    const fixables = normalizations.flatMap(colorizeFixableField);
+    const unfixables = normalizations.flatMap(colorizeUnFixableField);
```
louis-bompart, line 20: You end up going through the array twice, you're probably due to do a method that follows this signature:

```ts
const {fixables, unfixables} = colorizeFields(normalizations);
```

packages/cli/src/lib/push/addCommon.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (!isFixable) {
+    message = dedent`Cannot normalize some of the invalid field names detected in your data
+      Review your data and ensure all field names as shown in the second table have at least one alphabetic character.`;
+  }
+  return message;
+};
+
+const colorizeFixableField = (
+  field: PrintableUnsupportedField
+): PrintableUnsupportedField | [] => {
```
louis-bompart, line 41: I think you're not using flatmap properly, you're supposed to return an array at each steps iirc.
> Return a 1-element array to keep the item, a multiple-element array to add items, or a 0-element array to remove the item.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map

louis-bompart (commented):
lgtm overall

packages/cli/src/commands/source/push/add.ts
y-lakhdar: I understand, but this is a bit overkill I think.
If the first call of `setSourceStatus` returned a 2** status code, it is safe to assume that the second call will work as well.
If the first `setSourceStatus` call fails for some reason, the second will also fail.

Also since the API call responsable to change the status returns `null` as the response, it will be a bit too much to really know wether the CLI did actually change the source status.
We would need 2 API calls to fetch the source status (one at the very beginning and one after the status change).

packages/cli/src/lib/push/addCommon.ts
y-lakhdar: image

packages/cli/src/lib/push/addCommon.ts
y-lakhdar: No. That's the beauty of it. The callback can either return a element of the final array as the element itself or in an array
image

packages/cli/src/lib/push/addCommon.ts
y-lakhdar: true, but I find it harder to read
image

y-lakhdar: If we had a [partition](https://underscorejs.org/#partition) filter, it would have been cleaner.

louis-bompart: You just have to remove the unfixable normalize formating in another func:
https://www.typescriptlang.org/play?ssl=17&ssc=110&pln=17&pc=56#code/C4TwDgpgBACgTgSwHbAIYCMA2ECqSDOArmGAPZzAQAmAYghJlVALxSpIgDcAUAMakFgUfpnIIAXhDoMq+FlAAUAM3qN8ALliIUGbHiIlylWqqoBtALoBKFgD4oAb25RhA-EJUAPXRA1bkaFi4BMRkFNTSjJbyljwu-IJQhEhePn7wAT76oUYRptGssc5QSuSKCe4lplCkSlUy+DZOLi4IdcqmAHQAbqiYCFRNxS1V3kH4nWCE+AAWCg5QnUsqMgA0UEjkALZ9EtSaAOZwEBBIHTKdm3A7-ZKDUAC+VnEtD1AM+NDNI0kpCGPYCZTWbzYY-JadFaMdZXG57KiaUrXVDAABy212dxo5DwqSC50Ylwxt2oVjBLieLwpxQexWOwEIcCQjlGaXWyTxgMe3FpfDcHgxaOJ8OxcFx-x88gUyCmwE07m0BxszHsMsIQmYmqgAHJtVAAPxQKgILYKbUKCBbMCgeqMDaoLYQKzamyaY5UaVIWVWIA

louis-bompart (approved):
LGTM, if tests passes after merge conflict are resolved

olamothe (approved):

PR #788, opened 2022-05-18
feat: setup the Catalog Command

y-lakhdar:
## Proposed changes

Setting command for catalog creation.
Parsing the data and creating the catalog is the first step in the commerce recipe.

### DIFF
--- a/packages/cli/src/commands/commerce/catalog/create.ts
+++ b/packages/cli/src/commands/commerce/catalog/create.ts
@@ -25,14 +25,12 @@ export default class CatalogCreate extends Command {
     withVariants: Flags.boolean({
       char: 'v',
       default: false,
-      exclusive: ['catalogConfiguration'],
       helpGroup: 'Catalog structure',
       description: 'Whether your catalog contains product variants',
     }),
     withAvailabilites: Flags.boolean({
       char: 'a',
       default: false,
-      exclusive: ['catalogConfiguration'],
       helpGroup: 'Catalog structure',
       description:
         'Whether your catalog contains product location and/or availabilities',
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/commerce/catalog/create.ts
```diff
@@ -0,0 +1,139 @@
...
+import {Trackable} from '../../../lib/decorators/preconditions/trackable';
+import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
+
+export default class CatalogCreate extends Command {
+  public static description = `${bold.bgYellow(
+    '(alpha)'
+  )} Create a commerce catalog interactively`;
+
+  public static flags = {
+    // TODO: find an alternative to skip all the prompts without having too many flags.
```
louis-bompart, line 24: Consider this as a UI command, i.e. we don't expect nor will support machine-user for this one

packages/cli/src/commands/commerce/catalog/create.ts
```diff
@@ -0,0 +1,139 @@
...
+      helpGroup: 'Catalog structure',
+      description: 'Whether your catalog contains product variants',
+    }),
+    withAvailabilites: Flags.boolean({
+      char: 'a',
+      default: false,
+      helpGroup: 'Catalog structure',
+      description:
+        'Whether your catalog contains product location and/or availabilities',
+    }),
```
louis-bompart, line 37: can't we instead just do 'if there is at least one availability flag' ?

olamothe (approved):

packages/cli/src/commands/commerce/catalog/create.ts
y-lakhdar: not sure I get your proposition. How would you do that?

louis-bompart: `flags.availabilityFiles.length>0`

y-lakhdar: Will address that in CDX-977

PR #803, opened 2022-05-27
test(create-atomic): add more happy path e2e tests

ThibodeauJF:
https://coveord.atlassian.net/browse/CDX-1012


packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -77,182 +88,205 @@ describe('ui:create:atomic', () => {
...
+    // TODO: CDX-969 update search page id from new stg org.
+    setupTests({pageId: '85fe78b4-2e10-4ed9-a0b7-664f5d23887d'});
+    runTests();
+  });
 
-      expect(await page.$(searchInterfaceSelector)).not.toBeNull();
-    }, 60e3);
+  describe.skip('when using an existing pageId (using the list prompt of available pages)', () => {
+    // TODO: CDX-969 make sure stg org has at least 1 "new" hosted search page.
+    setupTests({promptAnswer: `\033[B${EOL}`});
```
ThibodeauJF, line 289: I really have no clue how to enter an "arrow down" to select one of the search pages 🤔

louis-bompart: A Node terminal works like a VT100, here's a manual: https://vt100.net/docs/vt100-ug/chapter3.html
There's an escape sequence that matches the down arrow key or the down arrow key action (i.e. move cursor 1 line down).
You need to write that in the stdin and voilà

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -77,182 +88,205 @@ describe('ui:create:atomic', () => {
...
+        processManagers.map((manager) => manager.killAllProcesses())
+      );
+    });
+  }
 
-    it('should not contain console errors nor warnings', async () => {
-      await page.goto(searchPageEndpoint, {
-        waitUntil: 'networkidle2',
-      });
+  function runTests() {
```
louis-bompart, line 128: Wrapping `test`, `describe` and `expect` in a function is a no-go, it can cause issues with Jest custom parsing sauce and lead, at best, to decrease performance, at worst, false positive/negative.
I'll get another look very soon to help you refactor that proper ;)

louis-bompart (changes requested):
Looks good overall, the down arrow key and some réfactor

packages/cli-e2e/__tests__/atomic.specs.ts
ThibodeauJF: So 
```suggestion
    setupTests({promptAnswer: `ESC D${EOL}`});
```
?

packages/cli-e2e/__tests__/atomic.specs.ts
louis-bompart: It's mostly OK when everything is 'in band' but for example, Jest is adding sharding in v28, which would be a boon for us to have ootb support to run tests on multiple GHA runners (we're doing it currently, but hackish way).

packages/cli-e2e/__tests__/atomic.specs.ts
ThibodeauJF: My original answer gave me this 
Screen Shot 2022-05-30 at 9 23 35 AM


`\x1B[B` doesn't go down 1 line

louis-bompart:
You should tweak the describe.each to add some kind of prefix or something so that the 'terminal' have differents name, if not, you'll keep on overwriting the precedent one in the log, and well, that's a bit of a pain for debugging

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -92,18 +93,18 @@ export default async function (plop: NodePlopAPI) {
...
         choices: async function (answers: PlopData) {
-          initPlatformClient(answers);
           return [
             {value: undefined, name: 'Start from scratch'},
             {type: 'separator'},
-            ...(await listSearchPagesOptions(platformClient)),
+            ...(await listSearchPagesOptions(initPlatformClient(answers))),
           ];
         },
         validate: async (input, answers: PlopData) => {
```
ThibodeauJF, line 102: Woops, I did a refactor to always fetch the page on the validate method. To my surprise, there is no validation when picking from the list.

scripts/releaseV2/phase2-bump-all-packages.mjs
```diff
@@ -10,7 +10,7 @@ import {
 import {spawnSync} from 'child_process';
```
ThibodeauJF, line 10: https://github.com/coveo/cli/pull/809

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -17,18 +17,19 @@ interface PlopData {
   'platform-url': string;
   'org-id': string;
   'api-key': string;
-  page: IManifest;
 }
 
 export default async function (plop: NodePlopAPI) {
   const currentPath = process.cwd();
-  let platformClient: PlatformClient;
+  let _platformClient: PlatformClient;
```
louis-bompart, line 24: This variable is never assigned.

olamothe (approved):

PR #806, opened 2022-05-30
fix: handle rowid fields in org:search:dump

y-lakhdar:
## Proposed changes

Never exclude `rowid` and `sysrowid` fields from the search query.
They will still be excluded from CSV output if the user adds them to the command.
 

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/src/commands/org/search/dump.ts
```diff
@@ -110,13 +113,24 @@ export default class Dump extends Command {
     throw err;
   }
 
+  private excludeFields(
+    fields: string[],
+    fieldsToExclude: string[] = Dump.mandatoryFields
```
olamothe, line 118: Suggestion: Move this to a `utils` file, and name it `without` or `omit`.

```
function without<T>(array: T[],  values: T[]): T[] {}
```

olamothe (approved):

PR #813, opened 2022-06-02
prompt user to define its catalog from the extracted fields IDs

y-lakhdar:
## Proposed changes

A series of questions to get the catalog required fields.
I used inquirer for the questions to which I had to include a [prompt-autocompletion plugin](https://www.npmjs.com/package/inquirer-autocomplete-prompt) mixed with a [fuzzy search](https://www.npmjs.com/package/fuzzy) to quickly find the desired choice.

Screenshots
image

## Before Merging
- [x] Merge https://github.com/coveo/push-api-client.js/pull/158
- [ ] Bump push api client
- [x] Manual Tests: It creates the catalog and sources as expected in the [organization](https://platformdev.cloud.coveo.com/admin/#clitestmfkr48rt/commerce/catalogs)


louis-bompart (approved):
lgtm for phase-0.
Try to keep the flow on paper, 'cause if we decide to continue after the phase-0, we will probably 'recode' bit by bit so the code is properly reviewed.

PR #814, opened 2022-06-02
try automatic catalog configuration detection

y-lakhdar:
## Proposed changes
Try to automatically generate a catalog configuration without user inputs. 
If the detection logic unexpectedly throw or is unable to find a valid configuration model, it will return a warning and ask the user to answer questions.

The idea is to quickly check if we could generate a valid catalog for the user without any using any prompts.
Otherwise, the user has to go through an unnecessary step, specially if his/her data is already following the Coveo catalog standards.

### Screenshot
Example where the automatic catalog configuration detection returned a valid configuration.
image

Right now it only works for the following catalog configurations:
- [x] Products
- [x] Products x Variants
- [ ] Products x Availabilities
- [ ] Products x Variants x Availabilities

## Performance
Tested with a set of 18,424 variants and 3,339 products (24 MB)
**Total amount of time** : ~940ms
**Total Heap allocated**: ~125 MB (_Don't know why it takes that much space_)

## Funny story
I tried using a combinatorial optimization approach to solve this problem but failed in building a model that could return a solution in a reasonable amount of time... I don't think this approach would be better than a plain ol' algorithm like the one from this PR. [Here](https://github.com/y-lakhdar/catalog-id-field-solver/blob/master/scenarios/product-variant/productVariant.mzn) is the source code if you are interested.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/cli/src/commands/commerce/catalog/create.ts
+++ b/packages/cli/src/commands/commerce/catalog/create.ts
@@ -1,6 +1,5 @@
 import {bold} from 'chalk';
 import {
-  CreateCatalogConfigurationModel,
   CatalogConfigurationModel,
   PlatformClient,
   SourceType,
--- a/packages/cli/src/commands/commerce/catalog/create.ts
+++ b/packages/cli/src/commands/commerce/catalog/create.ts
@@ -25,11 +25,9 @@ import {withSourceVisibility} from '../../../lib/flags/sourceCommonFlags';
 import {getDocumentFieldsAndObjectTypeValues} from '../../../lib/catalog/parse';
 import dedent from 'ts-dedent';
 import {without} from '../../../lib/utils/list';
-
-type PartialCatalogConfigurationModel = Pick<
-  CreateCatalogConfigurationModel,
-  'product' | 'variant' | 'availability'
->;
+import {PathLike} from 'fs';
+import {PartialCatalogConfigurationModel} from '../../../lib/catalog/interfaces';
+import {getCatalogPartialConfiguration} from '../../../lib/catalog/detect';
 
 export default class CatalogCreate extends Command {
   public static description = `${bold.bgYellow(
@@ -77,9 +75,7 @@ export default class CatalogCreate extends Command {
     const authenticatedClient = new AuthenticatedClient();
     const client = await authenticatedClient.getClient();
     const configuration = authenticatedClient.cfg.get();
-    const catalogConfigurationModel = await this.generateCatalogConfiguration(
-      client
-    );
+    const catalogConfigurationModel = await this.generateCatalogConfiguration();
 
     await this.ensureCatalogValidity();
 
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/commerce/catalog/create.ts
```diff
@@ -66,74 +70,134 @@ export default class CatalogCreate extends Command {
...
-        CliUx.ux.warn(
-          `missing ${bold('objecttype')} metadata on item ${uri} (${docPath})`
-        );
-      } else {
-        objectTypeValues.add(metadata.objecttype);
-      }
+    try {
+      CliUx.ux.action.start('Generating catalog configuration from data');
+      return getCatalogPartialConfiguration(flags.dataFiles);
+    } catch (error) {
```
olamothe, line 122: Is handling this as an thrown error the correct thing to do here ?

As opposed to just a "normal" return type from `getCatalogPartialConfiguration` function ?

I'm asking this because, naturally, this does not really seem like an error to me, it's just that we could not the the automatic job. 

So it seems like a denaturation of what an error should be.

packages/cli/src/commands/commerce/catalog/create.ts
```diff
@@ -142,53 +206,75 @@ export default class CatalogCreate extends Command {
...
     const {args} = await this.parse(CatalogCreate);
-    const model: New<CreateCatalogConfigurationModel> = {
-      fieldsMapping: this.getFieldMappings(),
+    return client.catalogConfiguration.create({
+      ...model,
       name: `${args.name}-configuration`,
-      ...this.getHierarchyModel(),
-    };
-    return model;
+      // Field mappings can later be defined in the Admin UI.
```
olamothe, line 220: What are the implications of this comment then ?

olamothe (approved):

packages/cli/src/commands/commerce/catalog/create.ts
```diff
@@ -66,74 +70,134 @@ export default class CatalogCreate extends Command {
...
-      const {metadata, uri} = docBuilder.build();
-      if (metadata?.objecttype === undefined) {
-        CliUx.ux.warn(
-          `missing ${bold('objecttype')} metadata on item ${uri} (${docPath})`
-        );
-      } else {
-        objectTypeValues.add(metadata.objecttype);
-      }
+    try {
+      CliUx.ux.action.start('Generating catalog configuration from data');
```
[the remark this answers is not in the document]
y-lakhdar: Maybe:
> Trying to automatically generate catalog configuration from data
> Automatic catalog generation failed.
> Switching to interactive mode

packages/cli/src/commands/commerce/catalog/create.ts
y-lakhdar: I am using a try catch because this is a very early implementation of the automatic catalog detection feature and a lot of things could go wrong here. 
Also, the detection currently does not work for every catalog type (e.g. catalogs with availabilities).

So to keep it simple, I just catch any error and redirect to the interactive mode.

louis-bompart (approved):
I didn't see any red flags, so lgtm for phase 0.

PR #821, opened 2022-06-03
feat(create-atomic): display npm install error details to user

ThibodeauJF:
https://coveord.atlassian.net/browse/CDX-1028

## Now:
Screen Shot 2022-06-03 at 5 08 01 PM


y-lakhdar (approved):

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -184,15 +184,20 @@ export default async function (plop: NodePlopAPI) {
...
             const process = spawn(getPackageManager(), ['install'], {
-              stdio: 'ignore',
+              stdio: ['ignore', 'ignore', 'pipe'],
               cwd: join(currentPath, project),
             });
 
+            let error = '';
+            process.stderr.on('data', (chunk) => {
+              error += chunk.toString();
+            });
```
louis-bompart, line 194: Logging just the stderr won't be enough IMO.
You should log stdout too for full context (Yeah, it'll be noisy, but when 💩 hit the fan, I think it's best to have too much than not enough info).

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -184,15 +184,20 @@ export default async function (plop: NodePlopAPI) {
...
+
             process.on('close', (code) => {
```
louis-bompart, line 196: https://nodejs.org/api/child_process.html#event-close gimme the signal too. When reported, it's often a bliss in term of information (i.e. the signal is often very specific and so with it fixing a problem is often a breeze [^1])

[^1]: famouslastwords.

louis-bompart (commented):
some change proposals

packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -184,20 +184,27 @@ export default async function (plop: NodePlopAPI) {
...
+            let out = '';
+            process.stdout.on('data', (chunk) => {
+              out += chunk.toString();
+            });
+
+            let err = '';
             process.stderr.on('data', (chunk) => {
-              error += chunk.toString();
+              err += chunk.toString();
             });
```
louis-bompart, line 199: this is opinionated, so feel free to not do as such.

IMO, it's best in this kind of case to have only one variable that is 'fed in'.
The idea is that as a user, you don't see a difference between stdout and stderr usually.

Moreover, on the present case, you'll log everything in stderr, so the distinction is even less valuable.

However, programs may log things in stdout that relies on human context that has been logged in stdout previously (e.g. `console.log('hello'); console.error('world');`, if you only have a lot of things like that, it'll be tough to understand what happen)

From the example, you can see that timing is more important here (IMO still).

So in essence, I would remove `err` and replace it by `out` in the stderr listener.

louis-bompart (approved):
lgtm, a suggestion, feel free to pick or leave it tho

packages/atomic/create-atomic/src/plopfile.ts
ThibodeauJF: > Moreover, on the present case, you'll log everything in stderr, so the distinction is even less valuable.

I kind of knew I only had stderr before... https://github.com/coveo/cli/pull/821#discussion_r890260786

PR #827, opened 2022-06-07
feat(atomic): create impersonate key inside create-atomic instead

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1032

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/atomic/create-atomic/src/plopfile.ts
```diff
@@ -38,6 +38,20 @@ export default async function (plop: NodePlopAPI) {
     return platformClientInstance;
   }
 
+  async function createImpersonateApiKey(
```
y-lakhdar, line 41: If we move that step at this early stage, what will the user get if he does not have the privileges to create API keys?

y-lakhdar (approved):

PR #828, opened 2022-06-08
fix(dump): csv header should have all fields associated with the source

louis-bompart:
## Proposed changes

New algo:
1. Search until dry or `chunkSize` amount of results is in RAM.
2. Append all fields name found into an array that stay alive till the end of the process.
3. Dump the aggregatedResults into a file, clear aggregatedResults and replace it with the overhead if any.
4. Rince and repeat till out of results.
5. When done, transform the array of field names into a set and into an array again (ensure unicity) then convert all chunks file into CSV.

## Testing

- [X] Unit Tests
- [X] Manual Tests: done, see slack.

### DIFF
--- a/packages/cli/src/commands/org/search/dump.ts
+++ b/packages/cli/src/commands/org/search/dump.ts
@@ -1,20 +1,26 @@
 import {Command, Flags, CliUx} from '@oclif/core';
+import {readJSONSync, writeFile, writeJSONSync} from 'fs-extra';
+import {Parser} from 'json2csv';
+import {SingleBar} from 'cli-progress';
+import PlatformClient from '@coveord/platform-client';
+import {dirSync} from 'tmp';
+
 import {AuthenticatedClient} from '../../../lib/platform/authenticatedClient';
 import {
   Preconditions,
   IsAuthenticated,
 } from '../../../lib/decorators/preconditions';
 import {Config} from '../../../lib/config/config';
-import PlatformClient from '@coveord/platform-client';
-import {writeFile} from 'fs-extra';
-import {Parser} from 'json2csv';
-// eslint-disable-next-line node/no-extraneous-import
-import {SingleBar} from 'cli-progress';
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/org/search/dump.ts
```diff
@@ -81,18 +89,25 @@ export default class Dump extends Command {
     }),
   };
 
+  private aggregatedResults: RawResult[] = [];
+  private aggregatedFieldsWithDupes: string[] = [];
+  private temporaryDumpDirectory = dirSync();
```
olamothe, line 94: Would it be better that the temp directory is generated when `run()` is executed ?

packages/cli/src/commands/org/search/dump.ts
```diff
@@ -32,7 +38,9 @@ interface FetchParameters {
 }
 
 export default class Dump extends Command {
+  private static readonly numberOfResultPerQuery = 1000;
```
olamothe, line 41: Quick win might be to also make this a command flag, default to 1000

Would help with having to exclude a ton of fields when they hit API limits.

ie: instead of clients having to exclude a lot of field, ask them to fiddle with/lower that number.

Could also detect internally when we get a max size response, and automatically retry with something like `numberOfResultPerQuery = numberOfResultPerQuery / 2`

olamothe (approved):

packages/cli/src/commands/org/search/dump.ts
louis-bompart: I'm working on this, there's a Jira for that CDX-986.

I'm taking the 'automatic' approach.

packages/cli/src/commands/org/search/dump.ts
louis-bompart: Good call, this actually cause the code to generate a temp folder at each command run, even others than dump. Oups.

PR #829, opened 2022-06-08
feat(dump): add automatic request size backoff

louis-bompart:
## Proposed changes

Add a mechanism to reduce automatically the amount of results per query if the API throws a `ResponseExceededMaximumSizeException` error.
If we reach 0 results in the query to be sent, we exit with an error and ask the user to exclude fields (that's pretty much the best we can do at this stage)
If another error occurs, we bubble it up, like before.

## Testing

- [X] Unit Tests: yep
- [X] Manual Tests: nope, corner case


y-lakhdar (approved):
nice algorithm

PR #830, opened 2022-06-09
feat(angular): update to angular 14

louis-bompart:
## Proposed changes

Updated Angular to 14, small changes here and there to match the breaking changes. No change in behavior

## Testing

- [x] Unit Tests:
- [x] Functionnal Tests:
- [x] Manual Tests:


packages/angular/src/search-token-server/files/scripts/port-allocator.mjs
```diff
@@ -81,9 +81,11 @@ const isPortAvailable = async (port, options) => {
 
 function updateAppPort(port) {
   const angularJsonPath = resolve('angular.json');
+  const packageJsonPath = resolve('package.json');
```
louis-bompart, line 84: No more projectDefault. Sadge. Using package name instead, same story in the end

packages/cli/src/lib/decorators/preconditions/ng.spec.ts
```diff
@@ -1,7 +1,7 @@
 jest.mock('../../utils/process');
```
louis-bompart, line 1: Simplified those tests to use snapshot. To have os-agnostic snapshot I mock everything and tada.

packages/cli/src/lib/decorators/preconditions/ng.ts
```diff
@@ -10,6 +10,7 @@ export const IsNgVersionInRange = getBinVersionPrecondition(
     howToInstallBinText:
       'You can install the Angular-CLI by running npm i -g @angular/cli',
     installLink: 'https://angular.io/guide/setup-local#install-the-angular-cli',
+    params: ['version'],
```
louis-bompart, line 13: No more --version flag in 13. 13 and 14 support `ng version`, so let's use that. (no idea for before, but well, gee)

olamothe (approved):

PR #841, opened 2022-06-20
test: fix daily tests 🤞 

louis-bompart:
## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -38,19 +38,20 @@ describe('ui:create:atomic', () => {
     processManager: ProcessManager,
     options: BuildAppOptions
   ) => {
-    let stderr = '';
-    const stderrListener = (chunk: string) => {
-      stderr += chunk;
-    };
-
+    let output = '';
```
louis-bompart, line 41:
```suggestion
    let output = '';
    
```

packages/cli-e2e/__tests__/atomic.specs.os.ts
```diff
@@ -92,7 +94,7 @@ describe('ui:create:atomic', () => {
       .do(answerPrompt(`y${EOL}`))
       .until(buildTerminalExitPromise);
 
-    return {stderr};
+    return {stderr: output};
```
y-lakhdar, line 97: stdout in stderr?

louis-bompart: derp, i'll rename the var output.
Long story short, when using the published version of the CLI, everything ends up in stdout

y-lakhdar (approved):

PR #848, opened 2022-06-22
fix: verify org with appropriate region and env

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-1041

## Proposed changes

Disable --region and --environment flags as it requires the user to get a new search token.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


olamothe (approved):

packages/cli/src/commands/config/set.ts
```diff
@@ -35,20 +38,19 @@ export default class Set extends Command {
   @Preconditions(IsAuthenticated())
   public async run() {
     const {flags} = await this.parse(Set);
+    if (flags.environment || flags.region) {
```
olamothe, line 41: That's a good deprecation trick. 💯

PR #849, opened 2022-06-23
feat(snapshot): add `deletionScope` on `/apply`

louis-bompart:
## Proposed changes

We do not want to delete everything when we apply a snapshot 😂 

## Testing

- [x] Unit Tests:

- [x] Functional Tests:

- [ ] Manual Tests:


packages/cli/src/lib/snapshot/snapshot.ts
```diff
@@ -108,7 +108,13 @@ export class Snapshot {
     deleteMissingResources = false,
     options: WaitUntilDoneOptions = {}
   ) {
-    await this.snapshotClient.apply(this.id, {deleteMissingResources});
+    await this.snapshotClient.apply(
+      this.id,
+      deleteMissingResources
+        ? //@ts-ignore
```
louis-bompart, line 114: https://github.com/coveo/platform-client/pull/470

PR #858, opened 2022-07-08
chore: san-check flag and argument consistency

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-1044

## Proposed changes

No code change. Only `TODOs` to tackle in other PRs

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/cli/src/commands/org/resources/push.ts
+++ b/packages/cli/src/commands/org/resources/push.ts
@@ -41,8 +41,7 @@ export default class Push extends Command {
 
   public static flags = {
     ...wait(),
-    ...sync(),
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/org/resources/push.ts
```diff
@@ -52,6 +51,11 @@ export default class Push extends Command {
       default: false,
       required: false,
     }),
+    // TODO: CDX-1004: remove skipPreview flag
```
louis-bompart, line 54: Very good point 👏

louis-bompart (approved):

PR #860, opened 2022-07-08
feat: remove skipPreview flag

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-1004

## Proposed changes

Remove redundant flag and add a preview level instead.

## Breaking changes

Still in beta, so it's fine...

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/cli/src/commands/org/resources/push.ts
+++ b/packages/cli/src/commands/org/resources/push.ts
@@ -51,18 +51,6 @@ export default class Push extends Command {
       default: false,
       required: false,
     }),
-    // TODO: CDX-1004: remove skipPreview flag
-    // Reasons:
-    //  * Skip logic can be taken care by the preview level (adding a None level)
-    //  * The flag char is already been used by another command and can create confusion
-    //  * Less flags, the better
-    skipPreview: Flags.boolean({
-      char: 's',
-      description:
-        'Do not preview changes before applying them to the organization',
-      default: false,
-      required: false,
-    }),
   };
 
   @Trackable()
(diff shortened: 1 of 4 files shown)


packages/cli/src/lib/flags/snapshotCommonFlags.ts
```diff
@@ -2,7 +2,7 @@ import {Flags} from '@oclif/core';
 import {Snapshot} from '../snapshot/snapshot';
 
 export enum PreviewLevelValue {
-  // TODO: CDX-1004: add a "None" value and remove the skipPreview flag
+  Skip = 'skip',
```
y-lakhdar, line 5: There are also other good candidates:
* `none`
* `off`
* `ignore`

which one would make more sense?

louis-bompart: I'd say, `None`, when compared to the other value of the enum it's the one that make the most sense

It's not the most user-friendly way, but the most user-friendly way will I think come when we will have a `--yes` flag overall

louis-bompart (approved):
LGTM ⚠️ do the refacto mentioned lower tho

packages/cli/src/commands/org/resources/push.ts
```diff
@@ -84,7 +72,7 @@ export default class Push extends Command {
       options
     );
 
-    if (!flags.skipPreview) {
+    if (flags.previewLevel !== PreviewLevelValue.Skip) {
```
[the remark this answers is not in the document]
louis-bompart:
```suggestion
    if (!(await this.shoudlSkipPreview())) {
```

PR #866, opened 2022-07-14
fix: correctly support `--snapshotId` flag

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-1078

## Proposed changes

* Enable preview for existing snapshot
* Make sure to not delete the snapshot if it was already present in the organzation

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/cli/src/commands/org/resources/preview.ts
+++ b/packages/cli/src/commands/org/resources/preview.ts
@@ -100,8 +100,13 @@ export default class Preview extends Command {
     return flags.previewLevel === PreviewLevelValue.Detailed;
   }
 
+  private async shouldDeleteSnapshot() {
+    return !(await this.parse(Preview)).flags.snapshotId;
+  }
   private async cleanup(snapshot: Snapshot, project: Project) {
-    await snapshot.delete();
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/org/resources/preview.ts
```diff
@@ -117,7 +122,7 @@ export default class Preview extends Command {
 
           Once the snapshot is created, you can preview it with the following command:
 
-            ${blueBright`coveo org:resources:preview -t ${target} -s ${snapshot.id}`}
+            ${blueBright`coveo org:resources:preview -o ${target} -s ${snapshot.id}`}
```
y-lakhdar, line 125: Ninja fix.
I'll have to do a sanity check at some point to dynamically generate these commands instead of hard coding them

louis-bompart: Yeah fair enuf. Outta see tho: command are not meant to change, 'cause breaking. So prolly not really worth it

louis-bompart (approved):

PR #869, opened 2022-07-21
feat(analytics): remove persistent analytics identifers

louis-bompart:
## Proposed changes

Removed persistent identifiers, such as the session id. Instead, Amplitude will manage to find whos whos using the other informations

## Testing

- [x] Unit Tests


tcamire (approved):
makes sense! This should really not affect session tracking, but I'll test it in dev once its merged to make sure its ok

y-lakhdar (approved):

PR #870, opened 2022-07-21
feat(analytics): always track

louis-bompart:
## Proposed changes

- Removed the possibility of disabling analytics [^1]
- Track API key user through a hash of the last subset of said API Key.

## Testing

- [x] Unit Tests

[^1]: _technically_ still possible by having CI=true in your env var


packages/cli/src/hooks/analytics/identifier.ts
```diff
@@ -42,15 +45,31 @@ export class Identifier {
...
+  private async getAnalyticsInfo(platformClient: PlatformClient) {
+    return this.configuration.anonymous
+      ? this.getApiKeyInfo()
+      : this.getUserInfo(platformClient);
+  }
+
+  private async getApiKeyInfo() {
+    const identifier = this.configuration.accessToken
+      ?.split('-')
+      .pop() as string;
```
y-lakhdar, line 62: Why use the last subset and not the entire thing

y-lakhdar (approved):

tcamire (commented):
not sure that I see a condition that blocks tracking if the user is not logged in, will that be in this PR? Correct me if I'm wrong, it's my first time roaming around in this code

y-lakhdar:
> not sure that I see a condition that blocks tracking if the user is not logged in, will that be in this PR? Correct me if I'm wrong, it's my first time roaming around in this code

Good catch. Most commands require user to be logged in both some few can be executed without being logged in.

y-lakhdar (changes requested):
Only track when user is logged in.

louis-bompart:
https://github.com/coveo/cli/blob/d81ffb03323acada148e1fa69c01fd4d8b60e0fa/packages/cli/src/hooks/analytics/analytics.ts#L18-L21 this does that

y-lakhdar (approved):

PR #872, opened 2022-07-26
chore: bump platform-client

y-lakhdar:
## Proposed changes

Bumping platform-client and remove legacy contract

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/cli/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -197,25 +186,22 @@ export class SnapshotReporter {
...
   ) {
     this.resourceInErrorCount++;
     let errorSet = this.resourceInError.get(resourceType);
     if (!errorSet) {
       errorSet = new Set();
       this.resourceInError.set(resourceType, errorSet);
     }
-    // TODO CDX-1038: Clean that.
-    errorSet.add(typeof err === 'string' ? err : err.message);
+    errorSet.add(`${err.resultCode}: ${err.message}`);
```
y-lakhdar, line 197: Should we include the resultCode in the error string?
I think it would not hurt

louis-bompart (approved):
Lgtm

PR #873, opened 2022-07-26
feat(cli): remove beta tag for snapshot

louis-bompart:
## Proposed changes

Removed the beta mentions

## Testing

- [x] Unit Tests (jest snapshots)


packages/cli/src/lib/decorators/preconditions/__snapshots__/ng.spec.ts.snap
```diff
@@ -0,0 +1,33 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
louis-bompart, line 1: Yeah it should have been commited before oopsies. Fixed.

y-lakhdar (approved):

PR #874, opened 2022-07-26
test(atomic): time optimization

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1061

## Proposed changes

- Fixed issue with the `skipInstall` option
- Replace 'runtime' tests with file checks for 3/4 of tests: this requires to run separately on win & Linux, because the hash changes due to varying EoL

Because this is the introduction of 'Jest Snapshots' for some of our tests, I added the concept of `specs.os.ts` test file & a manual workflow to update the snapshots when required, on both win&linux


.github/workflows/update-os-snapshots.yml
```diff
@@ -0,0 +1,253 @@
+name: Refresh E2E Snapshots
+on:
+  push:
+    branches:
+      - 'CDX-1061'
+  workflow_dispatch:
```
louis-bompart, line 6:
```suggestion
on:
  workflow_dispatch:
```

ThibodeauJF (approved):
I like the 📸 !

.github/workflows/update-os-snapshots.yml
```diff
@@ -0,0 +1,250 @@
+name: Refresh E2E Snapshots
```
louis-bompart, line 1: Pretty much a copy of .github/workflows/build.yml

There's a bigger effort to DRY the workflow one day ™️ 
Out of scope for now. I did not put TODO either because it's best to study all the workflows together before slating work

olamothe, line 1: Those are some complicated config files, with cryptic values... 😩

.github/workflows/update-os-snapshots.yml
```diff
@@ -0,0 +1,250 @@
...
+      - uses: actions/setup-node@2fddd8803e2f5c9604345a0b591c3020ee971a93 # tag=v3
+        with:
+          node-version: ${{matrix.node}}
+          cache: 'npm'
+      - name: Install locked NPM version.
+        # TODO: CDX-1010 | Remove in CDX-1010.
+        run: npm i -g npm@8.10.0
+      - run: |
+          npm ci
+          Xvfb :1 -screen 0 1024x768x16 & sleep 1
```
olamothe, line 36: II think a small comment that explains why these are needed for chrome to work properly (I assume?) would be nice ;)

olamothe (approved):

.github/workflows/update-os-snapshots.yml
```diff
@@ -0,0 +1,250 @@
...
+          xdg-settings set default-web-browser google-chrome.desktop
+      - name: Build
```
louis-bompart, line 38:
```suggestion
      - name: Install dependencies
         run: npm ci
      # `coveo auth:login` is always opening the default browser of the user, this ensure that the default browser is Chrome and that it'll be able to open by setting a X Display as the default Display
      - name: Prepare Google Chrome
         run: |
           Xvfb :1 -screen 0 1024x768x16 & sleep 1
           xdg-settings set default-web-browser google-chrome.desktop
      - name: Build
```

.github/workflows/update-os-snapshots.yml
louis-bompart: I agree that it's quite complicated and much love is required to get it more readable.
For now, I'm going on the 'just make it work' (that's why it's a copy of the CI workflow)

I want to keep the scope tight and not affect build.yml too much, but I planned an issue to refactor our workflow to match the quality of our more recent projects.

The overall complexity is overall inherent to the nature of what we test, which is essentially a native application, with bindings deep in the browser OS.


TL;DR: The overall complexity is not really shrinkable. It can be made more manageable by refactoring and splitting it into reusable (and digestible) actions. However, I keep that for another PR

PR #886, opened 2022-08-12
test(atomic): go back to os-agnostic specs+prod compat

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1108

## Proposed changes

- Reunify win/linux specs for atomic
- Enforce LF on verdaccio publish in CI
- Make sure `build` phase in atomic E2E is also binding to stdout when it was only binding to stderr (so that it's compat with the published version)

## Testing

https://github.com/coveo/cli/actions/runs/2844470266 If you check this run, you'll see some hash that doesn't match on windows. If you look up the hash even on master, you'll see that they do exist for Linux.

Tho, the testing itself won't be done for real till next release+daily 😅


.github/workflows/build.yml
```diff
@@ -115,6 +115,11 @@ jobs:
     name: End-to-end publish packages
     runs-on: ${{matrix.os}}
     steps:
+      - name: Setup runner
+        if: ${{matrix.os == 'windows-latest'}}
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
```
louis-bompart, line 122: This fix the 1st issue: Inconsistent end of line between the published version and the verdaccio version

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -79,20 +79,27 @@ describe('ui:create:atomic', () => {
...
-      .until(buildTerminalExitPromise);
+    await Promise.allSettled(
+      streams.map((stream) =>
+        buildTerminal
+          .when(/\(y\)/)
+          .on(stream)
+          .do(answerPrompt(`y${EOL}`))
+          .until(buildTerminalExitPromise)
+      )
+    );
```
louis-bompart, line 102: This ensure that whatever we did on stderr we at least try to do it on stdout

y-lakhdar (approved):

PR #893, opened 2022-08-18
fix(dump): remove recursion

louis-bompart:
CDX-1117

## Proposed changes
image
This is often due to recursion that goes too far.
The only recursion in the search dump is the one removed in this PR.

## Testing

Corner case, so no automated tests.
`@coveo/cli@1.34.3-0` was released as alpha for validation on customer organization by support. waiting for confirmation from them.

### DIFF
--- a/packages/cli/src/commands/org/search/dump.ts
+++ b/packages/cli/src/commands/org/search/dump.ts
@@ -219,9 +219,6 @@ export default class Dump extends Command {
       indexToken = response.indexToken;
       rowId = response.results[lastResultsLength - 1].raw.rowid;
     } while (lastResultsLength >= this.numberOfResultPerQuery);
-    if (this.aggregatedResults.length > 0) {
-      this.dumpAggregatedResults();
-    }
     this.progressBar.stop();
     return this.progressBar.getTotal() > 0;
   }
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/org/search/dump.ts
```diff
@@ -279,19 +276,18 @@ export default class Dump extends Command {
...
-    const amountOfNewResultsToExtract =
-      maxAggregatedResults - this.aggregatedResults.length;
-    this.aggregatedResults.push(
-      ...newResults.splice(
-        0,
-        Math.min(amountOfNewResultsToExtract, newResults.length)
-      )
-    );
-    if (newResults.length > 0) {
+    while (results.length > 0) {
```
louis-bompart, line 281: We can refactor like this because the splice was already noop if `results.length===0` :
1. `Math.min(amountOfNewResultsToExtract, results.length) == 0`
2. `[].splice(0,0) == []`
3. array.push(...[]) does nothing


However, that makes the method dumping the result no matter what in the end.
This was first made by the other chunk that was removed above.
I checked the order of operations and I don't think that it would change anything.

louis-bompart:
- [ ] Provide Jira #
- [ ] Support validated that the alpha version fix the issue

y-lakhdar (approved):

PR #898, opened 2022-08-24
refactor(cli): extract utils for source plugin

louis-bompart:
## Proposed changes

Extract all the things we need from `@coveo/cli` into separate packages for the extraction of `@coveo/plugin-cli-source`.
The `@coveo/cli-commons` is for what's used in multiple commands/plugins in _production_, `@coveo/cli-commons-dev` is for the dev stuff, and won't be used for end-user.

## Testing

- [x] Unit Tests: no new tests
- [x] Manual Tests: yep, **but**: release will be important, it's verdaccio tested but that's all.


packages/cli/commons/src/errors/preconditionError.ts
```diff
@@ -9,7 +9,7 @@ export enum PreconditionErrorCategory {
 }
 
 export interface PreconditionErrorOptions {
-  category?: PreconditionErrorCategory;
+  category?: PreconditionErrorCategory | string;
```
y-lakhdar, line 12: Why string?

packages/cli/commons/src/preconditions/apiKeyPrivilege.ts
```diff
@@ -34,7 +33,7 @@ export function HasNecessaryCoveoPrivileges(
             Boolean(anonymous)
           );
           throw new PreconditionError(message, {
-            category: PreconditionErrorCategory.MissingPlatformPrivilege,
+            category: PRECONDITION_ERROR_CATEGORY,
```
y-lakhdar, line 36: Why redefining the variable?

packages/cli/core/src/commands/org/resources/pull.ts
```diff
@@ -218,7 +218,7 @@ export default class Pull extends Command {
     const flags = await this.getFlags();
     return flags.model
       ? flags.model.resourcesToExport
-      : buildResourcesToExport(flags.resourceTypes);
+      : buildResourcesToExport(flags.resourceTypes!);
```
y-lakhdar, line 221: Why the Non-null assertion operator?
If not specified by the user, flags.resourceTypes should be an empty array

packages/cli/core/src/commands/org/search/dump.ts
```diff
@@ -144,7 +144,7 @@ export default class Dump extends Command {
 
   private async convertRawChunksToCSVs() {
     const {flags} = await this.parse(Dump);
-    const fields = this.getFieldsToExport(flags);
+    const fields = this.getFieldsToExport(flags.fieldsToExclude!);
```
y-lakhdar, line 147: same as above

packages/cli/commons/src/preconditions/apiKeyPrivilege.ts
louis-bompart: Because the precondition error might be defined in their own packages, the enum doesn't work anymore.
So I used const instead

packages/cli/commons/src/errors/preconditionError.ts
louis-bompart: To allow a plugin to define its own precondition error

packages/cli/core/src/commands/org/search/dump.ts
louis-bompart: Oclif changed their flag typing 🤷

packages/cli/commons/src/preconditions/apiKeyPrivilege.ts
y-lakhdar: Ok, in that case, I would just get rid of the enum and only accept a string

olamothe (approved):

PR #900, opened 2022-08-25
fix: deploy in hipaa

louis-bompart:
Deploy in HIPAA


.deployment.config.json
```diff
@@ -30,6 +39,9 @@
         "bucket": "{terraform.infra.infra.bucket_binaries}",
         "directory": "proda/StaticCDN/cli",
         "source": "./artifacts",
+        "hip": {
```
olamothe, line 42: Just make sure about the syntax on this one, I don't remember exactly if this is correct ... 🤔

olamothe (approved):

.deployment.config.json
louis-bompart: Yeah I checked the ref doc the e.g. was in yml, I just translated. I'm waiting for a 👍 👎 from devtooling tho

PR #909, opened 2022-08-30
test(e2e): publish cli in verdaccio

louis-bompart:
## Proposed changes

So far, when running the E2E tests, we were running them against 'dev' binary, which transpile on the fly.
This was not representative enough of what a enduser would do.

Now, we publish & install the cli from the verdaccio registry

## Testing

-E2E


ThibodeauJF (approved):
Elegant!

PR #911, opened 2022-08-31
chore(build): use `npm version --save`

louis-bompart:
## Proposed changes

Use `npm version --save` for the package bumps, it saves time, and remove code

## Testing

E2E CI


ThibodeauJF (approved):

louis-bompart:
Further investigation lead me to believe that the process doesn't work.

louis-bompart:
Req https://github.com/npm/cli/issues/5506 fixed

PR #914, opened 2022-09-01
chore: refactor base error

y-lakhdar:
## Proposed changes

Refactor the base error 
 - Remove PrintableError. Error formatting will be handled by the Base Command
 - Apply change to all existing errors

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


louis-bompart (commented):
Wondering if we should exposes less stuff in the CLIBaseError, so that it's less confusing for external implementer

louis-bompart (approved):

PR #915, opened 2022-09-01
chore: tracking standardization

y-lakhdar:
## Proposed changes

**Standardization part 2/3**

### Ensure decorators support non-void return functions.
Required to implement the base command but also to use the [--enableJsonFlag](https://oclif.io/docs/json) option on  commands.
* Build analytic event using the wrapError

## Testing

Available in #916


packages/cli/commons/src/errors/wrapError.ts
```diff
@@ -0,0 +1,23 @@
+import {APIError, isAxiosError, isErrorFromAPI} from './apiError';
+import {CLIBaseError} from './cliBaseError';
+import {UnknownError} from './unknownError';
+
+export function wrapError(err: unknown): CLIBaseError {
```
olamothe, line 5: This function could be easily unit tested.

packages/cli/commons/src/preconditions/trackable.ts
```diff
@@ -41,9 +46,9 @@ export function Trackable({
       };
 
       if (cmdArgs.length > 0) {
-        await trackError.call(this, properties, originalCommand, cmdArgs);
+        return trackError.call(this, properties, originalCommand, cmdArgs);
       } else {
-        await trackCommand.call(this, name, properties, originalCommand);
+        return trackCommand.call(this, name, properties, originalCommand);
```
louis-bompart, line 51: Could be ternarized

louis-bompart (approved):

PR #916, opened 2022-09-01
chore: command standardization

y-lakhdar:
## Proposed changes

**Standardization part 3/3**

* add a base command from which all other commands inherit
* move redundant error tracking into the base command

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


olamothe (approved):

packages/cli/commons/src/errors/cliBaseError.ts
```diff
@@ -43,6 +43,10 @@ export class CLIBaseError extends Error {
         color = yellow;
         break;
 
+      case SeverityLevel.Info:
+        color = cyan;
+        break;
+
       default:
```
louis-bompart, line 50: default and Error are the same, this could be more easily reflected with
```
default:
case SeverityLevel.Error:
  color = red;
  break;
```

And also: DRY

packages/cli/commons/src/utils/ux.ts
```diff
@@ -0,0 +1,16 @@
+import {CliUx} from '@oclif/core';
+import {red, green} from 'chalk';
+
+export function starSpinner(task: string) {
+  if (CliUx.ux.action.running) {
+    CliUx.ux.action.stop(green('✔'));
```
louis-bompart, line 6: Ain't the green check mark a bit optimisitic?
What make you sure that if we start another spinner the previous was successful?

packages/cli/commons/src/utils/ux.ts
```diff
@@ -0,0 +1,16 @@
...
+  }
+  CliUx.ux.action.start(task);
+}
+
+export function stopSpinner(err?: Error) {
```
louis-bompart, line 11: Not sure to understand, why a spinner can only finish with an error?

y-lakhdar: Now that you mention it, I think the error message in the spinner is redundant. We can just have an boolean argument like
`stopSpinner(success = true)`

packages/cli/commons/src/utils/ux.ts
y-lakhdar: Unless you intentionally wrap an async task (along with a spinner) within a try-catch block so you can continue with the execution, any caught error by the base command will stop the spinner with a failure.

I guess, if you manually prevent the command class from handling the error, you should also handle the spinner behaviour with `stopSpinner(err)`

louis-bompart (approved):

PR #919, opened 2022-09-06
fix(cli): parse region flag

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1139

## Proposed changes

In the OAuth library, we compare the region using the 'value' of the enum, and in the flag we use the 'key' of the enum.
This causes the comparison to fail (`'US' !== 'us'`).

So to fix that, I sanitize the input and ensure everything is lowercase only. I also improved the region flag to accept both the lower and upper case variations: it can be a pain to have to go upper case just for typing a region.

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests: tested with /bin/dev:
  - Run `bin/dev auth:login -r=us -e=hipaa`: it fails before, now it works
  - Run `bin/dev auth:login -r=US -e=hipaa`: before, the OAuth redirection was dropped, now it works.


packages/cli/core/src/lib/flags/platformCommonFlags.ts
```diff
@@ -9,7 +9,11 @@ import {
 export const withRegion = (withDefault = true) => ({
   region: Flags.string({
     char: 'r',
-    options: Object.keys(Region),
+    options: Object.keys(Region).flatMap((region) => [
```
samisayegh, line 12: What if we called [Object.entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) here? Would this remove a constraint/assumption that the region and value are the same aside from case?

samisayegh (approved):

packages/cli/core/src/lib/flags/platformCommonFlags.ts
louis-bompart: We still have to do a flatMap here, because otherwise you'd get a `string[][]`.

However it's a good idea to tie ourselves on both the value and the key instead of doing some processing (i.e. `toLowerCase`).

packages/cli/core/src/lib/flags/platformCommonFlags.ts
```diff
@@ -9,7 +9,11 @@ import {
...
+      region,
+      region.toLowerCase(),
+    ]),
```
louis-bompart, line 15:
```suggestion
    options: Object.entries(Region).flatMap(regionEntry=>regionEntry),
```

packages/cli/core/src/lib/flags/platformCommonFlags.ts
louis-bompart: https://github.com/coveo/cli/pull/919/files/958049fb2281199d6bf28bb2816d5eef3b98366c#r963916816

packages/cli/core/src/commands/auth/login.spec.ts
```diff
@@ -91,16 +91,26 @@ describe('auth:login', () => {
...
-      .stdout()
-      .stderr()
-      .command(['auth:login', '-r', region, '-o', 'foo'])
-      .it(`passes the -e=${region} flag to oauth and configuration`, () => {
-        expect(mockedOAuth.mock.calls[0][0]?.region).toBe(region);
-        expect(mockConfigSet).toHaveBeenCalledWith('region', region);
-      });
-  });
+  Object.keys(Region)
+    .flatMap((region) => [region, region.toLowerCase()])
```
louis-bompart, line 95:
```suggestion
  Object.entries(Region)
    .flatMap(regionEntry=>regionEntry)
```

ThibodeauJF (approved):

PR #922, opened 2022-09-06
test(atomic): improve page selection setup

louis-bompart:
## Proposed changes

Improve the instrumentation of the CLI in the E2E to ensure we select a specific page in the `ui:create:atomic` interactive prompt session.

## Testing

FT; manual & CI/CD


packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -151,7 +178,7 @@ describe('ui:create:atomic', () => {
         'when using an existing pageId (using the list prompt of available pages)',
       buildAppOptions: {
         id: 'with-page-id-prompt',
-        promptAnswer: `\x1B[B ${EOL}`,
+        selectWithPrompt: true,
```
ThibodeauJF, line 181: Should there be a `pageIdOrName` here too?

louis-bompart: yep, changed and fixed

packages/cli-e2e/utils/terminal/do.ts
```diff
@@ -70,6 +70,9 @@ export class Do implements Promise<void> {
    */
   public until(untilCondition: Condition): Promise<void> {
     let eventListener: EventListenerObject | undefined;
+    if (untilCondition instanceof RegExp) {
+      untilCondition = new RegExp(untilCondition);
+    }
```
louis-bompart, line 75: I need to 'clone' the regexp instance, otherwise if its used more than once, the removal of one would remove all uses.

Example above

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -51,6 +51,42 @@ describe('ui:create:atomic', () => {
...
+    );
+    const expectedPageSelectedMatcher = new RegExp(
+      `${osSpecificSelector} ${options.pageName}`,
+      'gm'
+    );
+    buildTerminal
+      .when(anotherPageSelectedMatcher)
+      .on('stdout')
+      .do(answerPrompt('\x1B[B'))
+      .until(expectedPageSelectedMatcher);
```
louis-bompart, line 82: First use of said regexp

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -51,6 +51,42 @@ describe('ui:create:atomic', () => {
...
+    buildTerminal
+      .when(expectedPageSelectedMatcher)
```
louis-bompart, line 84: A second use of said regexp: it was never triggered because the `until` above removed it first, because it was the first attached, and FIFO apply to listener [^1]

[^1]: https://nodejs.org/api/events.html#asynchronous-vs-synchronous

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -71,12 +109,10 @@ describe('ui:create:atomic', () => {
...
-    if (!options.pageId) {
-      await buildTerminal
-        .when(/Use an existing hosted search page/)
-        .on('stdout')
-        .do(answerPrompt(options.promptAnswer ? options.promptAnswer : EOL))
-        .once();
+    if (options.pageName) {
+      selectPageWithCli(options, buildTerminal);
+    } else if (!options.pageId) {
+      selectDefaultWithCli(buildTerminal);
```
louis-bompart, line 115: Refactored for clarity:

Sure we can have 'one-stop-shop' function for setting up, but let's split the bits so we don't have to read heaps of code useless for XYZ test cases

olamothe (approved):

ThibodeauJF (approved):

PR #925, opened 2022-09-07
chore: build all projects when building binaries

louis-bompart:
## Proposed changes

Now that we have a composite project for `@coveo/cli`, we need to build the reference first (or use `tsc --build`).

I chose the later to just reuse what's already there

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -42,7 +42,6 @@ jobs:
       - name: Get commit short hash
         run: node ./scripts/get-commit-short-hash.mjs
       - name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -42,7 +42,6 @@ jobs:
       - name: Get commit short hash
         run: node ./scripts/get-commit-short-hash.mjs
       - name: Build
-        working-directory: ./packages/cli/core
```
louis-bompart, line 45: ergo, will run from the root of the repo, thus triggering `nx` and the topological run

olamothe (approved):

PR #928, opened 2022-09-13
chore: sign macos install kit

y-lakhdar:
## Proposed changes

Sign, notarize and staple both MacOs (ARM64 and X64) packages.

### Additional change
Deleted `@angular/cli` since it contains a dependency (`fsevents`) with unsigned binary code, which was preventing us from signing the CLI.
Turns out `@angular/cli` is no longer needed, so there is no loss here.

## Testing

- [x] Unit Tests:
Not applicable
- [x] Functionnal Tests:
Not applicable
- [x] Manual Tests:
Examples of signed packages can be found in this [Artifact](https://github.com/coveo/cli/actions/runs/3596694972#artifacts)


louis-bompart (approved):
LGTM, great job, bravo for carrying this thru 👏 

Nit-pic, wrap the "Doing thing" with https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines

PR #929, opened 2022-09-13
chore: migrate to independent release

louis-bompart:
## Proposed changes

Modify the release process to be able to do independent releases.
This also means that package version will not be 'synchronized' anymore: So in the future, the default version of `create-atomic` for the `cli` might not be the version of the `cli`.

### How we do merge for changes that impact multiple projects:
- For now: If your changes impact multiple packages, you have to organize your commits so that one commit affects a single workspace every time.
- Soon ™️ : The PR description will be changed to include a new section that will be used by semantic release parser to make 'one commit nth commits'. This will be done in another PR

For example, if I do a breaking change in `@coveo/cli-commons`, I will need to change how `@coveo/cli` use the latter.
I will need to organize my branch in two commits:
- The first one will contain only the change relative to `@coveo/cli-commons`. Its message will contain the breaking change identifier. 
- The second one will contain only the change relative to `@coveo/cli`. Its message will not contain the breaking change container, it will be a chore or feat depending on the impact on the user. (e.g. if the change is not impacting the end user of the CLI, then it's a chore. If the change adds new behavior for the end user, it's a feat)

For this reason, I enabled the rebase merge option on the repository. It will be the one to use for the scenario described above.
### Tags format

- A tag for each release process, named `release-n`. `release-0` has been placed manually on the latest release.
- A tag for each released package, named `@coveo/package-name@version`.

### GitHub release
- Will occur only when the CLI is released.

## Testing

- [x] Manual Tests: I created a test branch locally where I disabled git push and npm publish and tested the process.


scripts/releaseV2/phase1-bump-all-packages.mjs
```diff
@@ -18,37 +20,44 @@ const rootFolder = resolve(dirname(fileURLToPath(import.meta.url)), '..', '..');
...
-  const versionPrefix = 'v';
+  const packageJson = JSON.parse(
+    readFileSync('package.json', {encoding: 'utf-8'})
+  );
+  const versionPrefix = packageJson.name;
   const convention = await angularChangelogConvention;
-  const lastTag = await getLastTag(versionPrefix);
+  // TODO: CDX-1147 Remove catch
+  const lastTag = await getLastTag(versionPrefix).catch(() =>
+    getLastTag('release-')
```
louis-bompart, line 30: Just for the first: because there's no version with the prefix yet.

ThibodeauJF:
> How we do merge for changes that impact multiple projects:

This will be necessary for minor changes as well or only breaking changes?

louis-bompart:
> > How we do merge for changes that impact multiple projects:
> 
> This will be necessary for minor changes as well or only breaking changes?

It will be necessary when 2+ public projects are impacted and than the change level in those is different

packages/cli/source/src/commands/source/push/delete.spec.ts
```diff
@@ -60,7 +60,7 @@ describe('source:push:delete', () => {
     .stdout()
     .stderr()
     .command(['source:push:delete', 'mysource', '-x', 'foo', '-d', 'bar'])
-    .catch(/--delete= cannot also be provided when using --deleteOlderThan=/)
+    .catch(/--delete=foo cannot also be provided when using --deleteOlderThan/)
```
louis-bompart, line 63: Small fix related to oclif update

packages/cli/core/src/commands/org/resources/pull.ts
```diff
@@ -75,17 +75,15 @@ export default class Pull extends CLICommand {
       description: 'Overwrite resources directory if it exists.',
       default: false,
     }),
-    resourceTypes: Flags.build<ResourceSnapshotType>({
-      parse: async (resourceType: ResourceSnapshotType) => resourceType,
-    })({
+    resourceTypes: Flags.enum<ResourceSnapshotType>({
```
louis-bompart, line 78: `Flags.build` has been deprecated, here `enum` seems the most adapted.

packages/cli/core/src/commands/org/resources/pull.ts
```diff
@@ -75,17 +75,15 @@ export default class Pull extends CLICommand {
...
       char: 'r',
       helpValue: 'type1 type2',
       description: 'The resources types to pull from the organization.',
       multiple: true,
       options: allowedResourceType,
       default: allowedResourceType,
     }),
-    model: Flags.build<SnapshotPullModel>({
+    model: Flags.custom<SnapshotPullModel>({
```
louis-bompart, line 86: `Flags.build` has been deprecated. Flags.custom it is instead

packages/angular/package.json
```diff
@@ -29,7 +29,7 @@
     "@coveo/search-token-server": "1.35.12",
     "@schematics/angular": "^14.0.1",
     "jest-config": "^27.4.7",
-    "typescript": "^4.3.5"
+    "typescript": "4.8.3"
```
louis-bompart, line 32: `@schematics/angular` is sneaky: it bundles typescript. If the version bundled and ours does not work nicely together 💥

packages/cli/core/src/commands/org/resources/pull.ts
y-lakhdar: not sure why this is past of this PR

y-lakhdar (approved):

packages/cli/core/src/commands/org/resources/pull.ts
louis-bompart: My own comments on the flags are outdated, I did fix that in another PR in the end. (that's why it's marked as outdated)

PR #931, opened 2022-09-15
fix: update ux symbol

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-1152

## Proposed changes

Fix the condition that determine the spinner symbol at the command termination.
Since anything can bubble up to the finally method, it's better to simply check if the error is undefined or not.

Example:
* A command that executes an HTTPs call.
* Axios returns an error (as an JSON object)
* The JSON object bubbles up to the base class
* The finally prints a success symbol because the JSON object is not an instance of Error.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


louis-bompart (approved):

louis-bompart:
How could it be tested 🤔

y-lakhdar:
> How could it be tested 🤔

Manual test:
Create a Snapshot with a key missing the resources privileges

PR #944, opened 2022-09-21
docs(search-token-server): add note about csurf

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1161

## Proposed changes

The only 'real' alternative for not being flagged would be to drop express, and given the nature of the project & vulnerability, I don't think that is necessary.

See https://snyk.io/blog/explaining-the-csurf-vulnerability-csrf-attacks-on-all-versions/


packages/ui/search-token-server/app.ts
```diff
@@ -1,6 +1,10 @@
 import express from 'express';
 import cors from 'cors';
 import cookieSession from 'cookie-session';
+/**
```
olamothe, line 4: ... meh ? 

I see that as a user, what am I supposed to think about that comment ? Why is this even included in the first place in this package ?

The comment you added in this PR is that "given the nature of the project", it's not that important. 

So if it's not important why are we even adding different security packages to try and... kinda make the server secure ?

Lets pick a side here:

1- We want to try and teach people how to create a production ready, secure (search token) server
2- We want to provide an example how to request and provide search token to an app.

I'm personally firmly in camp #2. That's what I've seen people asking question about when they want to implement Coveo, not #1.  #1 is not our business.

So let's not do an half assed job with #1.

louis-bompart: Then we have to do a whole new server.
There's no more crsf package maintained for express

y-lakhdar: I agree with #2. We either keep it and remove the comment or simply remove it keep cors (which is sufficient for a demo server IMO)

louis-bompart:
Because it's an example, no need to waste time

PR #947, opened 2022-09-22
tests(e2e): add package fallback to npm

louis-bompart:
## Proposed changes

Now that we release a package only if it changes, we can end up in a situation where a package is not published on verdaccio, which in turn causes the E2E that needs this package to fail. Oups.

This is actually working as intended; eventually, we want to not run the tests if no changes occur.
But we're not there yet, so in the meantime, if a package is not published, I'll pull it up from npm and push it back on verdaccio.

## Testing

CI

### DIFF
--- a/packages/cli-e2e/setup/utils/verdaccio.ts
+++ b/packages/cli-e2e/setup/utils/verdaccio.ts
@@ -45,7 +45,15 @@ export async function publishPackages() {
     publishTerminal.orchestrator.process.stderr.on('data', (data) => {
       console.log(data.toString());
     });
-    await publishTerminal.when('exit').on('process').do().once();
+    await publishTerminal
+      .when('exit')
+      .on('process')
+      .do((process) => {
+        if (process.exitCode) {
+          throw new Error('Error while publishing packages');
+        }
+      })
+      .once();
   }
 }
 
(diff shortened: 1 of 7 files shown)


packages/cli-e2e/setup/utils/verdaccio.ts
```diff
@@ -0,0 +1,120 @@
...
+      .when('exit')
+      .on('process')
+      .do((process) => {
+        if (process.exitCode) {
+          throw new Error('Error while publishing packages');
+        }
+      })
+      .once();
+  }
+}
```
louis-bompart, line 59: refactored from utils

packages/cli-e2e/setup/utils/verdaccio.ts
```diff
@@ -0,0 +1,120 @@
...
+    spawnSync(
+      appendCmdIfWindows`npm`,
+      ['publish', `--registry=${verdaccioRegistry}`],
+      {
+        cwd: tmpdir,
+        stdio: 'inherit',
+      }
+    );
+  }
+}
```
louis-bompart, line 120: refactored from utils

olamothe (approved):

PR #949, opened 2022-09-22
feat(cli/cli-commons): remove message from stopSpinner

louis-bompart:
## Proposed changes

- Remove `message` from the options of `stopSpinner`
- Remove the only usage
- Adapt E2E
- Fix the FIXME: `fancyIt` should not be used if no commands is instanciated. Instead, stderr-stdout should be used directly.

Motivation: From a UX perspective, the spinner in the CLI is used for 'Doing X, please wait'.
We don't need to say 'We've done X' after that, just to communicate the status of the operation. ✔️ and ❌ are perfect for that, so let's just use that and exclusively that.

## Breaking changes

Removed `message` from the options of `stopSpinner`.
It's technically a breaking change, sure we can consider that `@coveo/cli-commons` is so early that no one would be impacted. What I see here is the perfect excuse to test that the independent release works nice and proper ;)

## Testing

- [x] Unit Tests: adapted
- [x] Functional Tests: adapted

-----

⚠️  Should be merged with rebase strategy. ⚠️

### DIFF
--- a/packages/cli-e2e/__tests__/orgResources.specs.ts
+++ b/packages/cli-e2e/__tests__/orgResources.specs.ts
@@ -114,18 +114,21 @@ describe('org:resources', () => {
       debugName
     );
 
-    const pullTerminalExitPromise = pullTerminal
+    const pullTerminalExitPromise = async () => {
       // TODO: CDX-744: understand why cannot use process.on('exit')
-      .when(/Project updated/)
-      .on('stderr')
-      .do()
-      .once();
+      await pullTerminal
+        .when(/Updating project with Snapshot/)
+        .on('stderr')
+        .do()
+        .once();
+      await pullTerminal.when(/✔/).on('stderr').do().once();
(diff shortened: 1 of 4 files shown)


y-lakhdar (changes requested):
It's only working because of the test order. 
Need to consume first stderr before running the tests.

packages/cli/commons/package.json
```diff
@@ -17,6 +17,7 @@
     "@coveo/cli-commons-dev": "1.35.13",
     "@types/jest": "29.0.3",
     "jest": "29.0.3",
+    "jest-mock-process": "^2.0.0",
```
y-lakhdar, line 20: Where is this used?

louis-bompart: - [ ] uninstall `jest-mock-process`

y-lakhdar (approved):

PR #953, opened 2022-09-23
feat: handle missing domain privileges

y-lakhdar:
## Proposed changes

Check if the user has the right access to the resource he/she is about to manipulate via snapshot operations.
Large PR, but mainly .spec and .snap files.

## Before Merging
- [x] Merge https://github.com/coveo/platform-client/pull/521
- [x] Bump platform-client package

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/core/src/commands/org/resources/preview.spec.ts
```diff
@@ -159,15 +180,38 @@ describe('org:resources:preview', () => {
...
     });
 
-    afterAll(() => {
-      mockedSnapshotFactory.mockReset();
+    test
+      .stdout()
+      .stderr()
+      .command(['org:resources:preview'])
+      .catch((ctx) => {
+        expect(formatCliLog(ctx.message, true)).toMatchSnapshot();
```
louis-bompart, line 196: Why do you strip the ANSI esc sequences? It's often colour stuff, so it might be useful to know its been properly colored, no?

louis-bompart (commented):
Not sure why we don't snapshot the output with the ANSI, I think it would make sense to capture them too, because they're part of the tested behavior in fine

packages/cli/core/src/commands/org/resources/preview.spec.ts
y-lakhdar: That's a good point. It was to make the snapshot cleaner. but I guess we want to validate the styling as well

packages/cli/core/src/lib/snapshot/snapshot.spec.ts
```diff
@@ -2,6 +2,7 @@ import {
   ResourceSnapshotsModel,
```
louis-bompart, line 2: General comment on the change in this spec, relative to the change in its associated source file.

I think the change in these specs is testing the mocks, but not the actual change in the source code.
Indeed, if you read it out, you'll see that you mock a method and make the mock throw an error, and then test that the caller of this method does throw.

So, in essence, here we are asserting that the method is called by testing its mocked behavior.

This could be simplified by only testing what's actually happening in this method from a 'sandboxed' perspective:
Do I call the method, and do I call it with the right parameters?

Because the method is properly unit-tested, this should offer proper coverage and simplify those tests.

packages/cli/core/src/lib/snapshot/snapshotAccess.ts
```diff
@@ -0,0 +1,39 @@
...
+  ResourceSnapshotType,
+  SnapshotAccessType,
+} from '@coveord/platform-client';
+import {
+  MissingResourcePrivileges,
+  MissingSnapshotPrivilege,
+} from '../errors/snapshotErrors';
+import {isSubset, without} from '../utils/list';
+
+export async function ensureResourceAccess(
```
louis-bompart, line 12: nitpic but imo:
```suggestion
export async function ensureResourcesAccess(
```

packages/cli/core/src/lib/snapshot/snapshotFactory.spec.ts
```diff
@@ -1,15 +1,25 @@
 jest.mock('@coveo/cli-commons/platform/authenticatedClient');
 jest.mock('fs');
 jest.mock('./snapshot');
+jest.mock('./snapshotAccess');
```
louis-bompart, line 4: Same comment as on `snapshot.spec.ts`

packages/cli/core/src/lib/utils/list.ts
```diff
@@ -1,3 +1,7 @@
 export function without<T>(array: T[], values: T[]): T[] {
   return array.filter((field) => !values.includes(field));
 }
+
+export function isSubset<T>(subset: T[], superset: T[]): boolean {
+  return without(subset, superset).length === 0;
```
louis-bompart, line 6: This is 'nice' from a DRY standpoint but could be more efficient.
Indeed this requires going through the whole `subset` element.

Currently, the method needs to do at least `n+1` comparisons, with `n` being the number of elements in the `subset`.
Indeed, we don't have a 'quick-exit' condition if the first element (or second, etc...) of `subset` is actually not in `superset`

So to fix that:

```suggestion
  return subset.every(element => superset.includes(element))
```

With this, at best, you have 1 operation to do: "is the first element of subset in superset? No, ok return false"

Given that your method requires subset to be the first param, you can also ask a precondition by checking the length of the two arrays:

```suggestion
  return subset.length > superset.length ? false : subset.every(element => superset.includes(element));
```

packages/cli/source/package.json
```diff
@@ -16,7 +16,7 @@
   "dependencies": {
     "@coveo/cli-commons": "1.36.0",
     "@coveo/push-api-client": "2.5.15",
-    "@coveord/platform-client": "34.13.0",
+    "@coveord/platform-client": "34.14.0",
```
louis-bompart, line 19: Watch out: With the current scope of the PR (feat) + this change, your merge commit would wrongly bump the minor of the package. (bump of deps alone are just patch)

The easiest way for your changes to not do that would be to have the dep bump on another PR with the chore or fix scope.

(This apply to all the other packages 👇 )

louis-bompart (changes requested):
Src code lgtm 👍 
Tests could be simplified.
Nitty-gritty opti about set theory.
⚠️ about PR that affects a lot of packages.

packages/cli/core/src/lib/utils/list.ts
louis-bompart: > If we go that path, might as well compare real sets:
> 
> ```
>   if (new Set(subset).size > new Set(superset).size) {
>     return false;
>   }
>   return subset.every((element) => superset.includes(element));
> ```

I thought about it, but then I checked how the ctor of a Set work:
https://tc39.es/ecma262/multipage/keyed-collections.html#sec-set-iterable

You can see that, well, it will iterate through all values, so it would be costly to convert that as a set.

However, this made me think: the `if(subset.size>superset.size) return false` would not work: it assumes that every element of the subset and superset are unique, which we cannot know for sure:
```ts
const subset = [0,0,0,0,0,0,0,0]
const superset = [0,1,2,3]
// oops
```

The `subset.every(element => superset.includes(element))` is still OK tho.

packages/cli/core/src/lib/snapshot/snapshot.spec.ts
y-lakhdar: I see what you are saying, but should the method checks at least if it resolves or rejects based on the resource access precondition (regardless of the error thrown)?

louis-bompart: I think this is not necessary because we don't have error handling logic in this method: we just bubble up the error up the stack here.

louis-bompart (approved):
lgtm

PR #958, opened 2022-09-28
docs(cli): fix bad link

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1188

## Proposed changes

The link was bad, but now it's not. Readme will be updated in the next build automatically.


jpmarceau (approved):
Thanks!

PR #961, opened 2022-09-28
feat(cli): improve auth:login final feedback

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1191

## Proposed changes

Mention `coveo config:set -o` mostly

### DIFF
--- a/packages/cli/core/src/commands/auth/__snapshots__/login.spec.ts.snap
+++ b/packages/cli/core/src/commands/auth/__snapshots__/login.spec.ts.snap
@@ -24,7 +24,7 @@ exports[`auth:login succeeds when organization flag is valid 1`] = `
     Region: us
     Environment: prod
     
-To change organization either:
+To change organization, either:
   * Run coveo auth:login --help to see the available options to log in to a different organization, region, or environment.
   * Run coveo config:set -o=theOtherOrg to run the next commands against theOtherOrg.
 "
--- a/packages/cli/core/src/commands/auth/__snapshots__/login.spec.ts.snap
+++ b/packages/cli/core/src/commands/auth/__snapshots__/login.spec.ts.snap
@@ -1,0 +1,32 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`auth:login succeed when no organization flag is passed, and uses the first available org instead 1`] = `
+"
+    Successfully logged in!
+    Close your browser to continue.
+
+    You are currently logged in:
+    Organization: the_first_org_available
+    Region: us
+    Environment: prod
+    
+To changes organizations either:
+  * Run coveo auth:login --help to see the available options to log in to a different organization, region, or environment.
+"
+`;
+
+exports[`auth:login succeed when organization flag is valid 1`] = `
+"
+    Successfully logged in!
+    Close your browser to continue.
+
+    You are currently logged in:
+    Organization: foo
+    Region: us
+    Environment: prod
+    
+To changes organizations either:
+  * Run coveo auth:login --help to see the available options to log in to a different organization, region, or environment.
+  * Run coveo config:set -o=theOtherOrg to run the next commands against theOtherOrg
+"
+`;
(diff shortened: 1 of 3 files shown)


packages/cli/core/src/commands/auth/login.ts
```diff
@@ -46,8 +47,18 @@ export default class Login extends CLICommand {
...
+    const hasOrgFlag = Boolean((await this.parse(Login)).flags.organization);
+    this.log(
+      dedent`
+    To changes organizations either:
+      * Run coveo auth:login --help to see the available options to log in to a different organization, region, or environment.
+      ${
+        hasOrgFlag
+          ? '* Run coveo config:set -o=theOtherOrg to run the next commands against theOtherOrg'
+          : ''
+      }`.trimEnd()
```
y-lakhdar, line 60: Nitpick, But can we have the bullet list only if `hasOrgFlag` is set to true.
Otherwise it's a little bit weird to have this message:
```
To changes organizations either:
  * Run coveo auth:login --help to see the available options to log in to a different organization, region, or environment.
```

y-lakhdar: And also, to keep consistent with other lists in the CLI, let's use the same symbol. I think we used bull character.

y-lakhdar (changes requested):

packages/cli/core/src/commands/auth/__snapshots__/login.spec.ts.snap
```diff
@@ -0,0 +1,32 @@
...
+    Successfully logged in!
+    Close your browser to continue.
+
+    You are currently logged in:
+    Organization: the_first_org_available
+    Region: us
+    Environment: prod
+    
+To changes organizations either:
+  * Run coveo auth:login --help to see the available options to log in to a different organization, region, or environment.
```
jpmarceau, line 14: I don't really understand here: the `either` seems to imply that there are two options, but I only see one?

I'm guessing that the content of this file gets generated from the one below? I will suggest something in that file, but let me know if I'm wrong and I'll suggest something here too 👍

packages/cli/core/src/commands/auth/login.ts
```diff
@@ -46,8 +47,18 @@ export default class Login extends CLICommand {
...
-    Run auth:login --help to see the available options to log in to a different organization, region, or environment.
     `);
+    const hasOrgFlag = Boolean((await this.parse(Login)).flags.organization);
+    this.log(
+      dedent`
+    To changes organizations either:
+      * Run coveo auth:login --help to see the available options to log in to a different organization, region, or environment.
+      ${
+        hasOrgFlag
+          ? '* Run coveo config:set -o=theOtherOrg to run the next commands against theOtherOrg'
```
jpmarceau, line 58:
```suggestion
          ? '* Run coveo config:set -o=theOtherOrg to run the next commands against theOtherOrg.'
```

packages/cli/core/src/commands/auth/login.ts
```diff
@@ -46,8 +47,22 @@ export default class Login extends CLICommand {
...
+    this.log(this.getHowToChangeOrgMessage(hasOrgFlag));
+  }
+
+  private getHowToChangeOrgMessage(hasOrgFlag: boolean) {
+    const runAuthLoginMessage =
+      'Run coveo auth:login --help to see the available options to log in to a different organization, region, or environment.';
+
+    return hasOrgFlag
+      ? dedent`
+        To change organization either:
```
jpmarceau, line 61:
```suggestion
        To change organization, either:
```

packages/cli/core/src/commands/auth/__snapshots__/login.spec.ts.snap
```diff
@@ -0,0 +1,31 @@
...
+"
+    Successfully logged in!
+    Close your browser to continue.
+
+    You are currently logged in:
+    Organization: foo
+    Region: us
+    Environment: prod
+    
+To change organization either:
```
jpmarceau, line 27:
```suggestion
To change organization, either:
```

jpmarceau (approved):
👍

louis-bompart:
@y-lakhdar I'll need your review here ;)

y-lakhdar (approved):

PR #963, opened 2022-09-29
feat(cli): ensure user is webauth-ed for ui cmds

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1992

## Proposed changes

Add a precondition to ensure that the user authenticated using the OAuth flow and not an API Key for the `ui:create` commands.

## Testing

- [x] Unit Tests


y-lakhdar (approved):

packages/cli/commons/src/preconditions/isWebAuthenticated.ts
```diff
@@ -0,0 +1,17 @@
...
+import {PreconditionError} from '../errors/preconditionError';
+
+const PRECONDITION_ERROR_CATEGORY = 'Wrong Authentication Type';
+
+export function IsWebAuthenticated() {
+  return async function () {
+    const config = new Config(globalConfig.get().configDir).get();
+    if (config.anonymous) {
+      throw new PreconditionError(
+        'This command does not support token authentication. Please authenticate again using coveo auth:login',
```
olamothe, line 12: `token authentication` is not specific enough.

```suggestion
        'This command does not support API key authentication. Please authenticate again using coveo auth:login',
```

packages/cli/commons/src/preconditions/isWebAuthenticated.ts
```diff
@@ -0,0 +1,17 @@
+import {Config} from '../config/config';
+import globalConfig from '../config/globalConfig';
+import {PreconditionError} from '../errors/preconditionError';
+
+const PRECONDITION_ERROR_CATEGORY = 'Wrong Authentication Type';
+
+export function IsWebAuthenticated() {
```
olamothe, line 7: What about adding parameters to the `IsAuthenticated()` precondition ?

Say, an array of supported authentication scheme.

By default it could be an empty array (or an array with all schemes).

This just seems weird to me: 

```
@Preconditions(
    IsAuthenticated(),
    IsWebAuthenticated(),
    ....
)
```

vs: 
```
@Preconditions(
    IsAuthenticated(['OAuth']),
    ....
)
```

olamothe (commented):
CDX-1992 does not seem like a valid ticket. I'm lacking some context as to why this is needed.

packages/cli/commons/src/preconditions/isWebAuthenticated.ts
louis-bompart: Mh yeah we could do that too

louis-bompart: Done ✔️

packages/cli/commons/src/preconditions/isWebAuthenticated.ts
louis-bompart: Done

louis-bompart:
Back to draft, big refactor/tests clean up I'll need to sort out from this branch

packages/cli/source/src/commands/source/push/add.spec.ts
```diff
@@ -1,126 +1,102 @@
...
   const doMockErrorBatchUpload = () => {
     mockBatchUpdate.mockReturnValue(new BatchUploadDocumentsError());
   };
+  const doMockAuthenticatedClient = () => {
+    mockedClient.mockImplementation(
+      () =>
+        ({
+          cfg: {
+            get: () => ({
+              accessToken: 'the_token',
```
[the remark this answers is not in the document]
louis-bompart: Oh shush.

packages/cli-e2e/utils/login.ts
```diff
@@ -79,7 +79,7 @@ export function runLoginCommand(orgId: string, env: string) {
     .do(answerPrompt('n'))
     .once();
 
-  return loginTerminal;
+  return loginTerminal.when('exit').on('process').do().once();
```
louis-bompart, line 82: Funny thing: the E2E Login process was existing just after setting `accessToken` and was not waiting for the other configs to be set, notably the orgId.

PR #977, opened 2022-10-13
fix: ensure flags are applied to `org:resources:list` command

y-lakhdar:
## Proposed changes

Flags related to List command are not applying when running the command.

 
 
## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [x] Manual Tests:


packages/cli/core/src/commands/org/resources/__snapshots__/list.spec.ts.snap
```diff
@@ -0,0 +1,26 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
y-lakhdar, line 1: I moved everything into the snap file to make the spec file lighter.

louis-bompart: Nice

louis-bompart (approved):
lgtm, quick and right on target 👍

PR #978, opened 2022-10-13
chore: remove extra spinner

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-1209

## Proposed changes
Remove the spinner check at each step of a snapshot application. It becomes too verbose and can print the same string multiple times. This happens because we ensure the user access on the snapshot validate and snapshot apply.

If any of these access check returns an error, it will bubble up anyway

Before: 
```
Creating Snapshot... ✔
Validating resource access... ✔
Validating snapshot... ✔
Validating snapshot access... ✔
Applying snapshot... ✔
Validating snapshot access... ✔
```

After:
```
Creating Snapshot... ✔
Validating snapshot... ✔
Applying snapshot... ✔
```

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


louis-bompart (approved):
LGTM

PR #979, opened 2022-10-13
chore: fix release script

louis-bompart:
## Proposed changes

 - Target the proper tag for the release download on Jenkins.


scripts/get-release-artifact-and-commit-sha.mjs
```diff
@@ -1,9 +1,16 @@
-const {downloadReleaseAssets, getLatestTag} = require('./github-client');
-const fs = require('fs');
-const path = require('path');
+import {downloadReleaseAssets} from './github-client';
```
louis-bompart, line 1: Migrated to ESM because semantic-monorepo-tools is ESM only

y-lakhdar (approved):

PR #982, opened 2022-10-14
chore(release): better prefix

louis-bompart:
CDX-764

## Proposed changes

This is required to make sure:
 - `@coveo/cli` does not resolve on `@coveo/cli-commons` etc
 - Somehow it fixes angular/vue/react. No clue why if I'm honest.


y-lakhdar (approved):

louis-bompart:
Before:
https://github.com/coveo/cli/actions/runs/3252065746/jobs/5337854943
After:
https://github.com/coveo/cli/actions/runs/3252075513/jobs/5337875784

See that angular and cie are properly acquired.

PR #993, opened 2022-10-19
feat: update to node 18 

louis-bompart:
Update the CLI to make it compatible with Node 18


packages/cli-e2e/jest.config.ts
```diff
@@ -7,7 +7,8 @@ export default {
   testEnvironment: 'node',
   globalSetup: process.env.CI ? './setup/ci.ts' : './setup/local.ts',
   globalTeardown: './teardown.ts',
-  testMatch: process.env.CI
-    ? ['**/__tests__/**/*specs?(.ci).ts']
-    : ['**/__tests__/**/*specs.ts'],
+  testMatch:
+    process.env.CI || true
+      ? ['**/__tests__/**/*specs?(.ci).ts']
+      : ['**/__tests__/**/*specs.ts'],
```
louis-bompart, line 13:
```suggestion
  testMatch:
  testMatch: process.env.CI
    ? ['**/__tests__/**/*specs?(.ci).ts']
    : ['**/__tests__/**/*specs.ts'],
```

packages/cli-e2e/package.json
```diff
@@ -61,7 +61,7 @@
     "@babel/core": "7.19.1",
     "@babel/preset-env": "7.19.1",
     "@babel/preset-typescript": "7.18.6",
-    "@coveo/cli": "1.37.1",
+    "@coveo/cli": "1.37.2",
```
louis-bompart, line 64: #992

.github/workflows/build.yml
```diff
@@ -135,13 +135,14 @@ jobs:
...
-        spec: [
+        node: [18]
+        spec:
+          [
             'angular.specs.ts',
             'atomic.specs.ts',
             'auth.specs.ci.ts',
             'orgList.specs.ts',
-            # 'orgResources.specs.ts',
+            'orgResources.specs.ts',
```
louis-bompart, line 145: #996

packages/cli-e2e/utils/mitmproxy.ts
```diff
@@ -11,7 +11,16 @@ export const startMitmProxy = (
...
+    [
+      '-u',
+      mitmScript,
+      '-p',
+      '8080',
+      '--listen-host',
+      'localhost',
+      '--flow-detail',
+      '1',
+    ],
```
louis-bompart, line 23: More resilient: `-u` ensure `python` does not buffer the stderr/stdout, and the rest ensure we get enough deets

packages/cli-e2e/utils/browser.ts
```diff
@@ -32,9 +34,13 @@ export async function closeAllPages(browser: Browser) {
...
   const chromeDebugInfoRaw = (
-    await axios.get<JsonVersionFile>(CHROME_JSON_DEBUG_URL)
+    await axios.get<JsonVersionFile>(CHROME_JSON_DEBUG_URL, {
+      httpAgent: new HttpAgent({family: 4}),
+    })
   ).data;
-  return chromeDebugInfoRaw.webSocketDebuggerUrl;
+  const wsUrl = new URL(chromeDebugInfoRaw.webSocketDebuggerUrl);
+  wsUrl.hostname = (await dnsPromises.resolve4(wsUrl.hostname))[0];
+  return wsUrl.toString();
```
louis-bompart, line 43: Node 18 use ipv6 by default for pretty much everything, pupeeter doesn't like ipv6.

packages/cli/core/package.json
```diff
@@ -74,7 +75,7 @@
   },
   "peerDependencies": {
     "@angular/cli": "^14.0.1",
-    "@vue/cli": "^4.5.11 || ^5.0.0",
+    "@vue/cli": "^5.0.8",
```
louis-bompart, line 78: `@vue/cli@4.x` has a dep or transitive dep that is not compatible with Node 18.

packages/cli/core/src/lib/snapshot/snapshotFactory.ts
```diff
@@ -21,14 +22,15 @@ export class SnapshotFactory {
...
+    });
     const computedOptions: CreateFromFileOptions = {
       developerNotes: 'cli-created-from-zip',
     };
 
-    const model = await client.resourceSnapshot.createFromBuffer(
-      file,
-      ResourceSnapshotSupportedFileTypes.ZIP,
+    const model = await client.resourceSnapshot.createFromFile(
+      blob,
```
louis-bompart, line 33: Byeeee special Node hacked method, now we have blob and proper fetch that support it.

scripts/get-tags.mjs
```diff
@@ -18,7 +17,7 @@ const getReleaseTag = async () => {
     }
   }
   warning('Tag acquisition failed, fallbacking to the latest');
-  return getLastTag(releaseTagPrefix);
+  return getLastTag(cliTagPrefix);
```
louis-bompart, line 20: Not sure.
- [ ] uh?

packages/cli-e2e/__tests__/orgResources.specs.ts
```diff
@@ -210,8 +211,10 @@ describe('org:resources', () => {
     const getResourceFolderContent = (projectPath: string) =>
       readdirSync(join(projectPath, 'resources'));
 
-    beforeEach(() => {
-      rmSync(destinationPath, {recursive: true, force: true});
+    beforeEach(async () => {
+      await retry(() => {
+        rmSync(destinationPath, {recursive: true, force: true});
```
y-lakhdar, line 216: Why retry here

y-lakhdar (approved):

packages/cli-e2e/__tests__/orgResources.specs.ts
louis-bompart: On Windows, the directory is still flagged as "EBUSY" after the CLI execution.
This could be due to either the slowness of NTFS compared to extFS, or more likely that we don't kill the CLI properly on Windows (`fkill` works nice, but well windows can still be a pain sometimes).

So, because this is a 'setup' phase, I feel comfortable putting a retry if it increases the reliability. I would not have done so if it was part of the 'execute/verify' phases of our tests tho.

PR #1000, opened 2022-10-26
feat(vue)!: migrate to vue3

louis-bompart:
Vue3 PR, split into 3 commits:
 - 1 commit to deprecate vue2 plugin and flag vue3 new package for v1 release.
 - 1 commit for the CLI changes
 - 1 commit for e2e and cleaning

PR #1001, opened 2022-10-26
Vue3 ut

louis-bompart:
## Proposed changes

Unit test for new vue stuff.


packages/cli/core/src/commands/ui/create/vue.spec.ts
```diff
@@ -98,13 +118,32 @@ describe('ui:create:vue', () => {
     );
   };
 
+  const doMockNodeProcess = () => {
+    jest.mock('node:process', () => ({
```
y-lakhdar, line 122: I don't think it makes a difference to mock `cwd` at `beforeEach` since Jest will automatically hoist `jest. mock` calls to the top of the module

packages/cli/core/src/commands/ui/create/vue.spec.ts
```diff
@@ -146,43 +185,120 @@ describe('ui:create:vue', () => {
...
+            expect.stringMatching(fooBarDirectoryMatcher),
+            {
+              withFileTypes: true,
+            }
+          );
+          expect(mockedSpawnProcess).toBeCalledTimes(2);
+          expect(mockedSpawnProcess.mock.calls[0]).toMatchSnapshot();
+          expect(mockedSpawnProcess.mock.calls[1]).toMatchSnapshot();
+          expect(mockedWriteFileSync).toBeCalledTimes(1);
+          expect(mockedWriteFileSync.mock.calls[0]).toMatchSnapshot();
```
y-lakhdar, line 274: Seems like there are too many expects in the same test.

packages/cli/core/src/commands/ui/create/__snapshots__/vue.spec.ts.snap
```diff
@@ -0,0 +1,75 @@
...
+  "VITE_APP_PLATFORM_URL=https://platformdev.cloud.coveo.com
+VITE_APP_ORGANIZATION_ID=my-org
+VITE_APP_API_KEY=foo
+VITE_APP_USER_EMAIL=bob@coveo.com",
+]
+`;
+
+exports[`ui:create:vue when a directory with the projectName exists when the directory is not empty should exit with an error 1`] = `"/foo/bar/myapp is a non-empty directory, use an empty directory or unused name for your project"`;
+
+exports[`ui:create:vue when a file with the projectName exists should exit with an error 1`] = `"/foo/bar/myapp already exists and is not a directory, use an unused directory name for your project"`;
```
y-lakhdar, line 40: Are these error codes coming from Vite?

packages/cli/core/src/commands/ui/create/__snapshots__/vue.spec.ts.snap
```diff
@@ -0,0 +1,75 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ui:create:vue when a directory with the projectName exists when the directory is empty should use the dir and init the project 1`] = `
+[
+  "npm",
+  [
+    "init",
+    "@coveo/headless-vue@1.0.0",
```
y-lakhdar, line 8: I think you can use [Jest  property matcher](https://jestjs.io/docs/snapshot-testing#property-matchers) to not hard code the version in the snapshot

packages/cli/core/src/commands/ui/create/__snapshots__/vue.spec.ts.snap
louis-bompart: Nope, from #1000 : https://github.com/coveo/cli/blob/1a6856624558fcef273c54cb774e9ca9f59d32bd/packages/cli/core/src/commands/ui/create/vue.ts#L99-L115

Vite has nothing to do in the UT ;)

y-lakhdar: My bad. I did not search in the right branch 🤦

packages/cli/core/src/commands/ui/create/__snapshots__/vue.spec.ts.snap
louis-bompart: Ya right, but well, it become hen easier to proceed with a expect.tobeNthcalledwith. so that's why I did

packages/cli/core/src/commands/ui/create/vue.spec.ts
louis-bompart: Split

PR #1009, opened 2022-11-02
test(e2e): wake up org before daily tests/cleaning

louis-bompart:
## Proposed changes

Call resume on the org to ensure the tests don't fail because the org is sleepy.
We will maybe improve that later.


y-lakhdar (approved):
image

PR #1024, opened 2022-11-10
docs(core): fix links

louis-bompart:
## Proposed changes

We are using `@coveo/cli@X.Y.Z` for tags. not `vX.Y.Z`.

This fix the readme for now, will fix the gen on another PR

## Testing

Just click a link.


jpmarceau (approved):
Thanks Louis 👍

PR #1029, opened 2022-11-15
fix(cli): support prerelease for ui package version resolution

louis-bompart:
## Proposed changes
- Replace `semver.coerce` by `semver.valid`. `coerce` 'strip' the prerelease, 'valid' doesn't.

## Testing

- [x] Unit Tests


packages/cli/core/src/lib/utils/misc.spec.ts
```diff
@@ -1,4 +1,3 @@
-import {fancyIt} from '@coveo/cli-commons-dev/testUtils/it';
```
louis-bompart, line 1: Not required for utility functions that doesn't interact with the stdio

PR #988, opened 2022-10-17
chore(release): fix jenkins artifacting

louis-bompart:
## Proposed changes

- Fix ESM import (need file-ext)
- Recover tag from release (simpler)
- Infer file architecture from filename instead of 'redoing' oclif logic

## Testing

- [x] Manual Tests: Created a PAT and run it on my machine, checked the artifacts folder.

PR #1031, opened 2022-11-15
fix(cli): fix sha acquisiton for windows signing

louis-bompart:
## Proposed changes
Similar problem and similar fix as #988

## Testing

- [x] Manual Tests: Replicated file tree and ran the script. Assessed the hash was there and complete

### DIFF
--- a/scripts/get-commit-short-hash.mjs
+++ b/scripts/get-commit-short-hash.mjs
@@ -13,7 +13,6 @@ const someExe = artifacts.find(
 if (!binariesMatcher.exec(someExe.name)?.groups?.commitSHA) {
   setFailed('Binary does not match');
 }
-console.log(binariesMatcher.exec(someExe.name).groups.commitSHA);
 exportVariable(
   'commitSHA1',
   binariesMatcher.exec(someExe.name).groups.commitSHA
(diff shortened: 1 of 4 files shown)


scripts/get-commit-short-hash.mjs
```diff
@@ -1,14 +1,21 @@
...
-    const commitShortHash = await getCommitShortHash();
-    exportVariable('commitSHA1', commitShortHash);
-  } catch (error) {
-    setFailed(error.message);
-  }
-})();
+if (!binariesMatcher.exec(someExe.name)?.groups?.commitSHA) {
+  setFailed('Binary does not match');
+}
+console.log(binariesMatcher.exec(someExe.name).groups.commitSHA);
```
louis-bompart, line 16:
```suggestion
```

scripts/get-commit-short-hash.mjs
```diff
@@ -1,14 +1,20 @@
...
-  }
-})();
+if (!binariesMatcher.exec(someExe.name)?.groups?.commitSHA) {
+  setFailed('Binary does not match');
+}
+exportVariable(
+  'commitSHA1',
+  binariesMatcher.exec(someExe.name).groups.commitSHA
+);
+``;
```
louis-bompart, line 20:
```suggestion
```

PR #1033, opened 2022-11-16
fix(cli): fix readme generation

louis-bompart:
## Proposed changes

Add a script that modifies the readme after oclif.

## Testing

- [x] Manual Tests: I ran the script to modify the current readme.


y-lakhdar (approved):
Good catch

PR #1034, opened 2022-11-16
fix: abort pending requests

y-lakhdar:
https://coveord.atlassian.net/browse/CDX-1232

## Proposed changes

Abort pending requests to avoid exceeding the maximum amount of event listeners allowed by the process.
Otherwise,  the following error will occur:
```
(node:54396) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit
```

### TL;DR;
Too many requests with the same PlatformClient instance can cause memory leaks.

### Long explanation
Every new instance of PlatformClient will instantiate a new AbortController object. This controller is used by the Fetch API on every HTTPS requests to handle abort events from the DOM.

However, on every request, fetch adds a new abort listener to the same AbortSignal (signal associated to the AbortController). At a certain point, the limit of listener exceed the maximum authorized by the process and return MaxListenersExceededWarning warning.
To avoid this issue, we can ask for a new Abort Controller when we retry the same request. This is what the abortPendingGetRequests is doing. It aborts the existing requests and associates a new AbortController instance to the same PlatformClient instance

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/cli/core/src/lib/snapshot/snapshot.spec.ts
+++ b/packages/cli/core/src/lib/snapshot/snapshot.spec.ts
@@ -50,6 +50,7 @@ const mockedDeleteSnapshot = jest.fn();
 const mockedGetSnapshot = jest.fn();
 const mockedExportSnapshot = jest.fn();
 const mockedApplySnapshot = jest.fn();
+const mockedAbortPendingGetRequests = jest.fn();
 const mockedDryRunSnapshot = jest.fn();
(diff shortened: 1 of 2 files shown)


louis-bompart (approved):
LGTM

PR #1037, opened 2022-11-21
fix(angular): preconditions command printing

louis-bompart:
## Proposed changes

⚠️ This PR will be rebase & merge (so commits messages on this branch do matter) ⚠️ 

- Refactored most `Command` (if not all) to `CLICommand`.
- Added a printable method on CLICommand.

`id` is set as a static prop on the classes that extends Command. So I created an instance-based helper on CLICommand to get it more easily.

## Breaking changes
- `@coveo/cli-commons-dev`: removed getFakeCommand. Now available in `@coveo/cli-commons`

## Testing

- Corner case, no auto tests.
- [x] Manual Tests: Ran `bin/dev ui:create:angular` without `ng` globally installed, assessed the results.


packages/cli/commons/src/command/cliCommand.ts
```diff
@@ -43,4 +43,8 @@ export abstract class CLICommand extends Command {
 
     return super.finally(error);
   }
+
+  public get printableName(): string {
```
olamothe, line 47:
```suggestion
  public get identifier(): string {
```

olamothe (approved):

y-lakhdar (approved):

PR #1044, opened 2022-11-23
feat(atomic): reduce disk footprint of projects

louis-bompart:
## Proposed changes

- Reduce project footprint from 1GB to 380MB
- Externalize `netlify-cli`: it is now a prerequisite for the command to run, same as `ng` is for `angular` for example.
- Leverage symlink to avoid file duplications. (instead of copying `lambda` we now use a symlink)
- Replace old polyfill by the OG fetch from nodejs (aka undici)
## Testing

- [x] Unit Tests: UT written
- [x] Functionnal Tests: FT adapted: we used a few trick to calculate the hash of the folder that required a few changes to work with the symlink. Ran twice to ensure snapshot is OK
- [x] Manual Tests: ran the ft, started verdaccio and run the CLI from the verdaccio release. It works, and file footprint is indeed reduced.


packages/ui/atomic/template/scripts/setup-lambda.mjs
```diff
@@ -0,0 +1,19 @@
...
+import {dirname, resolve} from 'node:path';
+import {fileURLToPath} from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+function linkSearchTokenLambdaToRoot() {
+  symlinkSync(
+    resolve(__dirname, '..', 'node_modules', '@coveo', 'search-token-lambda'),
+    resolve(__dirname, '..', 'lambda'),
+    'junction'
```
louis-bompart, line 11: This is the magic word that allows it to work on windows
See this SO if you're curious: https://superuser.com/a/1291446

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -150,7 +147,7 @@ describe('ui:create:atomic', () => {
       args.shift()!,
       args,
       {
-        cwd: normalizedProjectDir,
+        cwd: originalProjectDir,
```
louis-bompart, line 150: We start the project from the "og" directory because rollup is confused by nested junctions (which only occurs here)

louis-bompart:
Bonus side-effect: E2E are twice as fast for atomic

packages/cli/core/package.json
```diff
@@ -75,14 +75,14 @@
   },
   "peerDependencies": {
     "@angular/cli": "^15.0.0",
-    "@vue/cli": "^5.0.8",
```
ThibodeauJF, line 78: Is the vue cli not needed as a peer dep anymore?

ThibodeauJF (approved):

packages/cli/core/package.json
louis-bompart: Indeed, we stopped using it since the CLI v2, it's mostly just some cleaning

PR #1046, opened 2022-11-25
fix(cli): remove duplicate message

louis-bompart:
## Proposed changes

image
The 'what to do' is printed twice.
Also, because of the use of this.log, it does a similar issue as #1030 

## Testing

- [x] Manual Tests: Try to preview a snapshot of a big org

### DIFF
--- a/packages/cli/core/src/commands/org/resources/preview.ts
+++ b/packages/cli/core/src/commands/org/resources/preview.ts
@@ -1,8 +1,6 @@
 import {CLICommand} from '@coveo/cli-commons/command/cliCommand';
 import {Flags} from '@oclif/core';
-import {blueBright} from 'chalk';
 import {cwd} from 'process';
-import dedent from 'ts-dedent';
 import {Config} from '@coveo/cli-commons/config/config';
 import {
   HasNecessaryCoveoPrivileges,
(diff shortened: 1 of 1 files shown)


y-lakhdar:
Maybe it's because I skip the coffee this morning, but I don't see any error to be fixed here.
They are different "next steps"

louis-bompart:
> Maybe it's because I skipped the coffee this morning, but I don't see any error to be fixed here.
> They are different "next steps"

Mh you're right, but there should be only one next step I think.

Also regardless, it still break the spinner

y-lakhdar:
> there should be only one next step I think.
Agreed, but the PR is not fixing that. You would need to wait until the operation completes (or times out) , then, you would print the right "next step" message.

louis-bompart:
> > there should be only one next step I think.
> 
> Agreed, but the PR is not fixing that. You would need to wait until the operation completes (or times out) , then, you would print the right "next step" message.

I think we should mute the error and only display the log instead (after the log).
Will try to rework that ⌛

louis-bompart:
Closing: The error was occuring because of the expanded preview current behavior.
It's not that easily fixable so I wager on the chances of it happening for now.
True fix is the other preview methods we thought of

PR #1060, opened 2022-12-02
chore(cleaning): change prefix match for apiKey

louis-bompart:
## Proposed changes

Same as #1047 but for the API Keys


.github/workflows/delete-resources.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Resume org
         timeout-minutes: 30
         if: always()
-        run: node ./scripts/cleaning/wake-org.mjs
+        run: node ./scripts/cleaning/wake-org.mjs || true
```
y-lakhdar, line 29: Not sure I understand this one

y-lakhdar (approved):

PR #1063, opened 2022-12-05
feat(cli): add fallback port for oauth

### LINKED ISSUE #1061: Port 32111 is unavailable and no option to change for auth
## Describe the bug

Running `coveo auth:login` is blocked on clients running VMWare. The port 32111 is used by VMWare. There is no option in the Coveo CLI to change the port through an option/flag

https://github.com/coveo/cli/blob/master/packages/cli/core/src/lib/oauth/oauth.ts#L25
https://github.com/coveo/cli/blob/master/packages/cli/core/src/lib/oauth/oauth.ts#L32

## To Reproduce

1. Run something on port 32111
2. Try to run `coveo auth:login` from the CLI
3. See EACCES issue on port not available

## Expected behavior

I would expect to be able to set an alternative port through a `--port=12345` option in the command line

## Desktop (please complete the following information):

- OS: [e.g. macOS, Windows]
	Windows 10
- Version of the CLI [e.g. 1.0.1]
  -  [v2.1.0](https://github.com/coveo/cli/releases/tag/%40coveo%2Fcli%402.1.0)
- Local Node version: [e.g. 16.17.0]
- Local NPM/Yarn version: [e.g. 8.15.0]

louis-bompart:
Fixes #1061 

## Proposed changes

Add 52296 as an alternative port for OAuth. I choose 52296 because it is part of the private port range as defined by IANA, so I think the collision chance is lesser than in the system or user range. A caveat, this could be argued as a misuse because the port cannot be 'dynamically assigned' (i.e. the program can't change it on a whim). But I reckon the benefits outweigh this technicality.

## Testing

- [x] Unit Tests:
- [x] Manual Tests: reproducing steps of #1061


sherakama:
👍 🥇

PR #1097, opened 2023-02-07
fix(atomic): support non-us regions

louis-bompart:
## Proposed changes

- give the good url to @coveo/create-atomic package, it manages afterward

## Testing

- [x] Unit Tests: ye
- [ ] Functionnal Tests: nah, not worth testing on each region IMO
- [x] Manual Tests: created an org in EU, logged to it, ran the command from bin/dev. It works. Extra miles, npm started the project, it works too.

### DIFF
--- a/packages/cli/core/src/commands/ui/create/atomic.spec.ts
+++ b/packages/cli/core/src/commands/ui/create/atomic.spec.ts
@@ -77,6 +77,18 @@ describe('ui:create:atomic', () => {
       configurationMock({
         accessToken: 'foo',
         environment: 'dev',
+        region: 'us',
+        organization: 'my-org',
+      } as Configuration)
+    );
+  };
+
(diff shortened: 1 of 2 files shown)


packages/cli/core/src/commands/ui/create/atomic.spec.ts
```diff
@@ -77,6 +77,18 @@ describe('ui:create:atomic', () => {
       configurationMock({
         accessToken: 'foo',
         environment: 'dev',
+        region: 'us',
```
louis-bompart, line 80: we're reading it, so if we send undefined... oopsie

louis-bompart: note also that its a required property of `Configuration`, so it should be OK

ThibodeauJF (approved):

PR #1099, opened 2023-02-09
feat(deploy): add new (but hidden) deploy command

ThibodeauJF:
KIT-2268

## Proposed changes

[Design Doc](https://docs.google.com/document/d/1QMDA6nsfE9Bpfrpvh4Nj2jLSo0wbmY-tHOK0gT8t5yM/edit#heading=h.m9xhwxnzkkh)

Takes a JSON config, validates it, reads the necessary files and uploads to whole to the new service.

## Breaking changes

New!

## Testing
All automated forms of tests will come soon.

- [x] Manual Tests:

#### Creation
Screen Shot 2023-02-09 at 2 22 04 PM

Screen Shot 2023-02-09 at 2 22 52 PM

#### Update

Screen Shot 2023-02-09 at 2 47 55 PM

Screen Shot 2023-02-09 at 2 42 37 PM


packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (flags.pageId) {
+      this.updatePage({...hostedPage, id: flags.pageId});
+      return;
+    }
+
+    try {
+      await this.createPage(hostedPage);
+    } catch (error) {
+      /**
+       * TODO: handle Platform Client 400 error e.g.:
```
ThibodeauJF, line 107: How should I standardize this one, it wasn't handled https://github.com/coveo/cli/pull/930

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -0,0 +1,184 @@
...
+    return {
+      name,
+      html: readFileSync(join(dir, htmlEntryFile.path), 'utf-8'),
+      javascript: [
+        ...javascriptEntryFiles.map(({path, isModule}) => ({
+          inlineContent: readFileSync(join(dir, path), 'utf-8'),
+          isModule,
+        })),
+        ...javascriptUrls.map(({path, isModule}) => ({
+          inlineContent: path,
```
btaillon, line 149: Is this meant to be `url`?
```suggestion
          url: path,
```

btaillon (approved):
Looks good to me

packages/cli/commons/src/errors/jsonError.ts
```diff
@@ -0,0 +1,14 @@
+import {ValidationError} from 'jsonschema';
```
louis-bompart, line 1:
```suggestion
import type {ValidationError} from 'jsonschema';
```

louis-bompart: And add jsonschema as dev dep in the project

packages/cli/core/src/commands/ui/deploy.ts
louis-bompart: I reckon you can prolly let it go (no need for try-catch).
The catch-finally clauses at the CLICommand level should handle it for you
(@y-lakhdar amarite?)

y-lakhdar: Yeah, no need for that. the CLI base command will handle the error.
If you need the handle the error in a specific way, you can use the `catch` public method. If that's the case, don't forget to call the `super.catch()` once your done.
Example:
```
 public async catch(err?: Error & {exitCode?: number}) {
    // custom stuff
    return super.catch(err);
  }
```

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -0,0 +1,184 @@
...
+    }),
+    config: Flags.string({
+      char: 'c',
+      description: 'The path to the deployment JSON configuration.',
+      helpValue: 'coveo.deploy.json',
+      default: 'coveo.deploy.json',
+      required: false,
+    }),
+  };
+
```
y-lakhdar, line 84: Do you think you could add an example:

```
import {Example} from '@oclif/core/lib/interfaces';
public static examples: Example[] = [
  {
    command: '...',
    description:
      '...',
  }
];
```

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -0,0 +1,184 @@
...
+    } catch (error) {
+      /**
+       * TODO: handle Platform Client 400 error e.g.:
+       * {
+       *  statusCode: 400,
+       *  message: "A configuration with that name already exist.",
+       *  type: 'DuplicateConfigurationException'
+       * }
+       */
+      console.error(error);
```
y-lakhdar, line 114: no need to `console.error` here. Same comment as above

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -0,0 +1,184 @@
...
+  private async createClient() {
+    const authenticatedClient = new AuthenticatedClient();
+    return await authenticatedClient.getClient();
+  }
+
+  private async createPage(page: New<HostedPage>) {
+    this.log(`Creating new Hosted Page named "${page.name}".`);
+
+    const client = await this.createClient();
+    const response = await client.hostedPages.create(page);
```
y-lakhdar, line 173: I think that would be a good place to use `startSpinner` since this is an async task:
example:
```
startSpinner('Creating hosted search page)
...
stopSpinner(`Hosted Page creation successful with id "${response.id}".`);
```

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -0,0 +1,192 @@
...
+    }),
+    config: Flags.string({
+      char: 'c',
+      description: 'The path to the deployment JSON configuration.',
+      helpValue: 'coveo.deploy.json',
+      default: 'coveo.deploy.json',
+      required: false,
+    }),
+  };
+  public static examples: Example[] = [
```
ThibodeauJF, line 86: @jpmarceau  Added a few examples 😄

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -0,0 +1,192 @@
...
+    },
+    {
+      command: 'coveo ui:deploy -p 7944ff4a-9943-4999-a3f6-3e81a7f6fb0a',
+      description:
+        'Update the Hosted Page whose ID is "7944ff4a-9943-4999-a3f6-3e81a7f6fb0a" according to the configuration in the file "coveo.deploy.json"',
+    },
+    {
+      command: 'coveo ui:deploy -c ./configs/myconfig.json',
+      description:
+        'Create a new Hosted Page according to the configuration in the file at the path "./configs/myconfig.json"',
```
jpmarceau, line 100:
```suggestion
        'Create a new Hosted Page according to the configuration in the file located at "./configs/myconfig.json"',
```

jpmarceau (approved):
👍

y-lakhdar (approved):

louis-bompart (approved):

PR #1100, opened 2023-02-09
chore: refactor release script

louis-bompart:
Refactor release script into an utils package, this saves us from the ../../../..././//....//./ hell.


ThibodeauJF (approved):

utils/release/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "async-retry": "1.3.3",
+    "conventional-changelog-angular": "5.0.13",
+    "npm-registry-fetch": "14.0.3",
+    "octokit": "2.0.14",
+    "semver": "7.3.8",
+    "ts-dedent": "2.2.0"
+  },
+  "bin": {
+    "coveo-release-1": "./phase1-bump-all-packages.mjs",
+    "coveo-release-2": "./phase2-git-commit-tag-push.mjs"
```
btaillon, line 18: I think both of these scripts should be renamed to something other than `1` and `2`.

`coveo-release-1` could instead be named something like `auto-bump-version`. Its script file could also be renamed to remove "all" from its name.

`coveo-release-2` could be named something like `commit-and-push-bumps`.

utils/release/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@coveo/release",
```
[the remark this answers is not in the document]
louis-bompart: I used `@coveord` just to be on the safer side.

utils/release/package.json
louis-bompart: Script renamed!

PR #1112, opened 2023-02-20
feat: improve custom component search page 

y-lakhdar:
## Proposed changes
* Enhanced results templates
* Updated facets
* Dedicated pipeline with featured results

### Before
image

### After
image

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/ui/atomic/create-atomic-component-project/template/src/pages/index.css
```diff
@@ -1,3 +1,7 @@
+:root {
+  --atomic-font-family: Lexend;
```
y-lakhdar, line 2: Updated font to match Barca font

packages/ui/atomic/create-atomic-component-project/template/src/pages/index.html
```diff
@@ -5,28 +5,42 @@
...
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Test-2121</title>
-    <link href="/index.css" rel="stylesheet" />
+    <link
+      rel="stylesheet"
+      href="https://fonts.googleapis.com/css?family=Lexend:200,300,400,500,600,700,900&display=swap"
+    />
     <link href="/atomic/themes/coveo.css" rel="stylesheet" />
-    <!-- 
+    <link href="/index.css" rel="stylesheet" />
```
y-lakhdar, line 13: Ninja fix:
index.css should come after coveo.css

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/results-manager.tsx
```diff
@@ -0,0 +1,35 @@
+import {Component, h} from '@stencil/core';
```
y-lakhdar, line 1: Should we not include any component and put all the templates directly in the index.html?

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/templates/youtube-template.html
```diff
@@ -0,0 +1,71 @@
...
+    margin-right: 0.25rem;
+  }
+
+  .smooth-border {
+    border-radius: 5px;
+    overflow: hidden;
+  }
+
+  .force-fit-content {
+    height: fit-content !important;
```
louis-bompart, line 19: We really !important ?
I have ptcssd when i see this keyword xD

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/results-manager.tsx
louis-bompart: wdym?

y-lakhdar: results-manager.tsx... should it be part of the starter project?
or if all the templates, should be added directly into index.html?

louis-bompart: Mhhh I don't know xD
I'll go with the public vote.

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/templates/youtube-template.html
y-lakhdar: I know. me too. Other option would be to use a more specific selector to overwrite the height property defined in atomic:
`.result-root.with-sections.display-list.image-small atomic-result-section-visual.force-fit-content`

louis-bompart: https://github.com/coveo/ui-kit/blob/639d19c57940607d85e431e1a042ee941c3cf6c7/packages/atomic/src/pages/examples/fashion.html#L171-L209

I think `atomic-result-section-badges.force-fit-content` might work

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/results-manager.tsx
ThibodeauJF: For a whole search page it was cleaner to have this component and split the templates in different files. Here I figure it's more of a hassle. You could just add it to the HTML

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/templates/youtube-template.html
btaillon: If a component needs to manipulate the height or positioning of a OOTB section, that means one of two things:
1. The original behaviour is a bug, we should fix it in Atomic.
2. Our case requires customization that isn't needed for most other result templates. We should not use sections at all.
 
In principle, only Atomic should be responsible for the layout of result sections. If we try to forcefully change part of it, then future changes in Atomic may break the layout. We should strongly discourage this.

In this particular case, what does it look like without `fit-content`? I wonder which of the above two cases it is.

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/results-manager.tsx
```diff
@@ -0,0 +1,35 @@
...
+@Component({
+  tag: 'results-manager',
+  shadow: true,
+})
+export class ResultsManager {
+  public render() {
+    return (
+      <atomic-folded-result-list display="list" image-size="small">
+        <atomic-result-template must-match-source="Youtube">
+          <template innerHTML={youtubeTemplate}></template>
```
btaillon, line 21: In case you decide that you prefer to keep the templates.
```suggestion
          {/* deepcode ignore ReactSetInnerHtml: static */}
          <template innerHTML={youtubeTemplate}></template>
```

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/results-manager.tsx
```diff
@@ -0,0 +1,35 @@
...
+        </atomic-result-template>
+        <atomic-result-template must-match-source="Sports - Blog">
+          <template innerHTML={blogTemplate}></template>
```
btaillon, line 24:
```suggestion
          {/* deepcode ignore ReactSetInnerHtml: static */}
          <template innerHTML={blogTemplate}></template>
```

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/results-manager.tsx
```diff
@@ -0,0 +1,35 @@
...
+        </atomic-result-template>
+        <atomic-result-template must-match-source="Chatter">
+          <template innerHTML={chatterTemplate}></template>
```
btaillon, line 27:
```suggestion
          {/* deepcode ignore ReactSetInnerHtml: static */}
          <template innerHTML={chatterTemplate}></template>
```

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/results-manager.tsx
```diff
@@ -0,0 +1,35 @@
...
+        </atomic-result-template>
+        <atomic-result-template>
+          <template innerHTML={defaultTemplate}></template>
```
btaillon, line 30:
```suggestion
          {/* deepcode ignore ReactSetInnerHtml: static */}
          <template innerHTML={defaultTemplate}></template>
```

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/results-manager.tsx
btaillon: I'm fine with either

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/templates/youtube-template.html
y-lakhdar: Without the fit-content, the image dimension is fixed to 10rem x 10rem 
image
I think this makes sense images with a 1:1 width/height ratio (which is not the case here). Otherwise, the space not used by the image is left blank.

I used `fit-content` to only use the necessary height needed by the image
image

y-lakhdar: @btaillon, does it make sense to have this change directly in Atomic?

btaillon: Is this with `image-size` small?

You can change the image-size on a per-template basis by adding `image-size="large"` to the visual section.

y-lakhdar: The image still has a 1:1 ratio
image

btaillon: Part of why I forced a 1:1 ratio was because, on mobile results (or any results where the image is at the top), it ensured the titles were aligned. However, I do see in this case that it doesn't really make sense...

I think for desktop list results in Atomic, it would be fine to add `fit-content` (for everyone using Atomic), or at-least align the contents of the section to the top.

ThibodeauJF (approved):

packages/ui/atomic/create-atomic-component-project/template/src/components/results-manager/templates/youtube-template.html
y-lakhdar: Created a JIRA 
https://coveord.atlassian.net/browse/KIT-2314

PR #1131, opened 2023-02-27
feat(create-atomic): adapt project to use ui:deploy instead of netlify

ThibodeauJF:
KIT-2269

## Proposed changes

The create atomic project:
- get rid of netlify & related things
- get rid of the lambda
- adds an esbuild bundle to package all of the custom code in 1 JS, 1 CSS & parse the html
- all of this to be able to leverage the new coveo ui:deploy command

Starting from the previous output, I made this PR in order to know what to change https://github.com/ThibodeauJF/atomic-deploy/pull/1

## Breaking changes

The output of the project will change, we could see it as a breaking change, or not.

## Testing

- [x] Unit Tests:

- [x] Functionnal Tests:

- [x] Manual Tests:


packages/ui/atomic/template/deploy.esbuild.mjs
```diff
@@ -0,0 +1,50 @@
+import {build} from 'esbuild';
+import {join} from 'path';
```
louis-bompart, line 2: Do prefer `node:` import
```suggestion
import {join} from 'node:path';
```

packages/ui/atomic/template/deploy.esbuild.mjs
```diff
@@ -0,0 +1,50 @@
...
+  const hostedPageContents =
+    /<atomic-hosted-page>((.|\n)*)<\/atomic-hosted-page>/gm
+      .exec(htmlContents)[1]
+      .trim();
+
+  writeFileSync(outputFile, hostedPageContents, 'utf-8');
+}
+
+async function main() {
+  ensureDirSync('dist/bundle');
```
louis-bompart, line 45: `mkdirSync` does not throw when the dir exists if `recursive:true`.
```suggestion
  mkdirSync('dist/bundle', {recursive:true});
```

packages/ui/atomic/template/package.json
```diff
@@ -10,13 +10,12 @@
...
-    "@coveo/search-token-lambda": "1.36.3",
-    "@stencil/core": "2.22.2"
+    "@stencil/core": "2.17.3"
   },
   "devDependencies": {
     "@rollup/plugin-replace": "5.0.2",
-    "fkill": "8.1.0",
-    "get-port": "6.1.2",
+    "esbuild": "0.17.10",
+    "fs-extra": "11.1.0",
```
louis-bompart, line 18: Note: you'll need to regen the lockfile tho.
```suggestion
```

louis-bompart (commented):
Remove fs-extra.

packages/ui/atomic/template/package.json
```diff
@@ -10,13 +10,12 @@
   "dependencies": {
     "@coveo/atomic": "2.19.4",
     "@coveo/headless": "2.8.3",
-    "@coveo/search-token-lambda": "1.36.3",
-    "@stencil/core": "2.22.2"
+    "@stencil/core": "2.17.3"
```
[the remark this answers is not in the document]
louis-bompart: Mh no: if the user uses the CLI through the binaries it would install twice, and the CLI ain't lightweight enough to be 'OK' about it.

Maybe isolating the `deploy` command into its own plugin would be OK to make it lightweight enough (and avoid a cyclic dependency), but I'm not sure.

packages/ui/atomic/template/deploy.esbuild.mjs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+function html() {
+  const entry = join('src/pages/index.html');
+  const outputFile = join('dist/bundle/index.html');
+  const htmlContents = readFileSync(entry, 'utf-8');
+  const hostedPageContents =
+    /<atomic-hosted-page>((.|\n)*)<\/atomic-hosted-page>/gm
+      .exec(htmlContents)[1]
+      .trim();
```
btaillon, line 38: If we ever add props to `atomic-hosted-page`, this will need to be updated. I think we should either:
1. Change the regexp to support props (e.g., by replacing `<atomic-hosted-page>` with `<atomic-hosted-page(\s[^>]*)?>`), which is still flaky if we add a prop which can take a string that may contain `>`.
2. Add an HTML parser as a dev dependency.
  * I would prefer this.

Also, I think we should add some error handing in case a confused implementer removes `atomic-hosted-page` or has multiple.

packages/ui/atomic/create-atomic/scripts/packageTemplate.json
```diff
@@ -9,15 +9,11 @@
...
-    "start": "node start-netlify.mjs",
+    "start": "stencil build --dev --watch --serve",
     "build": "stencil build",
-    "dev:live": "npm run site:link && node start-netlify.mjs --live",
-    "site:variables": "netlify env:import .env",
-    "site:init": "netlify init && npm run site:variables",
-    "site:deploy": "npm run site:link && npm run build && npm run site:variables && netlify deploy --prod --open",
-    "postinstall": "npm run setup-cleanup && prettier --write . --loglevel warn",
-    "prepare": "npm run setup-lambda"
+    "build:deploy": "node deploy.esbuild.mjs",
```
btaillon, line 15: The name `deploy` (in both the `package.json` script name and the file name), intuitively, makes me think that this script may also deploy something online.

I think we could use something like "deployable" instead.

packages/ui/atomic/create-atomic/scripts/packageTemplate.json
```diff
@@ -9,15 +9,11 @@
...
+    "deploy": "npm run build && npm run build:deploy && coveo ui:deploy",
```
btaillon, line 16: `deploy.esbuild.mjs` should run fairly fast, I think we could include it in the `build` script.

packages/ui/atomic/template/deploy.esbuild.mjs
ThibodeauJF: Nothing will be added here it's not a functional component, just to indicate what to parse in order to deploy that piece of html

https://github.com/coveo/cli/blob/KIT-2269/packages/ui/atomic/template/src/pages/index.html.hbs#L23

packages/ui/atomic/template/src/components/sample-component/sample-component.tsx
```diff
@@ -51,7 +52,8 @@ export class SampleComponent {
   // Using `componentWillLoad` will hang the parent atomic-search-interface initialization.
   public async connectedCallback() {
     try {
-      // Wait for the Atomic bindings to be resolved.
+      // Wait for the Atomic to load and bindings to be resolved.
+      await waitForAtomic();
       this.bindings = await initializeBindings(this.host);
```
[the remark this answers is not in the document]
ThibodeauJF: Not initialized defined here too, its rather unique to the HSP where the atomic script is added as an src and the custom code is added as inline.

packages/cli-e2e/__tests__/atomic.specs.ts
```diff
@@ -254,6 +251,8 @@ describe('ui:create:atomic', () => {
                 '**CHANGELOG.md',
                 '**package.json',
                 'stencil.config.ts',
+                'coveo.deploy.json',
+                'deployment.esbuild.mjs',
```
louis-bompart, line 255: Non-static file I guess?

packages/ui/atomic/create-atomic/scripts/packageTemplate.json
```diff
@@ -9,15 +9,10 @@
   "collection": "dist/collection/collection-manifest.json",
   "collection:main": "dist/index.js",
   "scripts": {
-    "setup-lambda": "node ./scripts/setup-lambda.mjs",
     "setup-cleanup": "node ./scripts/clean-up.js",
```
louis-bompart, line 12: Check if we still need clean-up?

packages/cli-e2e/__tests__/atomic.specs.ts
louis-bompart: Actually, deployment seems static :suspect: if so, do not exclude it :D

packages/ui/atomic/create-atomic/scripts/packageTemplate.json
ThibodeauJF: We do

louis-bompart (approved):
lgtm overall, if possible I'd prefer we plug vite in there instead of esbuild (tho I reckon its actually already there?)

packages/ui/atomic/template/package.json
```diff
@@ -10,13 +10,11 @@
...
   },
   "devDependencies": {
     "@rollup/plugin-replace": "5.0.2",
-    "fkill": "8.1.0",
-    "get-port": "6.1.2",
+    "esbuild": "0.17.10",
```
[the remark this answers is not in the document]
ThibodeauJF: It's not really important... the build is done with stencil mainly esbuild is just a quick easy to learn tool that compiles all the output together. We can see about it later, as you say and:

> Vite [pre-bundles dependencies](https://vitejs.dev/guide/dep-pre-bundling.html) using [esbuild](https://esbuild.github.io/). esbuild ...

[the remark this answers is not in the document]
louis-bompart: fair, carry on :D

jpmarceau (approved):
👌 
We'll remove mentions of Netlify from the doc, and we plan on reviewing the Atomic tutorials soon to use the CLI :hype:

srosenberg-coveo (approved):

PR #1135, opened 2023-03-03
fix(atomic): generate search/analytics API key instead of impersonate API key

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-2332

## Proposed changes
We don't need to generate an impersonate key anymore since the search token lambda is 💀. Just the straight search token will do.

## Testing

- [X] Functionnal Tests:


packages/cli/core/README.md
```diff
@@ -1067,6 +1067,36 @@ FLAGS
 
 DESCRIPTION
   Deploy your search application to the Coveo infrastructure.
+  The target `config` if present, must contain the following parameters:
```
ThibodeauJF, line 1070: Hope it doesn't delete it again https://github.com/coveo/cli/commit/65e57c953aca07c47ac36267752e363ca4661ada

louis-bompart: It will.
You need to use `summary` for the brief tldr of your command, and `description` for the lengthy doc
https://oclif.io/docs/commands

louis-bompart (changes requested):
should be done in the command's code.

packages/cli/core/README.md
ThibodeauJF: ahhh!

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -107,6 +107,36 @@ export default class Deploy extends CLICommand {
...
+      {
+        "path": "The path to a bundled CSS file."
+      }
+    ],
+    "cssUrls": [
+      {
+        "path": "The URL of the CSS stylesheet."
+      }
+    ]
+  }`;
```
louis-bompart, line 139:
```ts
export class MyCommand extends Command {
  static summary = 'A brief overview of your command.'
  static description = `
An in-depth description of the command.
It can be multiline.
`
```
https://oclif.io/docs/commands

You will probably need both, and the long stuff should be in the `description`.
I reckon that `summary` will be used to describe the command when checking a topic etc and `description` for the more verbose stuff.

louis-bompart (changes requested):
Use description for long stuff, and summary for short stuff.

packages/cli/core/src/commands/ui/deploy.ts
ThibodeauJF: read too fast

PR #1139, opened 2023-03-08
chore: propose new taxonomy for atomic commands

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1358

------

[See the proposed file!](https://github.com/coveo/cli/blob/rfc/cli-taxonomy-v2/rfcs/0001-cli-atomic-taxonomy.md)

### DIFF
--- a/rfcs/accepted/0001-cli-atomic-taxonomy.md
+++ b/rfcs/accepted/0001-cli-atomic-taxonomy.md
@@ -1,0 +1,106 @@
+- Feature Name: Taxonomy for Atomic in the Coveo CLI
+- Start Date: 2023-03-08
+
+# Summary
+
+[summary]: #summary
+
+Create a one-stop-shop topic for CLI user that uses Atomic:
+
+- `atomic:deploy` alias `ui:deploy`
+- `atomic:init` would either 'run' `ui:create:atomic` or start a blank lib project `npm init @coveo/atomic-component-project`, depending on user input.
+- `atomic:cmp` would either `npm init @coveo/atomic-component` or `npm init @coveo/atomic-result-component`
+
+# Motivation
+
+[motivation]: #motivation
+
+With the addition of the Custom Component creation in the Coveo CLI, an issue arise:
+A user using atomic would have to navigate in at least 3 topics:
+
+- `ui`, for `ui:deploy`
+- `ui:create`, for `ui:atomic`
+- `tbd` for the custom components commands, or mix it up in `ui:create`.
+
+To avoid this confusion, we should have everything related to atomic present in an eponym topic (directly or as an alias)
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+An Oclif command can be aliased to another with minimal configuration:
+
+```ts
+import {Command, Flags} from '@oclif/core';
+
+export class ConfigIndex extends Command {
+  static aliases = ['config:index', 'config:list'];
+}
+```
+
+https://oclif.io/docs/aliases
+
+That work if the logic required by the alias is identical to the aliased command, otherwise, it is recommended to "break up the code so that it can be called directly" ([Oclif doc](https://oclif.io/docs/running_programmatically))
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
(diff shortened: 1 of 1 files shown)


rfcs/0001-cli-atomic-taxonomy.md
```diff
@@ -0,0 +1,106 @@
...
+1. Add an alias on `ui:deploy`
+2. Refactor logic of `ui:create:atomic` in another file
+3. Create `atomic:init`.
+   - Flags: `type`, can be either `lib(rary)` or `app(lication)`. Optional.
+   - Args: `name`, name of the project. Required.
+   - Logic:
+     - if `type` undefined, ask the user what kind of project he want to init. Store in `type`
+     - then if `type` is `lib`, run `npm init @coveo/atomic-component-project <args.name>`
+     - then if `type` is `app`, run invoke the logic of `ui:create:atomic` that we refactored on step 2
+4. Create `atomic:init`.
```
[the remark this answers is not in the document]
louis-bompart: good catch 😂 copy pasted too fast

louis-bompart:
Based on #1141, here's what it would like for a user:

ThibodeauJF:
Why drop the ui: ? Otherwise I think it's clearer this way!

ThibodeauJF (approved):

louis-bompart:
> Why drop the ui: ? Otherwise I think it's clearer this way!

We would not drop the `ui:`, it's just aliasing.

> Deploy it for any project too it's not only atomic, so here it's just an alias?

At this stage, only a simple alias. If we see that more checks should be made specifically relative to the atomic:deploy, we would refactor the code to ensure `ui:deploy` still works, and `atomic:deploy` is enhanced

olamothe (approved):
What will end up appearing in the doc ?

When you explore the tree of available command with `--help`, how does the alias show up ?

louis-bompart:
> What version of the command will end up appearing in the doc ? The alias one or the "og" ?

Both will appear (so essentially the doc is 'duped')

> When you explore the tree of available command with `--help`, how does the alias show up ?

As any command might

rfcs/accepted/0001-cli-atomic-taxonomy.md
```diff
@@ -0,0 +1,106 @@
...
+
+# Summary
+
+[summary]: #summary
+
+Create a one-stop-shop topic for CLI user that uses Atomic:
+
+- `atomic:deploy` alias `ui:deploy`
+- `atomic:init` would either 'run' `ui:create:atomic` or start a blank lib project `npm init @coveo/atomic-component-project`, depending on user input.
+- `atomic:cmp` would either `npm init @coveo/atomic-component` or `npm init @coveo/atomic-result-component`
```
btaillon, line 12: I would prefer if we didn't try to shorten the word "component", and made this command use the full word: `atomic:component`. I don't feel like it's very obvious that `cmp` means "component".

rfcs/accepted/0001-cli-atomic-taxonomy.md
```diff
@@ -0,0 +1,106 @@
+- Feature Name: Taxonomy for Atomic in the Coveo CLI
+- Start Date: 2023-03-08
+
+# Summary
+
+[summary]: #summary
+
+Create a one-stop-shop topic for CLI user that uses Atomic:
+
+- `atomic:deploy` alias `ui:deploy`
```
btaillon, line 10: If we have to define one of those two as the "main" command and the other as an alias, I would prefer if we said "ui:deploy" is the main command and "atomic:deploy" is the alias, since AFAIK, `ui:deploy` is framework-agnostic.

btaillon (approved):
looks good

rfcs/accepted/0001-cli-atomic-taxonomy.md
louis-bompart: Alright, let's do both :D `atomic:component` and its alias `atomic:cmp`

rfcs/accepted/0001-cli-atomic-taxonomy.md
louis-bompart: Sorry I now noticed, I meant alias as a verb i.e. to alias something. Will clarify

PR #1140, opened 2023-03-08
chore: setup rfcs

louis-bompart:
Provide a minimal "how-to" rfcs.

Our RFC would be a bit like ADR (Architecture Decision Record), but I think the most important part is that when a non-trivial decision or a decision that we think might have an impact on the long run has to be made, it is done and associated with the code.


btaillon (approved):
makes sense IMO

PR #1141, opened 2023-03-08
feat(cli): implement atomic topic

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1358

------

## Proposed changes

See #1139.

- [x] Add beta decorator on all `atomic:init` and `atomic:cmp`
- [x] Extract beta decorator in another PR (scoping) : #1150 (will need to rebase)
- [ ] #1150 merged

## Reviewing instructions:
This PR follows the steps described in the RFC.
The PR is not too long to be reviewed in one shot, but you'll have a butter smooth experience if you review it commit by commit.

I'll do the unit tests in four commits as described in the section below.

## Testing

- [x] Unit Tests:
  - [x]  `atomic:deploy`: skip or refactor `ui:deploy`: skip.
  - [x] `atomic:cmp`: ensure we call `spawnProcess` nicely.
  - [x] `atomic:init`: ensure we call the proper `createAtomicXYZ` function
  - [x] `createAtomicProject`: ensure we call `spawnProcess` nicely.


packages/cli/core/src/commands/atomic/cmp.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  public static args = [
+    {name: 'name', description: 'the name of your component', required: true},
+  ];
+
+  public async run(): Promise<void> {
+    const {initializer, name} = await this.getSpawnOptions();
+
+    const cliArgs = ['init', initializer, name];
+    await spawnProcess(appendCmdIfWindows`npm`, cliArgs);
```
btaillon, line 32: There seems to be a decent amount of instances of `spawnProcess` being used to spawn NPM, and it may not be immediately intuitive that we need to append the CMD if we make other commands which spawn NPM. I think it may be worth extracting a `spawnNPM` util, or more specific functions like `npmInit`, `npmInstall`, etc.

btaillon (approved):
Looks good! Left some comments on the RFC in the other PR

packages/cli/core/src/commands/atomic/cmp.ts
louis-bompart: Could be nice, I'll keep that out-of-scope of this PR tho, there's already a lot.

louis-bompart: CDX-1372

packages/cli/core/src/commands/atomic/init.ts
```diff
@@ -0,0 +1,82 @@
...
+  createAtomicApp,
+  createAtomicLib,
+} from '../../lib/atomic/createAtomicProject';
+
+export default class AtomicInit extends CLICommand {
+  public static description =
+    'Scaffold a project to work with Coveo Atomic Framework';
+
+  public static examples = [
+    '<%= config.bin %> <%= command.id %> --type=app myAwesomeSearchPage',
```
olamothe, line 20: Is this the new format for examples in ocliff ?

packages/cli/core/src/commands/atomic/init.ts
```diff
@@ -0,0 +1,82 @@
...
+    '<%= config.bin %> <%= command.id %> --type=lib myCustomAtomicComponentsLibrary',
+  ];
+
+  public static flags = {
+    type: Flags.string({
+      description: 'the kind of project to initialize',
+      options: ['app', 'application', 'lib', 'library'],
```
olamothe, line 27: Why offer "aliases" for the option name ?

packages/cli/core/src/lib/atomic/createAtomicProject.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+          ({
+            createImpersonateApiKey: jest.fn(),
+            getUsername: () => Promise.resolve('bob@coveo.com'),
+          } as unknown as AuthenticatedClient)
+      );
+    };
+
+    const callOptions = {
+      cfg: {
+        accessToken: 'someToken',
```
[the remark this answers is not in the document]
olamothe: 🤫

olamothe (approved):

packages/cli/core/src/commands/atomic/component.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public static examples = [
+    '<%= config.bin %> <%= command.id %> --type=page myAwesomeComponent',
+    '<%= config.bin %> <%= command.id %> --type=result mySuperResultComponent',
+  ];
+
+  public static flags = {
+    type: Flags.string({
+      description: 'the kind of component to initialize',
+      options: ['page', 'result'],
```
olamothe, line 23: Hum, I'm not sure page is the correct name... 🤔 

In the doc, the classification that is used is `components` and `result template components`: https://docs.coveo.com/en/atomic/latest/reference/

packages/cli/core/src/commands/atomic/init.ts
louis-bompart: Yeah, if you do oclif generate command, what it spew out. I tested, it works nicely and replace the bin by coveo, and the id by the command or its alias depending on the use case.

packages/cli/core/src/commands/atomic/component.ts
olamothe: Could it be a boolean option instead ? `--resultComponent`

btaillon: I like the idea of a boolean option.

packages/cli/core/src/commands/atomic/init.ts
louis-bompart: I think it's a nice way to have an abbreviated flag option, while still documenting what they mean fully.
(granted ideally I'd prefer to be able to have `app[lication]` instead of `app, application` in the doc)

packages/cli/core/src/commands/atomic/component.ts
louis-bompart: I disagree here: if we want to add another kind of component/ component template, we would be stuck.

btaillon: Ah, the way I saw it, a result component is kind-of an extension of a page component: it behaves mostly the same, but it automatically receives a result as a parameter in some way.

I think it would be fine to have multiple boolean flags in the future if we add more components that are specific like that (e.g., `--quickviewComponent`). An error would be logged if we try to initialize with multiple of those flags.

However, I don't overall feel very strongly about using a boolean flag or an enum flag. An enum flag does have the advantage that it's more explicit, because if it's missing, we can warn the user that they must make a choice.

I don't know if I like the name "page" as an option, but I can't really think of a better name. Something like `default` implies that `result` may be the default, something like `generic` sounds like it refers to generic types, and something like `simple` implies that it's less powerful than `result`.

PR #1143, opened 2023-03-10
fix(cli-commons): color orgId magenta only when on unix or WT

louis-bompart:
https://coveord.atlassian.net/browse/CDX-1365

## Proposed changes

Some color doesn't work nicely on PS, so let's just skip'em'.
Windows users should use WT if they want niceties.

## Testing

- [x] Unit Tests
- [x] Manual Tests: `packages\cli\core\bin\dev auth:login` with powershell

### DIFF
--- a/packages/cli/commons/src/utils/ux.ts
+++ b/packages/cli/commons/src/utils/ux.ts
@@ -1,8 +1,17 @@
 import {CliUx} from '@oclif/core';
 import {red, green, magenta} from 'chalk';
 
+
+functions isWindows() {
+  return process.platform === 'win32'
+}
+
+function isWindowsTerminal() {
+  return Boolean(process.env['WT_PROFILE_ID']);
+}
+
 export function shouldUseColor() {
-  return process.platform !== 'win32' || Boolean(process.env['WT_PROFILE_ID']);
+  return !isWindows() || isWindowsTerminal();
 }
 
 export function startSpinner(task: string, status?: string) {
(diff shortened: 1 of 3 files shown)


packages/cli/commons/src/utils/ux.ts
```diff
@@ -1,6 +1,10 @@
 import {CliUx} from '@oclif/core';
 import {red, green, magenta} from 'chalk';
 
+export function shouldUseColor() {
+  return process.platform !== 'win32' || Boolean(process.env['WT_PROFILE_ID']);
+}
```
btaillon, line 6: Not a very strong preference, but I would prefer if this was divided in a more verbal(?) kind of way.

Like:
```suggestion
export function shouldUseColor() {
  const isWindows = (process.platform === 'win32');
  if (isWindows) {
    const isWindowsTerminal = Boolean(process.env['WT_PROFILE_ID']);
    return isWindowsTerminal;
  }
  return true;
}
```

btaillon (approved):

packages/cli/commons/src/utils/ux.ts
louis-bompart:
```suggestion

functions isWindows() {
  return process.platform === 'win32'
}

function isWindowsTerminal() {
  return Boolean(process.env['WT_PROFILE_ID']);
}

export function shouldUseColor() {
  return !isWindows() || isWindowsTerminal();
}
```

louis-bompart: Similar to your second proposition, but I found the following more readable.

PR #1205, opened 2023-03-23
feat: generate component docs on build

y-lakhdar:
## Proposed changes
Generate a json file with the component

The `npm run build` command will now generate a `atomic-docs.json` file with the following format:
```json
{
  "timestamp": "2023-03-23T14:53:34",
  "compiler": {
    "name": "@stencil/core",
    "version": "2.22.2",
    "typescriptVersion": "4.9.4"
  },
  "components": [
    {
      "filePath": "./src/my-custom-component.tsx",
      "encapsulation": "shadow",
      "tag": "my-custom-component",
      "readme": "...",
      "docs": "Sample custom Atomic component...",
      // ...
    }
  ]
}
```

This file can later be used for component name validation.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/ui/atomic/create-atomic-component/template/src/components/sample-component/stencil.config.ts
+++ b/packages/ui/atomic/create-atomic-component/template/src/components/sample-component/stencil.config.ts
@@ -13,6 +13,10 @@ export const config: Config = {
     {
       type: 'docs-readme',
     },
+    {
+      type: 'docs-json',
+      file: 'docs/atomic-docs.json',
+    },
   ],
   rollupPlugins: {
     before: [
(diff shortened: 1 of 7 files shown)


packages/ui/atomic/create-atomic-component/template/src/components/sample-component/stencil.config.ts
```diff
@@ -13,6 +13,10 @@ export const config: Config = {
     {
       type: 'docs-readme',
     },
+    {
+      type: 'docs-json',
+      file: 'docs/atomic-docs.json',
```
btaillon, line 18: Like my comment on the other PR, I think we should make this name less specific to Atomic since it's just docs generated by Stencil in our template.

We could name it something like `stencil-docs.json` instead.

btaillon (approved):

PR #1207, opened 2023-03-23
feat: add custom `elementName` property on generated component's package.json

y-lakhdar:
## Proposed changes

Use a dedicated property in package.json to identify the custom component name 

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:

### DIFF
--- a/packages/ui/atomic/create-atomic-component/index.js
+++ b/packages/ui/atomic/create-atomic-component/index.js
@@ -51,7 +51,7 @@ if (componentName) {
       srcPath: `src/components/${componentName}/package.json`,
       destPath: `src/components/${componentName}/package.json`,
       transform: (text) =>
-        text.replaceAll(/@coveo\/sample-component/g, componentName),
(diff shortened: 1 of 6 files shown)


packages/ui/atomic/health-check/src/assertions.ts
```diff
@@ -30,7 +30,17 @@ export function ensureDocFile() {
 }
 
 export function ensureConsistentElementName() {
-  throw 'TODO: CDX-1389';
+  ensureDocFile();
+  const {elementName} = getJsonPkg();
+  const {components} = JSON.parse(readFileSync(componentDocPath).toString());
```
btaillon, line 35: Instead of asserting that the file must be named `atomic-docs.json`, could we make it more generic and assert that it must be named `stencil-docs.json`? `atomic-docs` makes it sound like we're the ones generating it inside `@coveo/atomic` IMO

btaillon (approved):

PR #1213, opened 2023-03-24
fix(atomic): add schemaVersion for deploy.json

louis-bompart:
## Proposed changes

Add a schemaVersion property, no validation for now (we only have one version after all)

## Testing

- [x] Unit Tests


olamothe (approved):
🙏

PR #1224, opened 2023-03-27
chore(ci): protect master branch during releases

louis-bompart:
## Proposed changes

Our release failed because a PR was merged during a release.
To avoid that, we should prevent commits from going into master during a release.

To do that, I limit who can commit to master to our bot.
To use the bot trusted identity, I need more convoluted ways of committing:
 1. create the Git Tree and make it known to GitHub.
 2. create a commit via GitHub APIs using the Git Tree
 3. point master on the commit.

## Testing

I will have to trigger a new release soonish after we complete a stable one to check the process.


utils/release/git-publish-all.mjs
```diff
@@ -34,6 +34,37 @@ const getCliChangelog = () => {
...
+  await gitSetupUser(GIT_USERNAME, GIT_EMAIL);
+  // #endregion
+
+  //#region GitHub authentication
+  const authSecrets = {
+    appId: process.env.RELEASER_APP_ID,
+    privateKey: process.env.RELEASER_PRIVATE_KEY,
+    clientId: process.env.RELEASER_CLIENT_ID,
+    clientSecret: process.env.RELEASER_CLIENT_SECRET,
+    installationId: process.env.RELEASER_INSTALLATION_ID,
```
louis-bompart, line 59: - [ ] I need to scavenge those

utils/release/lock-master.mjs
```diff
@@ -0,0 +1,28 @@
...
+  await octokit.rest.repos.updateBranchProtection({
+    owner: REPO_OWNER,
+    repo: REPO_NAME,
+    branch: 'master',
+    ...currentProtection,
+    restrictions: canWrite
+      ? null
+      : {
+          users: [],
+          apps: ['developer-experience-bot'],
```
y-lakhdar, line 24: Do you need to add the `[bot]` prefix here?

utils/release/npm-publish-package.mjs
```diff
@@ -44,6 +47,22 @@ const isPrerelease = process.env.IS_PRERELEASE === 'true';
...
 // Run on each package, it generate the changelog, install the latest dependencies that are part of the workspace, publish the package.
 (async () => {
+  // Lock-out master
+  await changeMasterWriteAccess(false);
+
+  // Verify master has not changed
+  const local = await getSHA1fromRef('master');
+  await gitPull();
+  const remote = await getSHA1fromRef('master');
+  if (!process.env.NO_LOCK && local !== remote) {
```
y-lakhdar, line 57: wouldn't it be easier to read without the double negative?

louis-bompart: We can use another name for the variable name, but the `!process.env.OTHER_NAME` should stay:
Indeed, we're testing the presence of the variable, because it is not set outside of the E2E tests.

utils/release/lock-master.mjs
louis-bompart: No we're using the appSlug: https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app
https://github.com/apps/developer-experience-bot

y-lakhdar (approved):

utils/release/npm-publish-package.mjs
```diff
@@ -44,6 +47,22 @@ const isPrerelease = process.env.IS_PRERELEASE === 'true';
 
 // Run on each package, it generate the changelog, install the latest dependencies that are part of the workspace, publish the package.
 (async () => {
+  // Lock-out master
+  await changeMasterWriteAccess(false);
```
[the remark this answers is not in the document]
louis-bompart: 👍

louis-bompart:
Back to draft, there's quite some stuff to fix

utils/release/lock-master.mjs
```diff
@@ -0,0 +1,28 @@
+import {Octokit} from 'octokit';
+
+const REPO_OWNER = 'coveo';
+const REPO_NAME = 'cli';
+
+export async function changeMasterWriteAccess(canWrite) {
```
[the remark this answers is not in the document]
louis-bompart: fixed in 2d44708f

utils/release/npm-publish-package.mjs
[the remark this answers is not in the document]
louis-bompart: fixed in f9cbea8

utils/release/git-publish-all.mjs
```diff
@@ -63,33 +97,30 @@ const getCliChangelog = () => {
...
-    package.json
-    package-lock.json
-    packages/ui/cra-template/template.json
-    `,
-    PATH
+
+  const commit = await commitChanges(
+    releaseNumber,
+    gitNewTag,
+    packagesReleased,
```
btaillon, line 104: I think it's not so transparent what these two parameters are for (`gitNewTag` and `packagesReleased`).

Instead, I think we could build the message right here.

For example:
```js
const commit = await commitChanges({
  octokit,
  releaseNumber,
  message: dedent`
    [version bump] chore(release): release ${gitNewTag} [skip ci]
    ${packagesReleased}
    **/README.md
    **/CHANGELOG.md
    **/package.json
    README.md
    CHANGELOG.md
    package.json
    package-lock.json
    packages/ui/cra-template/template.json
  `,
});
```

utils/release/git-publish-all.mjs
```diff
@@ -102,6 +133,77 @@ const getCliChangelog = () => {
   });
 })();
 
+async function commitChanges(
```
btaillon, line 136: There's a lot of utilities in this file, I think it may be a good idea to extract this one into a separate file. I really don't feel strongly about it though.

utils/release/lock-master.mjs
```diff
@@ -0,0 +1,32 @@
...
+  changeBranchRestrictions(null);
+
+async function changeBranchRestrictions(restrictions) {
+  const octokit = new Octokit({auth: process.env.GITHUB_CREDENTIALS});
+
+  const currentProtection = await octokit.rest.repos.getBranchProtection({
+    owner: REPO_OWNER,
+    repo: REPO_NAME,
+    branch: 'master',
+  });
```
btaillon, line 23: With the typescript language server, it looks like `getBranchProtection` doesn't directly return the protections, it returns a `data` object. If so, it looks like you have an error.

```suggestion
  const {status, data: currentProtection} =
    await octokit.rest.repos.getBranchProtection({
      owner: REPO_OWNER,
      repo: REPO_NAME,
      branch: 'master',
    });
  if (status !== 200) {
    throw 'Could not get current branch protections.';
  }
```

btaillon (approved):
Looks good! Apart from one error

utils/release/git-publish-all.mjs
louis-bompart: 👎 I think the release scripts should be as much be of possible be scripts.

If we were doing manual releases, we would have a list of successive steps a human would do.
We have code that does it instead of a human, but I think it is best for the code to have the same structure as the 'instructions set' we would give to a human.

utils/release/lock-master.mjs
louis-bompart: Good point for the data obj.
We don't need any handling tho, octokit throw when a request is nok

utils/release/git-publish-all.mjs
```diff
@@ -31,9 +35,38 @@ const getCliChangelog = () => {
...
+    privateKey: process.env.RELEASER_PRIVATE_KEY,
+    clientId: process.env.RELEASER_CLIENT_ID,
+    clientSecret: process.env.RELEASER_CLIENT_SECRET,
+    installationId: process.env.RELEASER_INSTALLATION_ID,
+  };
+
+  const octokit = new Octokit({
+    authStrategy: createAppAuth,
+    auth: authSecrets,
+  });
```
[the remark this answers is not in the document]
louis-bompart: I think it's important to keep both very separated because its two very different authentication type, two different permissions set.
The distinction is important to keep.

However, we do not use the SSH here, so I removed it

utils/release/git-lock.mjs
```diff
@@ -0,0 +1,33 @@
+import {gitPull, getSHA1fromRef} from '@coveo/semantic-monorepo-tools';
+import dedent from 'ts-dedent';
+
+import {
+  limitWriteAccessToBot,
+  removeWriteAccessRestrictions,
+} from './lock-master.mjs';
+
+const isPrerelease = process.env.IS_PRERELEASE === 'true';
+const noLockRequired = Boolean(process.env.NO_LOCK);
```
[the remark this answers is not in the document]
louis-bompart: IMO addressed enough in 76e9486a0d299cbfcfdba2fdbceb24fc3da8857b

utils/release/git-lock.mjs
```diff
@@ -0,0 +1,33 @@
...
+
+const ensureUpToDateBranch = async () => {
+  if (isPrerelease || noLockRequired) {
+    return;
+  }
```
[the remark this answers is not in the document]
louis-bompart: 76e9486a0d299cbfcfdba2fdbceb24fc3da8857b

utils/release/git-lock.mjs
```diff
@@ -0,0 +1,33 @@
...
+  // Lock-out master
+  await limitWriteAccessToBot();
+
+  // Verify master has not changed
+  const local = await getSHA1fromRef('master');
```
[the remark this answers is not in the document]
louis-bompart: 76e9486a0d299cbfcfdba2fdbceb24fc3da8857b

PR #1234, opened 2023-03-28
feat: find pageId from name for update

louis-bompart:
## Proposed changes

When a user tries to upload a page with a name/orgId combination that already exists, ask him if he wants to overwrite it.

## Testing

- [x] Unit Tests:

### DIFF
--- a/packages/cli/core/src/commands/ui/deploy.ts
+++ b/packages/cli/core/src/commands/ui/deploy.ts
@@ -161,9 +161,9 @@ export default class Deploy extends CLICommand {
   )
   public async run() {
     await this.computeFlags();
-    await this.computeOrgId();
+    this.computeOrgId();
     await this.instantiatePlatformClient();
-    await this.computeHostedPage();
+    this.computeHostedPage();
     await this.computePageId();
 
     if (this.pageId) {
@@ -177,7 +177,7 @@ export default class Deploy extends CLICommand {
     this.flags = (await this.parse(Deploy)).flags;
   }
 
-  private async computeOrgId() {
+  private computeOrgId() {
     this.organization = getTargetOrg(
       this.configuration,
       this.flags.organization
@@ -190,7 +190,7 @@ export default class Deploy extends CLICommand {
     });
   }
 
-  private async computeHostedPage() {
+  private computeHostedPage() {
     const deployConfigPath = this.flags.config;
     const deployConfig: DeployConfig =
       readJsonSync(deployConfigPath, {throws: false}) || {};
@@ -230,7 +230,7 @@ export default class Deploy extends CLICommand {
       Overwrite page with ID: "${this.pageId}" in organization ${this.organization}?`))
     ) {
       this.error(
-        `Page name must be unique, try changing the "name" field in "coveo.deploy.json".`
+        'Page name must be unique, try changing the "name" field in "coveo.deploy.json".'
       );
     }
   }


packages/cli/core/src/commands/ui/deploy.spec.ts
```diff
@@ -465,6 +473,69 @@ describe('ui:deploy', () => {
...
+      .do(() => {
+        doMockConfirm();
+        mockedConfirm.mockResolvedValueOnce(true);
+        mockHostedPageList.mockResolvedValueOnce({
+          items: [{name: validJsonConfig.name, id: 'thisistheid'}],
+        });
+      })
+      .command(['ui:deploy'])
+      .it(
+        'should call hostedPages.list when no page id argument is passed and a page with the same already exists',
```
mrrajamanickam-coveo, line 488: "page with the same already exists" -> "page with the same id already exists" ?

packages/cli/core/src/commands/ui/deploy.spec.ts
```diff
@@ -465,6 +473,69 @@ describe('ui:deploy', () => {
...
+          items: [{name: validJsonConfig.name, id: 'thisistheid'}],
+        });
+      })
+      .command(['ui:deploy'])
+      .it(
+        'should call hostedPages.update when no page id argument is passed, a page with the same already exists, and the user confirms the overwrite',
+        () => {
+          expect(mockHostedPageUpdate).toHaveBeenCalledWith({
+            ...expectedHostedPage,
+            id: 'thisistheid',
```
mrrajamanickam-coveo, line 515: extract into a `testId` const e.g. for reusability ?

packages/cli/core/src/commands/ui/deploy.spec.ts
```diff
@@ -465,6 +473,69 @@ describe('ui:deploy', () => {
...
+        mockedConfirm.mockResolvedValueOnce(true);
+        mockHostedPageList.mockResolvedValueOnce({
+          items: [{name: validJsonConfig.name, id: 'thisistheid'}],
+        });
+      })
+      .command(['ui:deploy'])
+      .it(
+        'should call hostedPages.update when no page id argument is passed, a page with the same already exists, and the user confirms the overwrite',
+        () => {
+          expect(mockHostedPageUpdate).toHaveBeenCalledWith({
```
mrrajamanickam-coveo, line 513: could we also check that `mockHostedPageList` is not called in this case?

packages/cli/core/src/commands/ui/deploy.spec.ts
```diff
@@ -465,6 +473,69 @@ describe('ui:deploy', () => {
...
+      .stdout()
+      .stderr()
+      .do(() => {
+        doMockConfirm();
+        mockedConfirm.mockResolvedValueOnce(true);
+        mockHostedPageList.mockResolvedValueOnce({
+          items: [{name: validJsonConfig.name, id: 'thisistheid'}],
+        });
+      })
+      .command(['ui:deploy'])
```
mrrajamanickam-coveo, line 509: This block seems the same as 476 - 485? If yes how are expecting different results than the previous? Are the checks broken up to have one `expect` per test?

packages/cli/core/src/commands/ui/deploy.spec.ts
```diff
@@ -465,6 +473,69 @@ describe('ui:deploy', () => {
...
+        doMockConfirm();
+        mockedConfirm.mockResolvedValueOnce(false);
+        mockHostedPageList.mockResolvedValueOnce({
+          items: [{name: validJsonConfig.name, id: 'thisistheid'}],
+        });
+      })
+      .command(['ui:deploy'])
+      .catch((err) => expect(err).toMatchSnapshot())
+      .it(
+        'should not call hostedPages.update when no page id argument is passed, a page with the same already exists, and the user decline the overwrite',
```
mrrajamanickam-coveo, line 533: "when no page id argument is passed" - Isn't the `id` being passed above? the only diff i can see with the previous tests is the `mockedConfirm` is passed a negative value.

mrrajamanickam-coveo: "the user decline the overwrite" -> "the user declines the overwrite"

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -159,14 +160,47 @@ export default class Deploy extends CLICommand {
...
 
+  private async getPageIdFromPage(
+    hostedPage: New<HostedPage, null>
+  ): Promise<string | null> {
+    const {flags} = await this.parse(Deploy);
+    const orgId = getTargetOrg(this.configuration, flags.organization);
+    const client = await this.createClient();
+    const response = await client.hostedPages.list({
+      filter: hostedPage.name,
+      perPage: 1000,
```
mrrajamanickam-coveo, line 181: Is there a significance to the number 1000. Can it be extracted to a named constant to make its significance/purpose explicit?

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -159,14 +160,47 @@ export default class Deploy extends CLICommand {
...
+    });
+    const page = response.items
+      .filter((page) => page.name === hostedPage.name)
+      .pop();
```
mrrajamanickam-coveo, line 185: .find() might be more appropriate here

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -159,14 +160,47 @@ export default class Deploy extends CLICommand {
...
+
+    if (!page) {
+      return null;
+    }
+    if (await this.confirmOverwrite(page.id, orgId)) {
+      return page.id;
+    } else {
+      this.error(
+        `Page name must be unique, try changing the "name" field in "coveo.deploy.json".`
```
mrrajamanickam-coveo, line 194: How is the id related to name?

louis-bompart: Essentially:
- there's a bijection between `organizationId+pageName` and the `pageId`
- pageIds are unique.
- Therefore, for a given organization, a page name must be unique in this organization

packages/cli/core/src/commands/ui/deploy.ts
louis-bompart: There's no significance, we could probably leave it by default tbh., given https://github.com/coveo/cli/pull/1234#discussion_r1151161482

louis-bompart:
- [ ] fix e2e

packages/cli/core/src/commands/ui/deploy.spec.ts
louis-bompart: Is called you mean? But yeah.

packages/cli/core/src/commands/ui/deploy.spec.ts
louis-bompart: - [ ] group up the expect
- [ ] reword `it` to be clearer

packages/cli/core/src/commands/ui/deploy.spec.ts
louis-bompart: - [ ] refactor id

packages/cli/core/src/commands/ui/deploy.ts
louis-bompart, line 185:
```suggestion
    const page = response.items
      .find((page) => page.name === hostedPage.name);
```

packages/cli/core/src/commands/ui/deploy.ts
louis-bompart: Fun fact, this is what made the E2E fails:
```
[11:02:45]APIError: 
[11:02:45]Status code: 400
[11:02:45]Error code: OutOfRangeParameterValueException
[11:02:45]Message: Value '1000' for parameter 'perPage' must be between 1 and 100
[11:02:45]Request ID: 0fe55e1f-388a-4580-9a4d-76bea6b78067
[11:02:45]    at wrapError (E:\Repos\cli\packages\cli\commons\src\errors\wrapError.ts:12:12)
[11:02:45]    at Deploy.catch (E:\Repos\cli\packages\cli\commons\src\command\cliCommand.ts:28:28)
[11:02:45]    at Deploy.trackError (E:\Repos\cli\packages\cli\commons\src\preconditions\trackable.ts:91:31)
[11:02:45]    at processTicksAndRejections (node:internal/process/task_queues:95:5)
[11:02:45]    at async Deploy._run (E:\Repos\cli\node_modules\@oclif\core\lib\command.js:84:13)
[11:02:45]    at async Config.runCommand (E:\Repos\cli\node_modules\@oclif\core\lib\config\config.js:272:25)
[11:02:45]    at async Object.run (E:\Repos\cli\node_modules\@oclif\core\lib\main.js:74:5)
```

packages/cli/core/src/commands/ui/deploy.spec.ts
louis-bompart: > "when no page id argument is passed" - Isn't the `id` being passed above? the only diff i can see with the previous tests is the `mockedConfirm` is passed a negative value.

No actually, to see what the 'user' has input, assuming no interaction, you can check the `.command` function call
This function essentially is 'call the CLI with this'.
https://oclif.io/docs/testing

packages/cli/core/src/commands/ui/deploy.spec.ts
louis-bompart: I don't think this is relevant after 5683cac40da382d7558e717280ef58e15573dacb, but don't hesitate to reopen the conversation

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -159,14 +160,44 @@ export default class Deploy extends CLICommand {
...
+
+    if (pageId) {
+      await this.updatePage({...hostedPage, id: pageId});
       return;
     }
 
     await this.createPage(hostedPage);
   }
 
+  private async getPageIdFromPage(
```
olamothe, line 173: I added my review on this PR: https://github.com/coveo/cli/pull/1235#discussion_r1152126011 instead of this one ;) 

Woops !

louis-bompart: The continuing discussion here instead of https://github.com/coveo/cli/pull/1235#discussion_r1152126011
-----

I disagree, I think it's better for the 'frontend' code to limit the number of assertions that results in a network egress, especially for non-idempotent calls.
So essentially, I think we should first determine what kind of network call we want to do (update or create) and then have a single call to it.

I agree however that the getPageId is poorly named and could leave to confusion, I'll rework that.

louis-bompart: addressed in https://github.com/coveo/cli/pull/1234/commits/97116135311dc1c46c86fc8dc9d8bd2c6429e18d

olamothe: Ok but right now, with how the code is written its like this, right ? : 

- If page id is provided -> call updatePage -> exit command

- If page id is not provided 
    - call listPage to check if there is already a page with that name 
    - if yes + they want to overwrite, update page, exit command.
    - if yes + they do not want to overwrite, exit command.
    - if no, create page, exit command.

Or am I just reading it wrong ?
I'm trying to propose to do exactly the same thing, just re-ordered a bit differently in the branching.

y-lakhdar (approved):

PR #1235, opened 2023-03-28
fix(atomic): simplify spinner+OK message

louis-bompart:
## Proposed changes

- For 'positive' confirmation (i.e. OK we're done), rely on the ✔️ from the spinner
- Give the user the command to update its page after they created it.

## Testing

- [x] Unit Tests
- [x] E2E

### DIFF
--- a/packages/cli/core/src/commands/ui/deploy.ts
+++ b/packages/cli/core/src/commands/ui/deploy.ts
@@ -227,7 +227,7 @@ export default class Deploy extends CLICommand {
   private async confirmOverwrite(): Promise<void | never> {
     if (
       !(await CliUx.ux.confirm(dedent`
-      Overwrite page with ID: "${this.pageId}" in organization ${this.organization}?`))
+      Overwrite page with ID: "${this.pageId}" in organization ${this.organization}? (y/n)`))
     ) {
       this.error(
         'Page name must be unique, try changing the "name" field in "coveo.deploy.json".'
--- a/packages/cli/core/src/commands/ui/deploy.ts
+++ b/packages/cli/core/src/commands/ui/deploy.ts
@@ -284,15 +284,16 @@ export default class Deploy extends CLICommand {
     startSpinner(`Creating new Hosted Page named "${page.name}".`);
 
     const response = await this.platformClient.hostedPages.create(page);
-    this.log(`Hosted Page creation successful with id "${response.id}".`);
     stopSpinner();
+    this.log(
+      `To update your page, run "coveo ${this.identifier} -p=${response.id}".`
+    );
   }
 
   private async updatePage(page: HostedPage) {
     startSpinner(`Updating existing Hosted Page with the id "${page.id}".`);
 
-    const response = await this.platformClient.hostedPages.update(page);
-    this.log(`Hosted Page update successful with id "${response.id}".`);
+    await this.platformClient.hostedPages.update(page);
     stopSpinner();
   }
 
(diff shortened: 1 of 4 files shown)


packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -159,14 +160,47 @@ export default class Deploy extends CLICommand {
...
+  ): Promise<string | null> {
+    const {flags} = await this.parse(Deploy);
+    const orgId = getTargetOrg(this.configuration, flags.organization);
+    const client = await this.createClient();
+    const response = await client.hostedPages.list({
+      filter: hostedPage.name,
+      perPage: 1000,
+    });
+    const page = response.items
+      .filter((page) => page.name === hostedPage.name)
```
y-lakhdar, line 184: Why not simply use `.find()`

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -159,14 +160,47 @@ export default class Deploy extends CLICommand {
...
+    } else {
+      this.error(
+        `Page name must be unique, try changing the "name" field in "coveo.deploy.json".`
+      );
+    }
+  }
+
+  private async confirmOverwrite(pageId: string, orgId: string) {
+    return await CliUx.ux.confirm(dedent`
+      Overwrite page with ID: "${pageId}" in organization ${orgId}?`);
```
y-lakhdar, line 201: add `? (y/n)` at the end of the message since it is not added by CliUx.

y-lakhdar (approved):

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -159,14 +160,47 @@ export default class Deploy extends CLICommand {
...
+
+    if (pageId) {
+      await this.updatePage({...hostedPage, id: pageId});
       return;
     }
 
     await this.createPage(hostedPage);
   }
 
+  private async getPageIdFromPage(
```
olamothe, line 173: I was confused initially when reading this function. I think it's because of the name.

It's doing more than a simple `getPageId` as its name implies.

One suggestion would be on line 163, to modify the code similar to this: 

```typescript
if(flags.pageId) {
  await this.updatePage({...hostedPage, id: pageId});
   return;
} 

await (this.possiblyOverwritePageWithDuplicateName(hostedPage) || this.createPage(hostedPage))

```

louis-bompart: I disagree, I think it's better for the 'frontend' code to limit the number of assertions that results in a network egress, especially for non-idempotent calls.
So essentially, I think we should first determine what kind of network call we want to do (update or create) and then have a single call to it.

I agree however that the getPageId is poorly named and could leave to confusion, I'll rework that.

packages/cli/core/src/commands/ui/deploy.ts
```diff
@@ -159,14 +160,47 @@ export default class Deploy extends CLICommand {
...
+    hostedPage: New<HostedPage, null>
+  ): Promise<string | null> {
+    const {flags} = await this.parse(Deploy);
+    const orgId = getTargetOrg(this.configuration, flags.organization);
+    const client = await this.createClient();
+    const response = await client.hostedPages.list({
+      filter: hostedPage.name,
+      perPage: 1000,
+    });
```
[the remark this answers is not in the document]
btaillon: Also, in this scenario, if somebody doesn't provide a pageId in either the flags or in the configuration, then we can assume they mean to create a new page. Maybe in the console we can include a message about "If you meant to overwrite a page, do X".

packages/cli/core/src/commands/ui/deploy.ts
louis-bompart: addressed in 7259ddecceafa4d8d8e00001778ae8d29389fe56

louis-bompart:
Most comments were consequences of a lack of rebase (apart from the y/n).
So all addressed in https://github.com/coveo/cli/commit/54b20b48ca7ba3e207144b1b500cf9b67cd60f02

PR #1239, opened 2023-03-29
test: add E2E tests for custom component health-check

y-lakhdar:
## Proposed changes

Add E2E tests for health-check package

## Testing

- [x] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/ui/atomic/create-atomic-component/template/src/components/sample-component/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@coveo/sample-component",
   "version": "0.0.1",
-  "description": "@coveo/atomic component starter",
```
y-lakhdar, line 4: forgot to remove the description from the template (part of CDX-1323)

olamothe (commented):
Do we have any test for the non-happy path ?

Verifying that it blocks the publishing if the health check are not correctly respected ?

y-lakhdar:
> Do we have any test for the non-happy path ?
> Verifying that it blocks the publishing if the health check are not correctly respected ?

Yes, non-happy paths are all listed and covered in packages/ui/atomic/health-check/src/index.spec.ts

packages/ui/atomic/create-atomic-result-component/tests/test.spec.ts
```diff
@@ -108,5 +111,45 @@ describe(PACKAGE_NAME, () => {
         }).status
       ).toBe(0);
     });
+
+    describe('health-check', () => {
```
mrrajamanickam-coveo, line 115: would it make sense to extract the common code to a test utils or use [table driven tests](https://jestjs.io/docs/api#describeeachtablename-fn-timeout) with parameterized pkg names?

mrrajamanickam-coveo: > What part of the common code do you have in mind @mrrajamanickam-coveo?

most of the test code between the 2 pkgs seems the same - hence wondering about reducing duplication

louis-bompart: Everything that does not contain a 'jest' keyword is refactorable IIRC

@y-lakhdar that'd be a big ask, but maybe it'd be interesting to compare side-by-side having all the e2e for the health-check package in the health-check package?

packages/ui/atomic/health-check/e2e/test.spec.ts
```diff
@@ -0,0 +1,174 @@
...
+      testDirname: 'valid-arg',
+      packageName: 'valid-component-name',
+    },
+    {
+      describeName: 'when called with a component name without hyphen',
+      args: ['nohypen'],
+      testDirname: 'invalid-arg',
+      packageName: 'atomic-nohypen',
+    },
+  ])('$describeName', ({args, testDirname, packageName}) => {
```
louis-bompart, line 62: I think you do not need to test all the case here for the `health-check` package, it is  redundant.

y-lakhdar: Sorry, I forgot to move the PR to draft while I was refactoring E2E tests.
I was actually removing redundant tests :P

packages/ui/atomic/health-check/tsconfig.json
```diff
@@ -6,8 +6,8 @@
     "module": "Node16",
     "target": "ESNext",
     "moduleResolution": "Node16",
-    "sourceMap": false,
-    "declaration": false,
+    "sourceMap": true,
+    "declaration": true,
```
y-lakhdar, line 10: Turn on declaration files to run the E2E tests. But don't include in the build that will be published

packages/ui/atomic/health-check/src/e2e/test.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+
+      // TODO: decomment once tested are done with windows
+      // it('should be able to pass health checks', () => {
+      //   const {stdout} = publish(pkgName, testDirectory);
+      //   const message = stripAnsi(stdout.toString()).replace(/\n/g, '');
+      //   expect(message).not.toContain('✖');
+      //   expect(message).toContain('✔ Readme file');
+      //   expect(message).toContain('✔ Component name');
+      //   expect(message).toContain('✔ Required properties in package.json');
+      // });
```
louis-bompart, line 152: - [ ]  todo ;)

louis-bompart (changes requested):
lgtm, when todo is done :D

packages/ui/atomic/health-check/src/e2e/test.spec.ts
y-lakhdar: oups thanks

louis-bompart (approved):
:shipit:

PR #1251, opened 2023-03-31
ci: remove teams and lock admin out

louis-bompart:
## Proposed changes

The data ` getBranchProtection` requires a lot of processing to be usable for the update call.
Instead, let's do 2x APIs call (per lock or unlock action):
 - Include/remove admin in the write protection (just to be safe)
 - Remove/Add the team in the authorized folks.
 
 The app will always have access

### DIFF
--- a/utils/release/lock-master.mjs
+++ b/utils/release/lock-master.mjs
@@ -1,9 +1,18 @@
+/**
+ * Because our release process creates release commits on our main branch,
+ * it needs to reserve the branch when running, so that no 'new commits' come up.
+ */
+
 import {Octokit} from 'octokit';
 
 const REPO_OWNER = 'coveo';
 const REPO_NAME = 'cli';
 const MAIN_BRANCH_NAME = 'master';
-
+const COVEO_CLI_MASTER = {
+  owner: REPO_OWNER,
+  repo: REPO_NAME,
+  branch: MAIN_BRANCH_NAME,
+};
 export const limitWriteAccessToBot = () => changeBranchRestrictions(true);
 
 export const removeWriteAccessRestrictions = () =>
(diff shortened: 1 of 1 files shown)


utils/release/lock-master.mjs
```diff
@@ -4,32 +4,36 @@ const REPO_OWNER = 'coveo';
...
+    await octokit.rest.repos.setTeamAccessRestrictions({
+      branch: MAIN_BRANCH_NAME,
+      owner: REPO_OWNER,
+      repo: REPO_NAME,
+      teams: ['dx'],
+    });
+    await octokit.rest.repos.setAdminBranchProtection({
+      branch: MAIN_BRANCH_NAME,
+      owner: REPO_OWNER,
+      repo: REPO_NAME,
```
mrrajamanickam-coveo, line 36: can we extract the default params into an object so that it can be reused and it would also make it clear what additional params are being passed

utils/release/lock-master.mjs
```diff
@@ -4,32 +4,36 @@ const REPO_OWNER = 'coveo';
...
-    repo: REPO_NAME,
-    branch: MAIN_BRANCH_NAME,
-    ...currentProtection,
-    restrictions,
-  });
+      owner: REPO_OWNER,
+      repo: {},
+    });
+  } else {
+    await octokit.rest.repos.setTeamAccessRestrictions({
```
mrrajamanickam-coveo, line 27: is there a possibility of one of these calls failing due to some (github) error and leaving the permissions in an inconsistent state that results in the team being locked out?

utils/release/lock-master.mjs
```diff
@@ -4,32 +4,36 @@ const REPO_OWNER = 'coveo';
...
-    teams: [],
-  });
+export const limitWriteAccessToBot = () => changeBranchRestrictions(true);
 
 export const removeWriteAccessRestrictions = () =>
-  changeBranchRestrictions(null);
+  changeBranchRestrictions(false);
 
-async function changeBranchRestrictions(restrictions) {
+async function changeBranchRestrictions(onlyBot) {
```
mrrajamanickam-coveo, line 12: Not sure if its documented elsewhere but a brief doc on why this function is necessary and how it helps would be useful

mrrajamanickam-coveo (commented):
if this is CI related PR can't the pr title be `ci: ..`? it would help distinguish from src code fixes.

utils/release/lock-master.mjs
louis-bompart: There is a chance that the call fails and that we end up without write access longer than required.
However, lock out, no: this does not affect our admin rights on the repo (so anyone of us can at any time restore the write access)

utils/release/lock-master.mjs
louis-bompart: Yeah, the whole release process is in dire need of documentation, but no time has been allocated to that.
I'll write a small comment near the function just as a reminder tho.

louis-bompart: https://github.com/coveo/cli/pull/1251/files#diff-521eca7410150ac5cafa6afc7800332f5351d26b89e3b9f4245fd3c3d1eeb146R1-R5

btaillon (approved):

PR #1261, opened 2023-04-03
ci: fix missing imports, bad func usage, add fail check when release in progress

louis-bompart:
## Proposed changes

- Fixes missing imports that caused https://github.com/coveo/cli/actions/runs/4597808599/jobs/8121171408 :x:
- Fix-it-twice: Typescript checks on the JS files. (I elected a JS check instead of using compilation or ts-node to KISS)
- Add further protection against merging while a release is in progress

## New protections

Currently (assuming it was working), if a PR has all its checks ✅, we as admin could still merge it.
The only solution to prevent an admin to merge is to make a check of the PR to fail
To do that, I used the logic we had for the 'rogue' package-lock protection: a check that exit 1 or 0 depending on the presence of a file or not.
To enforce it, I need to "Requires branches to be up to date before merging", otherwise no further checks would be made.

So now to lock the branch:
 - We verify the main branch hasn't changed
 - We write and commit-push a 'lockfile' into the repo: this is the actual 'lock': from here, until this file get removed from master, nothing can go in or out.
 - We reset the local to the commit before (indeed we don't want to keep this file or commit)
 - We continue the process as usual
 - When we set where `master` should ref, we use 'force' because it is not a fast-forward operation. (i.e. we want to remove 'gitlock'

### DIFF
--- a/utils/release/git-lock.mjs
+++ b/utils/release/git-lock.mjs
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 import {gitPull, getSHA1fromRef} from '@coveo/semantic-monorepo-tools';
-import dedent from 'ts-dedent';
+import {dedent} from 'ts-dedent';
 
 import {
   limitWriteAccessToBot,
--- a/utils/release/git-lock.mjs
+++ b/utils/release/git-lock.mjs
@@ -1,15 +1,24 @@
 #!/usr/bin/env node
 
-import {gitPull, getSHA1fromRef} from '@coveo/semantic-monorepo-tools';
+import {
+  gitPull,
+  getSHA1fromRef,
+  gitCommit,
+  gitPush,
+  gitAdd,
+} from '@coveo/semantic-monorepo-tools';
 import {dedent} from 'ts-dedent';
 
 import {
   limitWriteAccessToBot,
   removeWriteAccessRestrictions,
 } from './lock-master.mjs';
+import {writeFileSync} from 'node:fs';
+import {spawnSync} from 'node:child_process';
 
 const isPrerelease = process.env.IS_PRERELEASE === 'true';
 const noLockRequired = Boolean(process.env.NO_LOCK);
+const PATH = '.';
 
 const ensureUpToDateBranch = async () => {
   // Lock-out master
@@ -29,6 +38,18 @@ const ensureUpToDateBranch = async () => {
   }
 };
 
+/**
+ * This will make .github\workflows\git-lock-fail.yml run and thus fail the associated check.
+ */
+const lockBranch = async () => {
+  writeFileSync('.git-lock', '');
+  await gitAdd('.git-lock');
+  await gitCommit('lock master', PATH);
+  await gitPush();
+  spawnSync('git', ['reset', '--hard', 'HEAD~1']);
+};
+
 if (!(isPrerelease || noLockRequired)) {
   await ensureUpToDateBranch();
+  await lockBranch();
 }
--- a/utils/release/git-publish-all.mjs
+++ b/utils/release/git-publish-all.mjs
@@ -204,6 +204,7 @@ async function commitChanges(releaseNumber, commitMessage, octokit) {
     repo: REPO_NAME,
     ref: `refs/heads/${mainBranchName}`,
     sha: commit.data.sha,
+    force: true,
   });
 
   // Delete the temp branch
--- a/utils/release/git-publish-all.mjs
+++ b/utils/release/git-publish-all.mjs
@@ -155,7 +155,7 @@ const getCliChangelog = () => {
 })();
 
 /**
- *
+ * "Craft" the signed release commit.
  * @param {string|number} releaseNumber
  * @param {string} commitMessage
  * @param {Octokit} octokit
--- a/utils/release/git-publish-all.mjs
+++ b/utils/release/git-publish-all.mjs
@@ -14,9 +14,17 @@ import {
   npmBumpVersion,
   getSHA1fromRef,
   gitSetupUser,
+  gitCreateBranch,
+  gitCheckoutBranch,
+  gitAdd,
+  gitWriteTree,
+  gitCommitTree,
+  gitUpdateRef,
+  gitPublishBranch,
 } from '@coveo/semantic-monorepo-tools';
 import {Octokit} from 'octokit';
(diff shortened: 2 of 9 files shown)


btaillon (approved):

utils/release/npm-publish-package.mjs
```diff
@@ -134,7 +146,8 @@ async function updateWorkspaceDependent(version) {
   )) {
     if (
       dependencies.find(
-        (dependency) => dependency.target === dependencyPackageName
+        (/** @type {{target:string}} **/ dependency) =>
+          dependency.target === dependencyPackageName
```
mrrajamanickam-coveo, line 150: Would using typecast `as string` work in this case instead?

utils/release/npm-publish-package.mjs
```diff
@@ -180,15 +204,28 @@ function isPrivatePackage() {
...
+ * @returns {Promise<string>}
+ */
 async function getNextBetaVersion(packageName, nextGoldVersion) {
+  let nextBetaVersion = `${nextGoldVersion}-0`;
   const registryMeta = await fetchNpm(packageName);
+
+  /**
+   * @type {string[]}
+   */
+  // @ts-ignore force-cast.
```
mrrajamanickam-coveo, line 220: why is this necessary?

utils/release/git-publish-all.mjs
```diff
@@ -146,6 +154,13 @@ const getCliChangelog = () => {
   });
 })();
 
+/**
```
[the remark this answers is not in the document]
louis-bompart: It's mostly to 'type' the param of the function, e.g. in typescript it would have been:
```ts
async function commitChanges(releaseNumber: string|number, commitMessage: string, octokit: Octokit) {
```

utils/release/npm-publish-package.mjs
louis-bompart: No, because we're in JS.

utils/release/npm-publish-package.mjs
louis-bompart: Because `fetchNpm` is loosely typed and return an `unknown` type.

utils/release/git-publish-all.mjs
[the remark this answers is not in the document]
mrrajamanickam-coveo: Yes a brief desc would suffice - it would also make it look less weird / incomplete.

utils/release/npm-publish-package.mjs
mrrajamanickam-coveo: Yes right 🤦

utils/release/npm-publish-package.mjs
mrrajamanickam-coveo: Got it, thanks. Maybe add a brief note as comment ?

mrrajamanickam-coveo (approved):

louis-bompart:
Sorry for the re-request @mrrajamanickam-coveo & @btaillon, https://github.com/coveo/cli/pull/1261/commits/c7e4e691deb7a6e2882def7c23d1fc40ddfab163 introduce yet another protection in the PR.

Indeed, without this changeset, we were still able to merge a PR if it passes all the checks.
To prevent that, I copied the logic we have to 'prevent' rogue package-lock.json and adapted it to protect the branch.
Updating pr descr and title with further details

utils/release/lock-master.mjs
```diff
@@ -18,20 +18,37 @@ export const limitWriteAccessToBot = () => changeBranchRestrictions(true);
...
     await octokit.rest.repos.setAdminBranchProtection(COVEO_CLI_MASTER);
   } else {
+    // Allow direct write access to the team
     await octokit.rest.repos.setTeamAccessRestrictions({
       ...COVEO_CLI_MASTER,
       teams: ['dx'],
     });
+    // Allow admins to bypass the status checks
     await octokit.rest.repos.deleteAdminBranchProtection(COVEO_CLI_MASTER);
+    // Do not requires branches to be up to date before merging
```
mrrajamanickam-coveo, line 48: Thanks for adding all the inline comments 👍

utils/release/git-lock.mjs
```diff
@@ -29,6 +38,18 @@ const ensureUpToDateBranch = async () => {
   }
 };
 
+/**
+ * This will make .github\workflows\git-lock-fail.yml run and thus fail the associated check.
+ */
+const lockBranch = async () => {
```
mrrajamanickam-coveo, line 44: So the purpose of this is to create a failed check that will prevent merging ?

Doesn't github have a way to setting that controls admins from bypassing checks?

Guessing this is not related a concurrency issue that could be solved using [Concurrency Workflow syntax for GitHub Actions - GitHub Docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) ?

utils/release/npm-publish-package.mjs
```diff
@@ -170,6 +189,11 @@ function updateDependency(packageJson, dependency, version) {
   }
 }
 
+/**
```
mrrajamanickam-coveo, line 192: Couple of functions missing desc here

utils/release/git-lock.mjs
louis-bompart: > So the purpose of this is to create a failed check that will prevent merging?

Exactly, because as long as we have to write access to the repo, if we pass the check we will be able to merge 'green' PR.
So I need to paint it red, because if I remove our write access... Well I have no means to reestablish them easily in case something goes wrong.

> Doesn't github have a way to setting that controls admins from bypassing checks?

Yep, and we enable this setting here:
https://github.com/coveo/cli/blob/c0b6610e156efed43cdb80d858333047e8f1b434/utils/release/lock-master.mjs#L38-L39

> Guessing this is not related a concurrency issue that could be solved using [Concurrency Workflow syntax for GitHub Actions - GitHub Docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) ?

No indeed, this cannot be resolved by concurrency for several reasons:
 1. If a PR has already passed the checks, it will be able to go through, concurrency wouldn't help prevent that.
 2. If we setup a concurrency constraint on the CI, that means that only one PR can run its checks at a time, this would slow us down

PR #1262, opened 2023-04-03
ci: use deploy key

louis-bompart:
## Proposed changes

Now with #1261, the bot has to overcome/bypass the checks. The way of doing it is using a Deploy Key.
This PR does that.

### DIFF
--- a/utils/release/git-lock.mjs
+++ b/utils/release/git-lock.mjs
@@ -6,6 +6,7 @@ import {
   gitCommit,
   gitPush,
   gitAdd,
+  gitSetupSshRemote,
 } from '@coveo/semantic-monorepo-tools';
 import {dedent} from 'ts-dedent';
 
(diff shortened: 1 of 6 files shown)


utils/release/git-publish-all.mjs
```diff
@@ -199,13 +201,11 @@ async function commitChanges(releaseNumber, commitMessage, octokit) {
...
-    repo: REPO_NAME,
-    ref: `refs/heads/${mainBranchName}`,
-    sha: commit.data.sha,
-    force: true,
-  });
+  await gitSetRefOnCommit(
+    GIT_SSH_REMOTE,
+    `refs/heads/${mainBranchName}`,
+    commit.data.sha
+  );
```
louis-bompart, line 208: Will need https://github.com/coveo/semantic-monorepo-tools/pull/148
```suggestion
  await gitSetRefOnCommit(
    GIT_SSH_REMOTE,
    `refs/heads/${mainBranchName}`,
    commit.data.sha,
    true
  );
```

utils/release/git-lock.mjs
```diff
@@ -42,10 +46,15 @@ const ensureUpToDateBranch = async () => {
  * This will make .github\workflows\git-lock-fail.yml run and thus fail the associated check.
  */
 const lockBranch = async () => {
+  const DEPLOY_KEY = process.env.DEPLOY_KEY;
+  if (DEPLOY_KEY === undefined) {
+    throw new Error('Deploy key is undefined');
+  }
+  await gitSetupSshRemote(REPO_OWNER, REPO_NAME, DEPLOY_KEY, GIT_SSH_REMOTE);
```
btaillon, line 53: I'm worried that updating the remote here may have unintended side-effects in workflows that call `lockBranch`, we should probably move `gitSetupSshRemote` out of here.

PR #1263, opened 2023-04-04
fix(hostedpages): add explanation about changes done in the header of the local html pages

olamothe:
https://coveord.atlassian.net/browse/KIT-2361

## Proposed changes

Adding an explanation about the title attribute of the local development page, which is discarded when uploaded to the Coveo platform


jpmarceau (approved):
👌

louis-bompart (approved):

btaillon (approved):

PR #1264, opened 2023-04-04
ci: remove duplicate scripts

louis-bompart:
## Proposed changes

We had duplicate instructions between the js files and the workflow. I think the JS is more readable. (i.e. logic is embedded in semantic-monorepo-tools)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,24 +62,6 @@ jobs:
           node-version-file: '.nvmrc'
       - name: Install dependencies
         run: npm ci
-      - name: Setup credentials
-        run: |
-          git config --global user.email action@github.com
-          git config --global user.name GitHub Action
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -62,24 +62,6 @@ jobs:
...
-        run: |
-          git config --global user.email action@github.com
-          git config --global user.name GitHub Action
-      - name: Setup git SSH remote
-        env:
-          DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
-        run: |
-          mkdir -p ~/.ssh
-          echo "${DEPLOY_KEY}" > ~/.ssh/id_rsa
-          chmod 400 ~/.ssh/id_rsa
```
louis-bompart, line 75: This in particular was causing issue.

y-lakhdar (approved):

PR #1272, opened 2023-04-11
feat: sanitize custom component name on creation

y-lakhdar:
## Proposed changes

* Move duplicated code to dedicated package `@coveo/create-atomic-commons`.
* Add validation rules `utils/create-atomic-commons/src/validation.ts` to ensure component name respects HTML standards. These rules are the same as the one used by StencilJs.

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


packages/ui/atomic/create-atomic-component/index.js
```diff
@@ -1,35 +1,27 @@
...
 cpSync(templateDirPath, cwd(), {
   recursive: true,
 });
 
 let componentName = process.argv[2];
 if (componentName) {
-  if (!componentName?.includes('-')) {
-    componentName = `atomic-${componentName}`;
-  }
-
```
y-lakhdar, line 32: No need to manipulate component name since it is now handled by the validation rules

packages/ui/atomic/create-atomic-component/index.js
```diff
@@ -1,35 +1,27 @@
...
+  ensureComponentValidity(componentName);
```
y-lakhdar, line 24: Throws (with a nice message :) ) whenever the component name does not respect HTML standards

packages/ui/atomic/create-atomic-component/tests/test.spec.ts
```diff
@@ -49,12 +53,6 @@ describe(PACKAGE_NAME, () => {
       testDirname: 'valid-arg',
       packageName: 'valid-component-name',
     },
-    {
-      describeName: 'when called with a component name without hyphen',
-      args: ['nohypen'],
-      testDirname: 'invalid-arg',
-      packageName: 'atomic-nohypen',
-    },
```
y-lakhdar, line 57: No need to test this use case because it will never pass the validation rules. The cli will simply throw an error message. this is tested in utils/create-atomic-commons/src/validation.spec.ts

utils/create-atomic-commons/src/transformers.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+  it("should camelize each word's first letter", () => {
+    expect(camelize('my-super-tag')).toBe('MySuperTag');
+  });
+
+  it('should have nothing to camelize', () => {
+    expect(camelize('MySuperTag')).toBe('MySuperTag');
+  });
+});
+
+// TODO: CDX-1406: add unit tests to transformer
```
y-lakhdar, line 17: This PR is long enough as it is. I'm keeping unit tests for another PR

utils/create-atomic-commons/src/console.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  $ npm start
+    Starts the development server.
+
+  $ npm run build
+    Builds your project in production mode.
+
+  Further reading:
+
+   - TODO: CDX-1403 Add link to documentation in source code and error message
```
olamothe, line 19: I would personally remove the TODO from the message, as those have the tendency to slip by in the released product.

We can leave it as a comment in this file though.

olamothe (approved):

utils/create-atomic-commons/src/validation.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (tag.length === 0) {
+    throw new Error(`Received empty tag value`);
+  }
+
+  if (tag.indexOf(' ') > -1) {
+    throw new Error(`"${tag}" tag cannot contain a space`);
+  }
+
+  if (tag.indexOf(',') > -1) {
+    throw new Error(`"${tag}" tag cannot be used for multiple tags`);
```
louis-bompart, line 24: 🤨

utils/create-atomic-commons/src/validation.ts
louis-bompart, line 1: Discussed through Slack with @y-lakhdar, we think those rules might be too complicated, and we could simplify it for both us and the end-user by limiting their choices with stringent rules. Essentially, we think that:
 - Limiting the authorized character set to `[az-]` (`qwertyuiopasdfghjklzxcvbnm-`)
 - Forbid leading hyphen `/^-/`
 - Forbid trailing hyphen `/-$/`
 - Forbid multi-hyphen `/-{2,}/`

We could also propose better error printing by inspiring ourselves from how jest hight error in strings.

Here's a snippet of code that does that for the 'no double dash' rule (it is tho simple to adapt adn
```js
import chalk from "chalk";

const invalidCmpName = "patate---foo--bar-baz";
const noDoubleHyphen = /-{2,}/g;

const badMatches = invalidCmpName.matchAll(noDoubleHyphen);
let cursor = 0;
let message = "";
for (const match of badMatches) {
  message += chalk.red(invalidCmpName.slice(cursor, match.index));
  cursor = match.index + match[0].length;
  message += chalk.red.inverse(invalidCmpName.slice(match.index, cursor));
}
message += chalk.red(invalidCmpName.slice(cursor));
console.log(message || chalk.green(invalidCmpName));
```
image

y-lakhdar: Seems like this approach might be more complex to include more regex as we go

louis-bompart: As discussed through  slack, you transform the code snippet above into a reusable functions quite easily, making this pretty reusable, the catch is to not do 'a big regex to catch'em'all), but several small ones that target specific, humanly explainable problem.

utils/create-atomic-commons/src/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Validates that a component tag meets required naming conventions to be used for a web component
+ * @param tag the tag to validate
+ * @returns an error message if the tag has an invalid name, undefined if the tag name passes all checks
+ */
+export const ensureComponentValidity = (tag: string): void | never => {
+  const errors: Error[] = [];
+  if (tag !== tag.toLowerCase()) {
+    errors.push(new Error(`Tag can not contain upper case characters`));
```
mrrajamanickam-coveo, line 11: "can not" -> cannot

utils/create-atomic-commons/src/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+  if (tag.length === 0) {
+    errors.push(new Error(`Received empty tag value`));
+  }
+
+  if (tag.indexOf(' ') > -1) {
```
mrrajamanickam-coveo, line 17: `.includes()` could be more readable than using `.indexOf()`

utils/create-atomic-commons/src/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  if (tag.indexOf('--') > -1) {
+    errors.push(
+      new Error(
+        `"${tag}" tag cannot contain multiple dashes (--) next to each other`
+      )
+    );
+  }
+
+  if (tag.indexOf('-') === 0) {
```
mrrajamanickam-coveo, line 48: `.startsWith()` would be more readable

utils/create-atomic-commons/src/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+    errors.push(new Error(`"${tag}" tag cannot start with a dash (-)`));
+  }
+
+  if (tag.length > 0 && tag.lastIndexOf('-') === tag.length - 1) {
```
mrrajamanickam-coveo, line 52: `.endsWith()` would be more readable

utils/create-atomic-commons/src/validation.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  if (tag.indexOf('-') === -1) {
+    errors.push(
+      new Error(
+        `"${tag}" tag must contain a dash (-) to work as a valid web component`
+      )
+    );
+  }
+
+  if (tag.indexOf('--') > -1) {
```
mrrajamanickam-coveo, line 40: Can't these `indexOf()`s be replaced by `includes()`?

louis-bompart (approved):

PR #1275, opened 2023-04-14
fix(cli-core,cli-source,ui-react): enforce LTS

louis-bompart:
## Proposed changes

Warn the user if the version of Node used by the CLI does not match our engines requirement

This come from one of our internal user using some version of Node 18 pre-LTS.
This version of Node he was using used a buggy implementation of fetch, which was causing the Authentication check to fail.

## Testing

- [x] Unit Tests:
- [x] Manual Tests:
  Install unsupported Node version (let's say 20) and run the CLI


packages/cli/core/src/hooks/prerun/checkNodeVersion.ts
```diff
@@ -0,0 +1,14 @@
+import {Hook} from '@oclif/core';
+import {satisfies} from 'semver';
+import dedent from 'ts-dedent';
+const hook: Hook<'prerun'> = function (opts) {
+  if (!satisfies(process.version, this.config.pjson.engines.node)) {
+    this.warn(
+      dedent`NodeJS ${process.version} is not supported. The Coveo CLI might malfunction.
+      Please update your NodeJS installation to the latest LTS version.`
```
louis-bompart, line 8: @srosenberg-coveo wdyt?

packages/cli/core/package.json
```diff
@@ -4,7 +4,7 @@
   "author": "Coveo",
   "description": "Coveo CLI",
   "engines": {
-    "node": "16.x || 18.x"
+    "node": "^16.13.0 || ^18.12.0"
```
mrrajamanickam-coveo, line 7: what are the pros/cons of specifying the minor version explicitly ?

mrrajamanickam-coveo: do we need to keep updating this number to reflect the latest LTS? Manually?

mrrajamanickam-coveo: What necessitated this change? Would be good to update PR desc with the reason.

mrrajamanickam-coveo (commented):
for PR title "enforce LTS" -> something like 'check and warn on node version' might be more appropriate

packages/cli/core/package.json
louis-bompart: > what are the pros/cons of specifying the minor version explicitly ?

So, for the CLI we only support a select set of LTS(s) versions of NodeJS, currently Gallium(16) and Hydrogen(18)
The trick is that Node 18.x.x (or 16.x.x) is not necessarily an LTS version. Indeed, Node 18 became LTS with 18.12.0.
So, `18.x` was not a good semver specifier for "Hydrogen", instead we need to support:
"all versions of Node 18 with a minor equal or superior to 12." this is well managed by a caret range:
https://github.com/npm/node-semver#caret-ranges-123-025-004

> do we need to keep updating this number to reflect the latest LTS? Manually?

Only when we decide to support a new Major (next one will be 20)

> What necessitated this change? Would be good to update PR desc with the reason.

- [ ]  update desc

mrrajamanickam-coveo: Got it, thanks - it is weird though that the minor version is not mentioned in https://nodejs.dev/en/about/releases/ against LTS but then the info is mentioned in https://nodejs.org/en/blog/release/v18.12.0

In order to keep tabs on new LTS releases wonder if we can add a check in CI or a warning to dev workflow (e.g. using [node-lts-versions](https://github.com/msimerson/node-lts-versions) or [nodejs-versions](https://github.com/wallet77/nodejs-versions)) - probably as part of common monorepo tooling - so that we (dev team) will get an automated notification/warning when there is a new LTS release?

louis-bompart: > In order to keep tabs on new LTS releases wonder if we can add a check in CI or a warning to dev workflow (e.g. using [node-lts-versions](https://github.com/msimerson/node-lts-versions) or [nodejs-versions](https://github.com/wallet77/nodejs-versions)) - probably as part of common monorepo tooling - so that we (dev team) will get an automated notification/warning when there is a new LTS release?

I think not, for two reasons:
 - We ain't good at following up on automated alerts, due to alerts fatigue (something we would have ultimately to find a solution to)
 - https://xkcd.com/1205/  we're pretty much in the upper right corner for this IMO, so the reasonable amount of time to work on that would be 5 seconds 🤣 

Any changes to our Node-supported version will be a breaking change (at least for the foreseeable future), so it's part of a wider process.
This process should include a check on the Node release calendar so that we can foresee when our next major release will be

y-lakhdar (approved):

mrrajamanickam-coveo (approved):

packages/cli/core/src/hooks/prerun/checkNodeVersion.ts
srosenberg-coveo: Sorry for the delay. Looks great to me!

srosenberg-coveo (approved):

PR #1277, opened 2023-04-17
test(verdaccio): ensure package publish order

louis-bompart:
## Proposed changes

Previous logic doesn't ensure that the local dependencies of the package are published prior.
This does it (very inefficiently)

The algorithm is _very_ naive but should work
I'm honestly too lazy to create a proper tree struct. here.

## Testing

https://github.com/coveo/cli/pull/1278
https://github.com/coveo/cli/pull/1279

If those are green, all good 😎

### DIFF
--- a/utils/verdaccio-starter/src/index.ts
+++ b/utils/verdaccio-starter/src/index.ts
@@ -22,10 +22,12 @@ export async function startVerdaccio(
       publishProxiedPackages
     );
   }
-  const computedPackagesToProxy = new Set<string>();
+  const computedPackagesToProxy = new Map<string, string[]>();
 
   if (!includeLocalDependent) {
-    packagesToProxy.forEach(computedPackagesToProxy.add);
+    packagesToProxy.forEach((packageName) =>
+      computedPackagesToProxy.set(packageName, [])
+    );
   } else {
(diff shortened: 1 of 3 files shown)


utils/verdaccio-starter/src/index.ts
```diff
@@ -129,12 +133,44 @@ function computeVerdaccioConfig(packagesToProxy: Set<string>) {
 }
 
 function doPublishProxiedPackages(
-  computedPackagesToProxy: Set<string>,
+  computedPackagesToProxy: Map<string, string[]>,
   verdaccioUrl: string,
   npmConfigPath: string
 ) {
```
btaillon, line 139: I feel like this would be easier to understand if it took an ordered list like before, but a separate function (e.g., `topologicalSort()`) was responsible for generating a properly ordered array.

btaillon (approved):

utils/verdaccio-starter/src/index.ts
louis-bompart: Probably, but it's more complex to do, and I don't think it's a good idea to spend time doing it, at least now

y-lakhdar (approved):

PR #1284, opened 2023-04-19
fix(atomic): wait for atomic to load/init on page component

louis-bompart:
## Proposed changes

The initializeBindings fail without the preliminary await

## Testing

Tricky, I don't know how to do an automated test for that.
My test is:
 - try use the HSP feature with a page component without this await, the page component should not initialize properly
 - try use the HSP feature with a page component with this await, the page component should initialize properly

### DIFF
--- a/packages/ui/atomic/create-atomic-component/template/src/components/sample-component/src/sample-component.tsx
+++ b/packages/ui/atomic/create-atomic-component/template/src/components/sample-component/src/sample-component.tsx
@@ -51,6 +51,8 @@ export class SampleComponent {
   // Using `componentWillLoad` will hang the parent atomic-search-interface initialization.
   public async connectedCallback() {
     try {
+      // Wait for the Atomic to load.
+      await customElements.whenDefined('atomic-search-interface');
       // Wait for the Atomic bindings to be resolved.
       this.bindings = await initializeBindings(this.host);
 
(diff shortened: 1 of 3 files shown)


packages/ui/atomic/create-atomic-component/template/src/components/sample-component/src/sample-component.tsx
```diff
@@ -51,6 +51,8 @@ export class SampleComponent {
   // Using `componentWillLoad` will hang the parent atomic-search-interface initialization.
   public async connectedCallback() {
     try {
+      // Wait for the Atomic to load.
+      await customElements.whenDefined('atomic-search-interface');
```
y-lakhdar, line 55: I think there is already a method for that in `packages/ui/atomic/template/src/utils/atomic.ts` called
```
waitForAtomic()
```

y-lakhdar (changes requested):

y-lakhdar (approved):

packages/ui/atomic/create-atomic-component/template/src/components/sample-component/src/sample-component.tsx
louis-bompart: Yep, it's a copy from there, but its in another package, so that's where the copy come from

PR #1295, opened 2023-04-21
chore: stabilize ci after release, ensure prettier run when it should

louis-bompart:
CDX-1438

Lot of issues:
 - `lint-staged` run before the files are actually 'staged'
 - Our snap now needs to be updated after a PR
 - Lockfile was not properly update for private package depending on public package (while both are in the workspace)


utils/release/lock-master.mjs
```diff
@@ -26,6 +26,7 @@ export const removeWriteAccessRestrictions = () =>
 async function changeBranchRestrictions(onlyBot) {
   const octokit = new Octokit({auth: process.env.GITHUB_CREDENTIALS});
   // Requires branches to be up to date before merging
+  //@ts-expect-error param is a bit messed-up, it expects {data:never}, which doesn't make any sense.
```
louis-bompart, line 29: Temp: Used expect-error instead of ignore so that we'll be flagged when it's been fixed by GH

olamothe (approved):

PR #1318, opened 2023-06-14
fix(cli): use ci-friendly confirm

louis-bompart:
## Proposed changes

When in a CI, you cannot use confirm. This add the possibility of using a default value when in a CI for confirm.

## Testing

- [x] Unit Tests:

- [x] Functionnal Tests: existing (ish)

### DIFF
--- a/packages/cli/core/src/lib/utils/cli.ts
+++ b/packages/cli/core/src/lib/utils/cli.ts
@@ -1,4 +1,4 @@
-import {CliUx} from '@oclif/core';
+import {confirm} from '@coveo/cli-commons/utils/ux';
 
 import globalConfig from '@coveo/cli-commons/config/globalConfig';
 import {buildEvent} from '@coveo/cli-commons/analytics/eventUtils';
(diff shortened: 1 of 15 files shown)


louis-bompart:
will be rebased merged.

packages/cli/core/src/lib/utils/cli.ts
```diff
@@ -14,7 +14,7 @@ export async function confirmWithAnalytics(
   message: string,
   questionName: string
 ): Promise<boolean> {
-  const doAction = await CliUx.ux.confirm(message);
+  const doAction = await confirm(message, false);
```
y-lakhdar, line 17: `false`? So essentially, we're saying that all command requiring input will fail in a CI.

Also, there are other command that uses confirm like the deploy command

louis-bompart: > essentially, we're saying that all command requiring input will fail in a CI.

Yes, this is the safest path: all action where we prompt the user for a confirmation should require an input from the user, through flag or otherwise (or it should never reach the stage where another input is required)

For now, it's a blanket move: if the CLI somehow asks for a confirmation and you're in a CI environment, we default to no, let's not do that.

> Also, there are other command that uses confirm like the deploy command

Indeed: They will fail 'smoothly' instead of waiting forever for an input.
If we want to tackle the behavior of the deploy command in a CI (i.e. should it overwrite or not), we can tackle it in its own Jira.

The goal of this PR is just to ensure we do not 'hang' anymore.

y-lakhdar (approved):

PR #1321, opened 2023-08-07
fix(cli): remove deploy.example

louis-bompart:
## Proposed changes

Unused file, some customers were using as documentation/example, but we do not maintain it.
So remove it to reduce confusion (instead customers should check docs.coveo.com)


mrrajamanickam-coveo (approved):
Can we add a link in Readme to the example in docs.coveo ?

PR #1322, opened 2023-09-05
chore: migrate to new amplitude node sdk

louis-bompart:
## Proposed changes

Tldr; https://www.docs.developers.amplitude.com/data/sdks/typescript-node/migration/

Also refactor dup code between core and commons, removed duplicate from core and make it point to commons

## Breaking changes

`@coveo/cli-commons` some function sig changed, flagged as such in appropriate commits, will rebase merge.

## Testing

- [x] Unit Tests


btaillon (approved):
From looking at their migration docs, this looks good to me

PR #1335, opened 2023-10-18
feat(cli): change node support to only 18-20

olamothe:
https://coveord.atlassian.net/browse/CDX-1408

## Proposed changes

I recently changed these files to drop support for 16. However, we want to only support 18-20 ranges (not everything above 18)

## Breaking changes

Drop support for older version of node js

## Testing

- [ ] Unit Tests:

- [ ] Functionnal Tests:

- [ ] Manual Tests:


louis-bompart (commented):
We also need to test on 20, so let's bump the CI thing

dmbrooke (approved):

PR #1409, opened 2024-01-17
feat(atomic): add support for ngsp

louis-bompart:
Add support for NGSP in `ui:create:atomic` (and `atomic:init -t=app`)
Includes as well:
 - E2E stabilization: _Actually_ the latest published version of the CLI was tested, not the one of the PR 😅 (because CLI build were silently failing)
  - Update of a bunch of deps, again, to avoid silent failures.
  - Normalization of the createSearchHub confirm (so that it matches the others)
 
 
 All that E2E tested.


packages/ui/atomic/create-atomic/src/fetch-page.ts
```diff
@@ -7,20 +7,32 @@ import {
...
+  pageId: string,
+  type: 'next-gen' | 'legacy' | 'unknown'
 ) {
-  const manifest = await client.searchInterfaces.manifest(pageId, {
-    pagePlaceholders: {
-      results: '--results--',
-    },
-  });
-  return replaceResultsPlaceholder(manifest);
+  let manifestGetters = [];
```
olamothe, line 19: Do we really still need the support for `legacy` ? 🤔 I don't see where we would be using or needing it in the CLI.

Could we just delete all the code related to downloading legacy page (function `getLegacyManifest` in this file) ?

olamothe (approved):

packages/ui/atomic/create-atomic/src/fetch-page.ts
louis-bompart: Yes imho. Removing this would be a breaking change for people using the legacy search pages still.
It's important to remember that, from the user perspective, there's no distinction between legacy and next Gen search pages in the UI or anywhere else.

PR #1416, opened 2024-01-30
chore(tests): try to fix failing e2e tests by increasing beforeAll timeout

fbeaudoincoveo:
## Proposed changes

There are a couple of e2e tests in the atomic spec that keep failing at the same spot, namely in the `beforeAll` at `when using an existing pageId (--pageId flag specified)` > `when the project is configured correctly`.

I’ve already verified that running the same command locally against the stg org / pageId used in the test works, so that should not be the issue.

The test fails with a timeout, so maybe increasing it will solve the issue.


louis-bompart:
I think it works in the end? Should we close @fbeaudoincoveo

PR #1417, opened 2024-01-31
feat(cli): add support to associate a search hub when creating a search page

olamothe:
https://coveord.atlassian.net/browse/CDX-1485

## Proposed changes

On all search page creation process, when an API needs to be created, prompt the user to either create a new search hub, or associate an existing one.

Use the search usage metric service to list available hubs in the current org if the client wants to re-use an existing one, or leave the choice to create a new one.

It is still possible to create a new key with no search hub by explicitly denying the prompt.

## Testing

- [X] Unit Tests:

- [ ] Functionnal Tests:

Existing suite of test for page creation should continue to work. However, all prompts will be skipped in CI.

- [X] Manual Tests


packages/cli/core/src/commands/atomic/init.ts
louis-bompart, line 1: Should be in `packages\cli\core\src\lib\atomic\createAtomicProject.ts` IMO, to be reflected in also `atomic:init`

louis-bompart (approved):

PR #1453, opened 2024-03-12
fix(snapshot): define resourceToPreview from resources changed. 

louis-bompart:
## Proposed changes

The current algorithm doesn't take into account the resources the user can access.
To solves that, we could either:
 - Query/cache what resources the user can access when we query it
 - Only do the expanded preview on resources that change.

## Testing

- [ ] Unit Tests:
Snapshot UTs are tangled, changing it will be difficult/require a rewrite.
- [ ] Functionnal Tests:
Corner-case, would require to create orgs without some access right or some stuff like that.
- [ ] Manual Tests:
In the only known affected environment (customer-owned), I managed to reproduce the issue the same customer reported.
 - Tested with the current version of the CLI, dev build from master: Crash
 - Tested with dev build from this branch: No Crash.

### DIFF
--- a/packages/cli/core/src/lib/snapshot/snapshotReporter.ts
+++ b/packages/cli/core/src/lib/snapshot/snapshotReporter.ts
@@ -71,13 +71,8 @@ export class SnapshotReporter {
     const isResourceTypeReportAffected = (
       report: ResourceSnapshotsReportOperationModel
     ) => {
-      for (const operation of Object.values(report)) {
-        if (operation > 0) {
-          return true;
-        }
-      }
-      return false;
-    };
+    
+    Object.values(report).reduce((operation, total) => operation + total, 0) === 0;
     const changedResourcesTypes: ResourceSnapshotType[] = [];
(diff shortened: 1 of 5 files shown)


packages/cli/core/src/lib/snapshot/snapshotReporter.ts
```diff
@@ -67,6 +67,30 @@ export class SnapshotReporter {
...
+    const isResourceTypeReportAffected = (
+      report: ResourceSnapshotsReportOperationModel
+    ) => {
+      for (const operation of Object.values(report)) {
+        if (operation > 0) {
+          return true;
+        }
+      }
+      return false;
+    };
```
olamothe, line 80: Suggestion: 

```suggestion
    const isResourceTypeReportAffected = (
      report: ResourceSnapshotsReportOperationModel
    ) => {
    
    Object.values(report).reduce((operation, total) => operation + total, 0) === 0;
```

olamothe (approved):

PR #1474, opened 2024-06-19
fix(cli): org:search:dump issues when dump is fetched across multiple queries

fbeaudoincoveo:
## Proposed changes

**Context:**

The `org:search:dump command` executes queries under the hood. When we execute this command, it's possible that we hit the Search API maximum response size. The likelihood of hitting that limit is increased in the HIPAA environment, as the HIPAA maximum response size is 4x smaller than in PROD.

When the dump is executed across multiple queries, we ensure that results are fetched from the same index by passing the `indexToken` of the initial query to each subsequent queries. This is, by definition, incompatible with using the index cache, as stated in the `indexToken` parameter documentation:

> The Base64 encoded identifier of the index mirror to forward the request to. See also the index parameter.
> If you do not specify an indexToken (or index) value, any index mirror could be used.
> Note: Passing an indexToken (or index) value has no effect when the results of a specific request can be returned from cache (see the maximumAge parameter).

Another issue is that in order to know what result to start from in each subsequent query after the first, we request results to be ordered by ascending `rowid` value, and we request only results whose `rowid` is greater than the `rowid` of the last result returned in the previous query. However, the `rowid` value is returned as a very large integer (greater than the maximum safe integer size JavaScript can handle without losing precision). Therefore, the expression `rowid>ROW_ID_OF_LAST_RESULT` typically behaves like `rowid>=ROW_ID_OF_LAST_RESULT` because JavaScript rounds the `rowid` during the JSON.parse conversion of the response body.

Fix:

- Setting `maximumAge: 0` on every query performed for fetching source dumps trivially addresses the index cache issue.

- Using response handler to the platform client that uses the new `context` parameter of the `JSON.parse` reviver (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#parameters) allows us to avoid losing precision when converting rowid values. However, this new parameter is only available from Node 21, so we have to polyfill it for now. **EDIT:** After discussion with the DX team, this change was made directly in the platform-client project. See https://github.com/coveo/platform-client/pull/834.

## Breaking changes

None

## Testing

- [x] Unit Tests: Yes
- [x] Functionnal Tests: No, unit tests are sufficient to ensure that maximumAge is set to 0
- [x] Manual Tests: Tested the `org:search:dump` command locally before / after platform-client update against problematic HIPAA test organization. The duplicate / inconsistent dump result issue is no longer present after the fix.

### DIFF
--- a/packages/cli/commons/src/platform/authenticatedClient.ts
+++ b/packages/cli/commons/src/platform/authenticatedClient.ts
@@ -1,9 +1,11 @@
+import 'core-js/actual/json/parse';
 import {setGlobalDispatcher, ProxyAgent, FormData, fetch} from 'undici';
-import PlatformClient from '@coveo/platform-client';
+import PlatformClient, {ResponseHandlers} from '@coveo/platform-client';
 import {Config, Configuration} from '../config/config';
 import {castEnvironmentToPlatformClient} from './environment';
 import globalConfig from '../config/globalConfig';
 import 'fetch-undici-polyfill';
+import {Response} from 'node-fetch';
 
 export class AuthenticatedClient {
   public cfg: Config;
@@ -39,12 +41,47 @@ export class AuthenticatedClient {
     if (!Object.keys(global).includes('FormData')) {
       Object.assign(global, {FormData});
     }
+
     return new PlatformClient({
       globalRequestSettings,
       environment: castEnvironmentToPlatformClient(resolvedConfig.environment),
       region: resolvedConfig.region,
       organizationId: resolvedConfig.organization,
       accessToken: resolvedConfig.accessToken!,
+      responseHandlers: [
+        {
+          canProcess: (response: Response): boolean => {
+            return response.ok;
+          },
+          process: async <T>(response: Response) => {
+            const content = await response.text();
+            return (
+              JSON.parse as (
+                text: string,
+                reviver?: (
+                  key: string,
+                  value: any,
+                  context: {source: any}
+                ) => any | undefined
+              ) => any
+            )(content, (key, value, context) => {
+              if (
+                key === 'rowid' &&
+                typeof value === 'number' &&
+                !Number.isSafeInteger(value)
+              ) {
+                return context.source;
+              }
+              return value;
+            });
+          },
+        },
+        ResponseHandlers.badGateway,
+        ResponseHandlers.blockedByWAF,
+        ResponseHandlers.error,
+        ResponseHandlers.htmlError,
+        ResponseHandlers.noContent,
+      ],
     });
   }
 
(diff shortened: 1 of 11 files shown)


packages/cli/commons/src/platform/authenticatedClient.ts
```diff
@@ -39,12 +41,47 @@ export class AuthenticatedClient {
...
+                typeof value === 'number' &&
+                !Number.isSafeInteger(value)
+              ) {
+                return context.source;
+              }
+              return value;
+            });
+          },
+        },
+        ResponseHandlers.badGateway,
```
olamothe, line 79: So we are "copying" the handlers from platform client as a "snapshot" of what it does today ?

Which means these will drift over time....

packages/cli/commons/src/platform/authenticatedClient.ts
```diff
@@ -39,12 +41,47 @@ export class AuthenticatedClient {
...
+                  key: string,
+                  value: any,
+                  context: {source: any}
+                ) => any | undefined
+              ) => any
+            )(content, (key, value, context) => {
+              if (
+                key === 'rowid' &&
+                typeof value === 'number' &&
+                !Number.isSafeInteger(value)
```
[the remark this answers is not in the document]
fbeaudoincoveo: I limited it to the `rowid` key because it's the only value that actually needs to be fixed at the moment (at least, as far as we know), but it's probably a good idea to do that, yes.

packages/cli/commons/src/platform/authenticatedClient.ts
fbeaudoincoveo: True, there could be additional default response handlers in the future. I'll see if I can do something better than that.

packages/cli/commons/src/platform/authenticatedClient.ts
```diff
@@ -1,9 +1,11 @@
+import 'core-js/actual/json/parse';
```
louis-bompart, line 1: Create a Jira & todo for CLI v4 + comment here

packages/cli/commons/src/platform/authenticatedClient.ts
louis-bompart: Why not put the handler in the platform-client code?

fbeaudoincoveo: I'm hesitant to do that because it would technically be introducing a breaking change in the platform-client...

fbeaudoincoveo:
I moved the bulk of the changes to the platform-client project: https://github.com/coveo/platform-client/pull/834

So basically all I'm doing in this PR now is setting `maximumAge: 0` on dump queries.

louis-bompart (approved):
LGTM, after the release of the platform-client and its update in our project ofc

PR #1500, opened 2025-02-19
chore: disable all tests

louis-bompart:
We'll need to bypass the broken tests to manage a release in a reasonable time. To do so, let’s disable them.

CDX-1610


fbeaudoincoveo (approved):
Seems safe enough 🙃

PR #1509, opened 2025-03-25
fix: typo in update cmd

louis-bompart:
CDX-764

Fix a typo.
The branch was also used to hotfix a botched release, more details in Slack.


louis-bompart:
Note-to-self: should be rebase merged.

alexprudhomme (approved):

PR #1515, opened 2025-08-19
feat(cli): move over all atomic-template projects to ui-kit

olamothe:
https://coveord.atlassian.net/browse/KIT-4522

## Proposed changes

Delete all atomic template projects from the CLI repository, in order to instead rely on those that have been moved over in ui-kit.

The atomic related command will now support specifying the version of the template project to use, as opposed to the "local" project version.


louis-bompart (approved):

packages/cli/core/src/lib/atomic/createAtomicProject.ts
```diff
@@ -33,7 +33,7 @@ export const atomicAppInitializerPackage = '@coveo/create-atomic';
 export const atomicLibInitializerPackage =
   '@coveo/create-atomic-component-project';
 
-const supportedNodeVersions = '^18.18.1 || ^20.9.0';
+const supportedNodeVersions = '^22.0.0';
```
louis-bompart, line 36: This is a breaking change

louis-bompart (changes requested):

PR #1547, opened 2026-04-29
chore(node): add Node 24 support across monorepo

jfcere-coveo:
KIT-4630

## Proposed changes

This PR introduces a minimal set of updates to support Node 24 in the CLI monorepo and ensure installation works correctly on modern macOS environments.

### Node 24 support

Added Node 24 to supported engine ranges in:
- `packages/cli/core/package.json`
- `packages/cli/source/package.json`
- `packages/ui/cra-template/package.json`

Updated Node version validation used during Atomic project creation:
- `packages/cli/core/src/lib/atomic/createAtomicProject.ts`

Update Node version used for CI
- Updated`.nvmrc` to enable CI to use Node 24

### Installation script portability

Updated e2e postinstall script to use Python 3 explicitly:
- Changed `pip install ...` to `python3 -m pip install ...` in `packages/cli-e2e/package.json`

### Lockfile update

- Updated `package-lock.json` to reflect the engine/version metadata changes above.

## Testing

- [ ] Unit Tests: Not modified
- [ ] Functionnal Tests: Not modified
- [x] Manual Tests: Use a few commands


louis-bompart (approved):
Should be OK I think

jfcere-coveo:
@louis-bompart I've updated `.nvmrc` to enable Node 24 in the CI

.nvmrc
```diff
@@ -1 +1 @@
-22.18.0
\ No newline at end of file
+24.15.0
```
louis-bompart, line 1: Nit; Guidance around "how to specify node version" has slightly evolved;
To avoid "mapping" a specific version of NodeJS, nowadays, I think it's best to use `lts/codename` instead, so that we always get the latest patch of the lts.

So here

```suggestion
lts/krypton
```

Ref: https://nodejs.org/en/about/previous-releases

louis-bompart (approved):

.nvmrc
jfcere-coveo: It is considered a best practice to use a pinned version as it avoids potential Node.js regressions that could break the CI without any changes to the code.

